PR #1, opened 2017-10-05
initial commit of Eclipse Ditto

thjaeckle:
Signed-off-by: Thomas Jaeckle <thomas.jaeckle@bosch-si.com>


geglock:
Great preparation!
We have to wait for write permissions before we can accept this pull request.

Issue #5, opened 2017-10-09
Provide pre-built docker images in public repository


crenz:
Please consider making docker images available e.g. on hub.docker.com for those that are interested to try out ditto, but do not want to set up a complete build environment for it.

zantiu:
+1

thjaeckle:
Sure, will do once we have the build process set up.

thjaeckle:
Images are available for the [0.1.0-M1](https://github.com/eclipse/ditto/releases/tag/0.1.0-M1) release on Docker Hub.

PR #6, opened 2017-10-09
Provides instructions how to build Ditto from within a Maven Docker image

crenz:
For those that do not have an up-to-date Maven and JDK environment available, it might be helpful to use a Docker-based build environment instead. Since this is non-trivial, I provided some instructions how this can be achieved. This might also be a temporary solution for issue #5 .

Signed-off-by: Christian Renz <crenz@web42.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ You have now running:
 
 #### Build from within Docker image
 
-If you do not have the right Maven and JDK version available, you can also use a Maven Docker image as build environment:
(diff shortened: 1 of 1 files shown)


thjaeckle:
Hi Christian.

Thanks for the PR - good idea!
This however will only work for Linux systems, won't it? Should be documented that way then..

Regards Thomas

crenz:
Good point. I have tried to clarify the instructions.

thjaeckle:
Congrats for the first merged "foreign" pull request to Ditto :clinking_glasses:

PR #12, opened 2017-10-13
introduced live commands to ditto-signals

jokraehe:
Signed-off-by: Johannes Kraeher <johannes.kraeher@bosch-si.com>


thjaeckle:
Reviewed and merged

PR #26, opened 2017-10-27
calculate "read subjects" for events in advance

thjaeckle:
... when processing commands:

* got rid of "external" types and therefore events
* changed "-dev" profiles to connect to mongoDb on localhost instead of 192.168.56.11

Signed-off-by: Thomas Jaeckle <thomas.jaeckle@bosch-si.com>

### DIFF
--- a/services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/AbstractPolicyEnforcerActor.java
+++ b/services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/AbstractPolicyEnforcerActor.java
@@ -32,6 +32,7 @@ import org.eclipse.ditto.json.JsonValue;
 import org.eclipse.ditto.model.base.exceptions.DittoJsonException;
 import org.eclipse.ditto.model.base.exceptions.DittoRuntimeException;
 import org.eclipse.ditto.model.base.headers.DittoHeaders;
+import org.eclipse.ditto.model.base.headers.DittoHeadersBuilder;
 import org.eclipse.ditto.model.base.headers.WithDittoHeaders;
 import org.eclipse.ditto.model.policies.Permissions;
 import org.eclipse.ditto.model.policies.PoliciesResourceType;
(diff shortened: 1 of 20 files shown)


services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/AbstractPolicyEnforcerActor.java
```diff
@@ -190,10 +188,26 @@ protected boolean isEnforcerAvailable() {
 
     protected <T extends WithDittoHeaders> T enrichDittoHeaders(final WithDittoHeaders<T> withDittoHeaders,
```
jufickel-b: IMHO it would be better to replace `WithDittoHeaders` with `Signal`. Then the ugly `instanceof` checks could be replaced with a check for the Signal type like for example `if (CreatePolicy.TYPE == signal.getType() ...)`.

thjaeckle: ah, good point - this is a relict from where `CommanResponses` were not yet a `Signal`

thjaeckle: but regarding the `instanceof` checks vs a `CreatePolicy.TYPE.equals(signal.getType())` i personally think the instanceof check is clearer

jufickel-b: Well, this is personal perception and thus, for you, this might be true. `instanceof` is a technical feature of Java. In this case we are dealing with domain knowledge ("What signal type I am handling right now?") and thus I prefer using the means of the domain (aka `TYPE` and `getType()`) for solving the problem.

thjaeckle: understood - but in this case I also cast to the type in order to get the "policy" out of the commands - therefore the technical check for instance is also relevant :)

PR #29, opened 2017-11-09
Optimize json conversion of ditto headers

yufei-cai:
`AbstractDittoHeaders.toJson` spends a lot of time in the method `getTypeForKey` iterating over all possible values of `DittoHeaderDefinition`. This commit creates a static index for Ditto header definition values for a faster class lookup during JSON serialization.

### DIFF
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
@@ -154,10 +154,10 @@ public abstract class AbstractDittoHeaders extends AbstractMap<String, String> i
 
     private Class<?> getTypeForKey(final CharSequence key) {
         return getSpecificDefinitionByKey(key)
-                .map(Optional::of)
-                .orElseGet(() -> DittoHeaderDefinition.forKey(key))
                 .map(HeaderDefinition::getJavaType)
-                .orElse(String.class);
+                .orElseGet(() -> DittoHeaderDefinition.forKey(key)
+                        .map(HeaderDefinition::getJavaType)
+                        .orElse(String.class));
     }
 
     @Override
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
@@ -17,10 +17,8 @@ import java.util.AbstractMap;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
```diff
@@ -148,23 +146,11 @@ public JsonObject toJson() {
     }
 
     private Class<?> getTypeForKey(final CharSequence key) {
-        final Collection<HeaderDefinition> definitions = getDefinitions();
-
-        return definitions.parallelStream()
-                .filter(definition -> Objects.equals(definition.getKey(), key))
-                .map(HeaderDefinition::getJavaType)
-                .findAny()
+        return DittoHeaderDefinition.forKey(key)
```
jufickel-b: Does this solution still work with for example `ImmutableMessageHeaders` which have additional HeaderDefinitions to `DittoHeaderDefinition`? Line 163 of the replaced code did take this into account.

model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
```diff
@@ -88,10 +90,16 @@
...
+     * Map to speed up lookup of header definition by key.
+     */
+    private static final Map<CharSequence, DittoHeaderDefinition> VALUES_BY_KEY = Arrays.stream(values())
+            .collect(Collectors.toMap(DittoHeaderDefinition::getKey, Function.identity()));
+
     private final String key;
     private final Class<?> type;
 
-    private DittoHeaderDefinition(final String theKey, final Class<?> theType) {
+    DittoHeaderDefinition(final String theKey, final Class<?> theType) {
```
jufickel-b: IMHO there is no need to give a constructor of an enum more visibility than `private`. I know that IntelliJ has another opinion.

yufei-cai: enum constructors are always private, the modifier has no effect.

model/base/src/main/java/org/eclipse/ditto/model/base/headers/AbstractDittoHeaders.java
yufei-cai: Indeed, all values of headers not defined in `DittoHeaderDefinitions` are serialized as JSON strings. However, the code behaves that way also before this pull request. I will make a separate issue for it.

yufei-cai: Oops, actually `MessageHeaders` uses `getSpecificDefinitions()` to serialize extra non-string headers. Thanks for pointing it out; I will add a fix.

Issue #41, opened 2017-11-14
Minor fixes for DittoProtocolAdapter


dguggemos:
- fix deserialization of messages with content-type `application/json`
- fix serialization of `null` attributes

PR #44, opened 2017-11-14
 #41 fix message (de)serialization in DittoProtocolAdapter

dguggemos:


model/things/src/test/java/org/eclipse/ditto/model/things/NullAttributesTest.java
```diff
@@ -61,14 +60,26 @@ public void nullAttributesIsJsonNull() {
...
     }
 
+    @Test
+    public void nullAttributesIsJsonObject() {
+        assertThat(underTest.isObject()).isTrue();
+    }
+
+    @Test
+    public void nullAttributesAsJsonObject() {
+        assertThat(underTest.asObject()).isEqualTo(JsonFactory.nullObject());
```
jufickel-b: Wouldn't it rather be ... `isEqualTo(underTest)` or even `isSameAs(underTest)`?

dguggemos: Depends on what we want to test. I just wanted to make sure that `asObject()` returns a null object.

PR #48, opened 2017-11-15
Improve ThingUpdater

yufei-cai:
Bugs fixed:
1. ThingUpdater state was modified in a future.
2. ThingUpdater stayed in initial behavior forever if the first retrieve-cache-entry roundtrip did not complete.

Improvements:
1. Removed retrieve-cache-entry messages. Its response is not handled during synchronization, and it is not necessary during initialization.
2. Streamlined shortcuts for CreateThing and ModifyThing events.
3. Made the special handling of the first TagThing message more explicit. Removed unnecessary actor state checkThingTagOnly.


thjaeckle:
Reviewed by @danielFesenmeyer

PR #50, opened 2017-11-22
Bugfixes for payload handling in "live" messages

thjaeckle:
The serialization/deserialization had errors which caused some message payloads not to be handled correctly.
The streaming actors also were not yet 100% working for live commands/responses + messages.


thjaeckle:
Fix of @dguggemos also LGTM

PR #56, opened 2017-11-28
fixed DittoBsonJSON conversion of special characters

thjaeckle:
which we could no longer use at is was in package "com.mongodb.util"

* added JsonFieldConverter interface which can transform JsonField->JsonField
* added a #toString(JsonFieldConverter) method to JsonValue and implemented accordingly
* added JsonFieldConverter also in JsonFactory#readFrom methods
* rewrote DittoBsonJSON in order to operate on JsonObject and use a MongoDB specific JsonFieldConverter implementation taking care of "." and "$" in Json keys


thjaeckle:
@jufickel-b it would be great if you could review this if you find some time ..

jufickel-b:
Yep, I will review the pull request ASAP.

jufickel-b:
This functionality will be implemented completely different without the need to extend the JSON API.

PR #67, opened 2017-12-12
Bug/policy enforcement json view

yufei-cai:
Fix policy enforcement with non-empty resource key such as `/features/gyroscope.0` accessible from routes such as `/things/<thingId>/features/gyroscope.0`.

Previous behavior:
When filtering JSON objects, policy enforcers assume the input JSON object to be an entire Thing. Users may get empty responses from routes with nonempty resource keys if they do not have READ privilege on the entire Thing.

Current behavior:
Policy enforcers assume that the input JSON is obtained from a Thing by descending along the resource key.


yufei-cai:
Commits in this pull reequest are already in the master branch with different IDs.

Issue #89, opened 2018-01-08
Add Openshift Template for ditto


marziman:
It would be great to have the possibility to make ditto run in openshift/kubernetes. As fas as I can see docker-compose is used. It might makes sense to provide a yaml template for kubernetes/openshift like e.g. Kapua.

marziman:
Just to let you know: I can support to „convert“ it from docker compose notation to openshift yaml with the tool called kompose. But doing it natively in a openshift template might be the preffered approach.

thjaeckle:
Sure, that would be great.
As we are not yet using openshift/kubernetes ourselves we would appreciate some help here, we are noobs in this area 😉

marziman:
Hi @thjaeckle,

I will soon make a PR with k8/OpenShift ressources for ditto. Sorry been a bit busy.
But soon to come ..

PR #90, opened 2018-01-09
Initial Kubernetes configuration (for Minikube).

ghandim:
Could be a good start for issue https://github.com/eclipse/ditto/issues/89.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -1,0 +1,65 @@
+## Eclipse Ditto :: Kubernetes
+
+This folder contains an example yaml files which can be used to start Eclipse Ditto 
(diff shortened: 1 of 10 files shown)


fdbeirao:
@ghandim thank you for this PR. I was looking exactly for how to achieve this, because Konvert wasn't doing such a good job, and I am too new to k8s. This just worked.

Would it be possible to also have the `sandbox` as a Kubernetes configuration? 😄

ghandim:
Hi @marziman,
thanks for reviewing my PR and extend it in a more enterprise may. Was my first experience with K8s.
Kind regards, Michael

PR #93, opened 2018-01-11
Handle unparseable exceptions when building a ThingErrorResponse from JSON

dguggemos:


signals/commands/things/src/main/java/org/eclipse/ditto/signals/commands/things/ThingErrorResponse.java
```diff
@@ -168,7 +170,23 @@ public static ThingErrorResponse fromJson(final ThingErrorRegistry thingErrorReg
...
+            final String errorCode =
+                    payload.getValue(DittoRuntimeException.JsonFields.ERROR_CODE).orElse("unknown:unknown");
+            final String errorMessage =
+                    payload.getValue(DittoRuntimeException.JsonFields.MESSAGE).orElse("An unknown error occurred");
+            final String errorDescription = payload.getValue(DittoRuntimeException.JsonFields.DESCRIPTION).orElse("");
+            exception =
+                    DittoRuntimeException.newBuilder(errorCode,
+                            HttpStatusCode.forInt(status).orElse(HttpStatusCode.INTERNAL_SERVER_ERROR))
+                            .message(errorMessage)
+                            .description(errorDescription)
```
thjaeckle: Any reason why the `dittoHeaders` are not passed into the builder here?
`thingErrorRegistry.parse(payload, dittoHeaders)` also passes the DittoHeaders ..
Should be set I think.

PR #108, opened 2018-01-23
Improve logging of search-updater

danielFesenmeyer:
- make sure that the policy has been loaded before handling ThingEvents (errors caused by missing policy could be recovered by the update mechanism, but made it a bit inefficient and produced confusing log messages)
- add debug logs to MongoThingsSearchUpdaterPersistence
- log an INFO whenever a stream (i.e. sync) is scheduled


danielFesenmeyer:
Authored by @ffendt

PR #158, opened 2018-05-16
NullPointerException trying to retrieve things without ids-parameter

danielFesenmeyer:
A NullPointerException occured when a user tried to retrieve things with an empty ids-parameter, e.g. by running the following REST request:
`GET '/api/2/things?ids=`

The user got status code 503 as a consequence. This PR fixes this by returning status code 400 with an appropriate error message instead.

In this context, we also changed the way missing parameters are handled (for ditto in general). If the user did not specify the ids-parameter at all, she got a 404 status code (akka default for _MissingQueryParamRejection_). We changed this to a 400 response, because we consider such a request as bad input, so "Bad Request" is more helpful than "Not Found".

### DIFF
--- a/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/utils/DittoRejectionHandlerFactory.java
+++ b/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/utils/DittoRejectionHandlerFactory.java
@@ -13,24 +13,28 @@ package org.eclipse.ditto.services.gateway.endpoints.utils;
 
 import static akka.http.javadsl.server.Directives.complete;
 
+import java.text.MessageFormat;
+
 import akka.http.javadsl.model.StatusCodes;
 import akka.http.javadsl.server.MissingQueryParamRejection;
 import akka.http.javadsl.server.RejectionHandler;
 import akka.http.javadsl.server.Route;
 
 /**
- * Factory to create a custom {@link akka.http.javadsl.server.RejectionHandler}
+ * Factory to create a custom {@link akka.http.javadsl.server.RejectionHandler}.
  */
 public final class DittoRejectionHandlerFactory {
 
-    private static final String MISSING_QUERY_PARAM_TEMPLATE = "Request is missing parameter: '%s'";
+    private static final String MISSING_QUERY_PARAM_TEMPLATE = "Request is missing required query parameter '{0}'";
 
-    private DittoRejectionHandlerFactory() {}
+    private DittoRejectionHandlerFactory() {
+        throw new AssertionError();
+    }
 
     /**
-     * Creates a new instance of {@link akka.http.javadsl.server.RejectionHandler} with custom behaviour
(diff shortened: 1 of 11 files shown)


services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/utils/DittoRejectionHandlerFactory.java
```diff
@@ -0,0 +1,51 @@
...
+import akka.http.javadsl.server.MissingQueryParamRejection;
+import akka.http.javadsl.server.RejectionHandler;
+import akka.http.javadsl.server.Route;
+
+/**
+ * Factory to create a custom {@link akka.http.javadsl.server.RejectionHandler}.
+ */
+public final class DittoRejectionHandlerFactory {
+
+    private static final String MISSING_QUERY_PARAM_TEMPLATE = "Request is missing required query parameter '{0}'";
```
thjaeckle: I think that this won't work in a MessageFormat. AFAIR when you want a `'` in a string formatted by MessageFormat, you have to write it like this: `''{0}''`

signals/commands/things/src/main/java/org/eclipse/ditto/signals/commands/things/exceptions/MissingThingIdsException.java
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.ditto.model.base.exceptions.DittoRuntimeException;
+import org.eclipse.ditto.model.base.exceptions.DittoRuntimeExceptionBuilder;
+import org.eclipse.ditto.model.base.headers.DittoHeaders;
+import org.eclipse.ditto.model.things.ThingException;
+
+/**
+ * This exception is thrown when multiple things are requested without specifying IDs.
+ */
+@Immutable
+public class MissingThingIdsException extends DittoRuntimeException implements ThingException {
```
thjaeckle: There is no generated `serialVersionUid` present in the RuntimeException. Probably just cosmetics, but RuntimeException is Java `Serializable`

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/utils/DittoRejectionHandlerFactory.java
danielFesenmeyer: Thanks for the finding. I fixed this with commit 293b7a107e76c7684b4d19fd662ff35400fe8822.

signals/commands/things/src/main/java/org/eclipse/ditto/signals/commands/things/exceptions/MissingThingIdsException.java
danielFesenmeyer: Thanks for the hint, fixed with commit 6c7f5318025f12948974c92729e9a0ed18a0809b.

PR #165, opened 2018-05-24
Remove unused import

yufei-cai:
Self-explanatory.


yufei-cai:
Closing because travis fails repeatedly for different reasons each time. Change isn't important anyway.

PR #171, opened 2018-06-06
Avoid NullPointerException caused by missing DittoHeaders.

jufickel-b:
Signed-off-by: Juergen Fickel <juergen.fickel@bosch-si.com>


yufei-cai:
Looks good to me.

PR #172, opened 2018-06-06
Bugfix/npe policypersistenceactor

jufickel-b:
PolicyPersistenceActor threw NullPointerException asking whether its own policy exists when it has nothing in the event journal or the snapshot store.


yufei-cai:
Good job with the formatting pass.

yufei-cai:
Added retry mechanism in actor-recreation tests to stabilize Travis build.

PR #175, opened 2018-06-11
Stabilize background search index synchronization

yufei-cai:
- Removed a stuck state of ThingUpdater: If MongoDB went down during `awaitSyncResultBehavior` then ThingUpdater became stuck until killed. The error was caused by [NullPointerException when unboxing](https://stackoverflow.com/q/2382058) a null reference of type Boolean.

- Added activity check for stream supervisor: If it does not interact with stream forwarder for a long time then it is likely stuck and should restart.

- Prevented LastSuccessfulStreamCheckingActor from reporting ERROR within a time window from startup so that the cluster does not become unhealthy immediately.


ffendt:
LGTM

PR #176, opened 2018-06-11
Enhance extensibility of Search service

ffendt:
All services that are part of Eclipse Ditto are built from a simple abstract class to allow enhancing their functionality. The Search service does not provide this abstraction and thus lacks extensibility in comparison to the other services. This PR solves this issue and supports extending an abstract implementation of the service.


yufei-cai:
LGTM

PR #177, opened 2018-06-11
Make props of AMQP client actors serializable

yufei-cai:
AMQP client actors are started by ClusterRouterPool possibly on a remote node. Their props should be serializable.

This patch moved the nonserializable connection factory arguments inside actor constructor so that each nodes uses connection factories on their own class path. Connection factories have no connection-specific information to be exchanged between nodes anyway.

Unit tests use the Akka-internal class `DaemonMsgCreate` because the serializer for `Props` is the Java serializer, which does not invoke our JSON serializer on elements of `Props.args()`.


ffendt:
LGTM

PR #185, opened 2018-07-09
Fix activity check in Policy- and ThingPersistenceActor

yufei-cai:
- Activity check should not be ignored by nonexistent things;
  it could lead to memory leak.

- Fixed concurrency issue in unit tests: Activity checks cause
  persistence actors to terminate parents, after which they
  cannot process any message. Tests now wait for actor
  termination instead of any reply from persistence actors.


dguggemos:
LGTM  👍

PR #191, opened 2018-07-20
fixed bug in top-level PUT /thing/{thing-id} merging existing fields

thjaeckle:
* added test which failed for v1 on existing codebase
* fixed ThingPersistenceActor to also merge v1 correctly
* documented behavior in OpenAPI descriptors


yufei-cai:
LGTM

PR #195, opened 2018-07-24
Header markers

yufei-cai:
Added 2 flags to Ditto header definitions to signify whether they should be read from Ditto protocol messages or written to Ditto protocol messages. It is then possible to prevent sending of duplicate/sensitive information in protocol messages and tampering by untrusted clients.


protocol-adapter/src/main/java/org/eclipse/ditto/protocoladapter/HeaderTranslator.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Read Ditto headers from external headers.
+     *
+     * @param externalHeaders external headers as a map.
+     * @return Ditto headers initialized with values from external headers.
+     */
+    public DittoHeaders fromExternalHeaders(final Map<String, String> externalHeaders) {
+        final DittoHeadersBuilder builder = DittoHeaders.newBuilder();
+        externalHeaders.forEach((key, value) -> {
+            final HeaderDefinition definition = headerDefinitionMap.get(key);
```
thjaeckle: Case sensitivity plays a role here - if a "Authorization" key is in the `externalHeaders`, a "authorization" in the `headerDefinitionMap` won't match

thjaeckle: We should also add a test for case sensitivity

PR #201, opened 2018-07-30
Akka Management - Kubernetes API support

stmaute:
Ditto now supports the kubernetes-api discovery mechanism from the [akka-management](https://github.com/akka/akka-management) project. 

An example on how to run Ditto with Kubernetes is provided in the _/kubernetes_ directory. 
With this PR #90 is obsolete.

Fixes: #89


marziman:
hi @stmaute 
just took a quick look.
Do you need any support, I can review and test the contribution (I was planning to do k8s stuff for ditto, but was too busy).
Let me know if you want me to put that on my cluster..

BR Mehmet

thjaeckle:
It would be great if you could try it in an OpenShift cluster - we use `minikube` locally and a self managed Kubernetes instead of OpenShift.
But you would have to build the docker images by yourself as long as we don't release this in a next milestone release (which we want to do soon).

thjaeckle:
Fixes: #89

marziman:
@thjaeckle,

is this still relevant?
I was redering to habing a real Ditto K8s support, not to minikube. Minikube cant be really used in a more production grade setup. Let me know if I can support and we maybe open new Issues for that acivities to track them.

WDYT?

BR mehmet

thjaeckle:
Hi Mehmet.

Ditto has a real k8s support - we are using it in our commercial service on AWS EKS.
We however did not try it on OpenShift as we simply have no experience with it and no real need to do so.

So if you could test getting Ditto running on OpenShift (or find out what would still be missing) would be a great help for us.

PR #203, opened 2018-08-03
updated jjwt library to 0.10.1

thjaeckle:
* implemented own jjwt serializer and deserializer using minimal-json in order to get rid of jackson dependency


jufickel-b:
Was using ditto-json an option? Why using `Assert.notNull(bytes, "JSON byte array cannot be null")` instead of `ConditionChecker.checkNotNull()`?

thjaeckle:
Sure, using ditto-json would also work - but I didn't see a benefit vs. using minimal-json.
Did I miss something?
I was using the `Assert.notNull(bytes, "JSON byte array cannot be null")` as this is provided by the `jjwt` library itself, but we can also use our own.

jufickel-b:
Using ditto-json here would not make a difference in the outcome. It would, however, channel JSON library dependencies to one place (ditto-json).

Using Ditto's `ConditionChecker`, too, would not affect the outcome but why not eat your own dog food?

thjaeckle:
sure, that is one approach - another approach is to minimize dependencies wherever possible ;)
But I don't have strong feelings in keeping it that way, so I will follow your suggestion :)

services/utils/jwt/src/main/java/org/eclipse/ditto/services/utils/jwt/JjwtDeserializer.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    private static Object parse(final String json) {
+
+        return toJavaObject(JsonFactory.readFrom(json));
+    }
+
+    private static Map<String, Object> toJavaMap(final JsonObject jsonObject) {
+        return StreamSupport.stream(jsonObject.spliterator(), false)
```
jufickel-b: JsonObject directly provides `stream` method.

services/utils/jwt/src/main/java/org/eclipse/ditto/services/utils/jwt/JjwtDeserializer.java
```diff
@@ -0,0 +1,116 @@
...
+                .collect(Collectors.toMap(
+                        JsonField::getKeyName,
+                        field -> toJavaObject(field.getValue())));
+    }
+
+    private static List<Object> toJavaList(final JsonArray jsonArray) {
+        return StreamSupport.stream(jsonArray.spliterator(), false)
```
jufickel-b: JsonArray directly provides `stream` method.

thjaeckle: ah, the benefit of `ditto-json` :)

services/utils/jwt/src/main/java/org/eclipse/ditto/services/utils/jwt/JjwtSerializer.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private static JsonObject toJsonObject(final Map<?, ?> map) {
+
+        final JsonObjectBuilder objectBuilder = JsonObject.newBuilder();
+        map.forEach((key, value) -> objectBuilder.set(String.valueOf(key), toJson(value)));
+        return objectBuilder.build();
+    }
+
+    private static JsonArray toJsonArray(final Collection<?> c) {
+
```
jufickel-b: Maybe use `JsonCollectors.valuesToArray()` here?

services/utils/jwt/src/main/java/org/eclipse/ditto/services/utils/jwt/JjwtDeserializer.java
jufickel-b: :-)

PR #206, opened 2018-08-10
block the main() thread until the ActorSystem terminates so that OpenJ9

thjaeckle:
correctly invokes shutdown hooks on a SIGTERM


thjaeckle:
Workaround for https://github.com/eclipse/openj9/issues/2571

PR #226, opened 2018-09-05
Feature/conditional headers

Yannic92:
This PR adds support for the following precondition headers:

* If-Match
* If-None-Match

For things and policies including all their subresources.
When requesting those resources there will now be an ETag-Header in the response.


danielFesenmeyer:
PR looks good. The commits have all already been reviewed.

PR #229, opened 2018-09-06
physically delete Things from search index after configurable age

thjaeckle:
* up to now, orphaned search index entries in collection 'thingEntities' were never deleted
* this can now be enabled (and is by default) via search configuration


ffendt:
LGTM

Issue #228, opened 2018-09-06
Reconnect not working correctly for AMQP 0.9.1 Connections


ffendt:
When playing around with AMQP 0.9.1 Connections, i found a bug in the reconnect mechanisms of Ditto.

If the AMQP broker (to which a connection is pointing) is unreachable for a while, Ditto will try to reestablish the connection and finally succeed when the broker is reachable again. 
However, after the connection is established, consuming or publishing won't work for the connection until the service is restarted.

jufickel-b:
Closed with PR https://github.com/eclipse/ditto/pull/230.

PR #230, opened 2018-09-07
Bugfix for AMQP 0.9.1 reconnected state

ffendt:
This PR fixes Issue #228 which describes the problem that AMQP 0.9.1 connections can't consume or publish if the connection had to reconnect to the broker.

The library used to establish the connections to AMQP 0.9.1 provides failover and notifies us after a reconnect with the new created channels. With this PR, the actors responsible for consuming and publishing will use the new channels.

While trying to find the issue itself, i realized that we should have some more logging which is why i also added some more log messages.


dguggemos:
LGTM 👍

Issue #149, opened 2018-05-02
Enhance subscribing for change notifications by optional filter


thjaeckle:
When currently subscribing for Events/change notifications via

* WebSocket
* SSE
* Connectivity

the consumer always gets all change notifications it is allowed to see.

On SSE this can be reduced by providing specific `thingId`s and `fields` so that changes are only published if the `thingId` match or a change was affected by a specified `field`.

The subscription should and could be more fine-grained.
The idea is to support adding an optional `filter` defined via Ditto's RQL syntax it already uses [for the search](https://eclipse.org/ditto/intro-overview.html).

That way the following subscription rules could be applied:
* notify me of a change only if the `thingId` starts with `org.eclipse.ditto:*` : `like(thingId,"org.eclipse.ditto:*")`
* notify me of a change only if the feature `temperature` was affected by this change: `exists(feature/temperature)`
* notify me of a change only if in one change the `temperature` was greater 25: `gt(feature/temperature/properties/value,25)`
* notify me of changes in the namespace `org.eclipse.ditto:*` affecting the `temperature`: `and(like(thingId,"org.eclipse.ditto:*"),exists(feature/temperature))`

PR #232, opened 2018-09-11
Use namespaces and rql filter for event subscriptions

thjaeckle:
Fixes: #149

When subscribing for events at all Ditto streaming APIs (SSE, WebSocket, connectivity) optionally following parameters can now be specified:
* `namespaces`: a comma separated list of namespaces to filter (so e.g. only events from those namespaces are emitted)
* `filter`: a RQL filter of which events to emit (as described in #149)


thjaeckle:
Pair-Reviewed together with @phalski

Issue #220, opened 2018-08-24
Add MQTT support


yufei-cai:
MQTT is a super popular IoT protocol, Ditto should definitely support it!

Interoperability with [Amazon IoT](https://aws.amazon.com/iot/) weren't half bad either.

thjaeckle:
Can this be closed? Should be fixed by #224

yufei-cai:
#224 does not interoperate with Amazon IoT due to lack of client certificate authentication.

yufei-cai:
Fixed by #235.

PR #235, opened 2018-09-20
Add support for client certificate authentication to MQTT connectivity

dguggemos:
This PR is part of #220 and adds the possibility to use client certificates to authenticate against an MQTT broker. Moreover this also enables to upload CA certificates for AMQP1.0/0.9.1 connections.


yufei-cai:
Fixes #220.

PR #266, opened 2018-10-19
encode surplus @-signs in a connection uri. 

ffendt:
Currently i can not connect to  connections with a connection URI that contains an `@`-sign in its credentials without encoding the sign manually. Since e.g. Eclipse Hono requires the user in the URI to have the format `user@tenant`, the `@`-sign gets encoded automatically with this PR.


thjaeckle:
Is it sufficient to only handle the `@` sign?
What about other characters which are not allowed in URIs (e.g. `#`)?

PR #268, opened 2018-10-24
Allow to create a new thing that uses a copied policy

Yannic92:
Adds support to copy a policy from a thing reference or an explicit policy id


ffendt:
LGTM

PR #271, opened 2018-10-26
Reslove duplicate ERROR_CODE in common error registry

stmaute:
Extend de-serialization of all DittoRuntimeExceptions by de-serializing all properties of a Exception.
Now the message, description and hRef are de-serialized correctly.

### DIFF
--- a/json/src/main/java/org/eclipse/ditto/json/JsonParseException.java
+++ b/json/src/main/java/org/eclipse/ditto/json/JsonParseException.java
@@ -23,7 +23,7 @@ public final class JsonParseException extends JsonRuntimeException {
     /**
      * Error code of this exception.
      */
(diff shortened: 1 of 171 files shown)


json/src/main/java/org/eclipse/ditto/json/JsonParseException.java
```diff
@@ -23,10 +23,10 @@
     /**
      * Error code of this exception.
      */
-    public static final String ERROR_CODE = "json.invalid";
+    public static final String ERROR_CODE = "json.format.invalid";
```
thjaeckle: Hm, did you break the error code on purpose?
Might be that users see this as API and assume that exact string.

ffendt:
LGTM

PR #272, opened 2018-10-31
Dont set source address to source header

Yannic92:
* Removes setting of source address in AmqpConsumerActor and RabitMQConsumorActor
* Set Source header in Message MappingProcessActor instead


dguggemos:
LGTM 👍

PR #278, opened 2018-11-09
remove fallback queries on policiesBasedSearchIndex

ffendt:
removed fallback queries that assumed that `_thingId` might be missing in `policiesBasedSearchIndex` collection.

Signed-off-by: Florian Fendt <Florian.Fendt@bosch-si.com>


Yannic92:
LGTM @thjaeckle could you merge this?

PR #285, opened 2018-11-13
fixed retrieving shard region stats and aggregation

thjaeckle:
* on HTTP route /stats/things/details
* was affected by max cluster size limits and Java serialization of response messages
* added RetrieveStatisticsDetailsResponseSupplier which is used in the services in order to calculate the "hot entities" per namespace

### DIFF
--- a/services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/StatisticsActor.java
+++ b/services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/StatisticsActor.java
@@ -282,15 +282,10 @@ public final class StatisticsActor extends AbstractActor {
                         retrieveStatisticsDetailsResponse.getStatisticsDetails().getValue(shardRegion)
                                 .map(JsonValue::asObject)
                                 .map(JsonObject::stream)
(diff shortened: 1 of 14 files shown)


yufei-cai:
Looks good to me.

services/gateway/proxy/src/main/java/org/eclipse/ditto/services/gateway/proxy/actors/StatisticsActor.java
```diff
@@ -282,15 +282,10 @@ private void becomeStatisticsDetailsAwaiting(final Consumer<StatisticsDetails> s
...
-                                        shardStatistics.hotnessMap.put(field.getKeyName(),
-                                                shardStatistics.hotnessMap.get(field.getKeyName()) +
-                                                        field.getValue().asLong());
-                                    } else {
-                                        shardStatistics.hotnessMap.put(field.getKeyName(), field.getValue().asLong());
-                                    }
-                                }));
+                                .ifPresent(namespaceEntries -> namespaceEntries.forEach(field ->
+                                        shardStatistics.hotnessMap
+                                                .merge(field.getKeyName(), field.getValue().asLong(), Long::sum)
```
thjaeckle: :+1:

PR #293, opened 2018-11-22
fixed potential NPE in RootRoute and logged warning

thjaeckle:
in order to identify the causing DittoRuntimeException


yufei-cai:
Looks good to me.

PR #311, opened 2019-01-10
use OSS akka cluster downer

thjaeckle:
Replace own implementation in `ClusterMemberAwareActor` with `simple-akka-downing` dependency taking care of split-brain/majority/minority situations in the cluster based on the cluster role.


thjaeckle:
Already added CQ for the new depdendency: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18601

PR #313, opened 2019-01-14
Replace Dropwizard with Kamon for MongoDB Metrics

Yannic92:


bom/pom.xml
```diff
@@ -437,27 +449,6 @@
...
-            </dependency>
-            <dependency>
-                <groupId>io.prometheus</groupId>
-                <artifactId>simpleclient_dropwizard</artifactId>
-                <version>${prometheus.java-client.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.prometheus</groupId>
-                <artifactId>simpleclient_common</artifactId>
-                <version>${prometheus.java-client.version}</version>
```
thjaeckle: We can then also remove the maven property `prometheus.java-client.version`, right?

bom/pom.xml
```diff
@@ -437,27 +449,6 @@
                     </exclusion>
                 </exclusions>
             </dependency>
-            <dependency>
-                <groupId>io.dropwizard.metrics</groupId>
-                <artifactId>metrics-core</artifactId>
-                <version>${metrics.version}</version>
```
thjaeckle: Can we remove the property `metrics.version` then?

PR #316, opened 2019-01-16
Dont allow empty thing names

Yannic92:
Signed-off-by: Klem Yannic (INST/ECS1) <Yannic.Klem@bosch-si.com>


thjaeckle:
We should also adjust the Policy regex accordingly

PR #319, opened 2019-01-21
Bugfix/validate policy id of thing

Yannic92:


model/things/src/main/java/org/eclipse/ditto/model/things/ThingPolicyIdInvalidException.java
```diff
@@ -0,0 +1,140 @@
...
+ * Thrown if the ID set as policyId of the {@link Thing} is not valid
+ * (for example if it does not comply to the Entity ID REGEX).
+ */
+@Immutable
+public final class ThingPolicyIdInvalidException extends DittoRuntimeException implements ThingException {
+
+    /**
+     * Error code of this exception.
+     */
+    public static final String ERROR_CODE = ERROR_CODE_PREFIX + "policyId.invalid";
```
thjaeckle: ah, I just saw that the error code is camel cased here - AFAIR we never camel cased error codes - I would rather use `"policyid.invalid"` (similar to the `FeatureIdInvalidException`)

Yannic92: Oh good catch, sorry. Fixed it

stmaute:
LGTM :+1:

PR #327, opened 2019-02-04
Provide helm chart for Eclipse Ditto

ghandim:
Signed-off-by: Michael Gantert <michael.gantert@bosch-si.com>


kaizimmerm:
Hi, 

I can confirm that this works on AKS as well. 

As soon as you have this on master I can offer to contribute a few more flexibility options, i.e. service type, namespace support etc.

Kai

kaizimmerm:
@thjaeckle do you have already an idea when this will be on master?

Yannic92:
It's currently in review and some changes are required. Can't give you an exact date.

deployment/helm/eclipse-ditto/files/openapi/ditto-api-1.yml
```diff
@@ -0,0 +1,3060 @@
+# Copyright (c) 2017-2018 Bosch Software Innovations GmbH.
```
Yannic92: I think we shouldn't duplicate the swagger documentation. How about a symbolic link?

deployment/helm/eclipse-ditto/files/openapi/ditto-api-2.yml
```diff
@@ -0,0 +1,4134 @@
+# Copyright (c) 2017-2018 Bosch Software Innovations GmbH.
```
Yannic92: I think we shouldn't duplicate the swagger documentation. How about a symbolic link?

kaizimmerm:
@Yannic92 great! I have actually quite a few changes that I can PR as soon as you have this on master. So far:

- Support for namespace deployment.
- Support for other service types then `NodePort`.
- Support for external MongoDB including SSL support.
- Added [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).
- Support for other Docker repositories or tags.

deployment/helm/eclipse-ditto/files/openapi/ditto-api-1.yml
ghandim: Thanks for the hint. Prepare an additional commit on this PR.

deployment/helm/eclipse-ditto/files/openapi/ditto-api-2.yml
ghandim: Thanks for the hint. Prepare an additional commit on this PR.

Issue #317, opened 2019-01-18
Collect connectivity metrics and provide via devops command


thjaeckle:
Goal: Provide information about
- how many messages have been consumed (e.g. from a MQTT or AMQP endpoint)?
- how many of those messages were successfully processed / how many failed?
- how many times was the payload mapping successful?
- how many messages were dropped after the payload mapping?

Do that for "inbound" messages (e.g. commands, messages) as well as "outbound" (e.g. events, responses).

PR #329, opened 2019-02-05
Provide advanced connection status and metrics for Ditto managed connections

thjaeckle:
Resolves: #317

The format of connection status and metrics did change with this PR, but is now way more expressive.


thjaeckle:
@ffendt just had an idea that we probably should handle responses with missing 'reply-to' information as 'dropped failure'.
Because in that case the command did not set 'response-required false', so expected a response but did not provide information about where to send it to.

Could you adjust that before merging the pr?

ffendt:
@thjaeckle but if the user expects a response and we can't send the response because the reply-to information is missing, shouldn't it be a 'published failure' instead of a 'dropped failure'?

thjaeckle:
Published failure is used for technical problems when eg there is no amqp subscriber ready to receive the message.
And since dropped failure is currently not used at all, this would be a better fit in order to find out what went wrong.

ffendt:
I agree, but it is also kind of misleading, since there is no failure while dropping the message, but the message is dropped because of a 'failure' (misconfiguration).
Since there is no better fit to an existing category i will add it to the 'dropped failure' category.

PR #332, opened 2019-02-13
Kubernetes deployment

kaizimmerm:
Hi,

while using your Kubernetes deployment I found a few things:

- MongoDB is on default bind_ip which means it will only accept connections on 127.0.0.1 interface. At least on my local as well as cloud setup that ain't working.
- Kubernetes compared to the ditto docker swarm setup does not provide devOps and health interfaces. Also the header config was a bit different compared to swarm for status.
- My editor prettified the indentation a bit. I hope you do not mind :)

I thought you might be interested in my changes. Tested on Azure Kubernetes Service as well as Docker Desktop for Mac.

Another bigger issue I had is that the Kubernetes stuff is a bit inflexible to work with. I would volunteer to create a Helm setup, let me know if you are interested and if yes if this should go extra or replace the Kubernetes deployment you already have. I would go for option two but let me know what you think.

Kai

### DIFF
--- a/deployment/kubernetes/mongodb/mongodb.yaml
+++ b/deployment/kubernetes/mongodb/mongodb.yaml
@@ -5,9 +5,9 @@ metadata:
 spec:
   clusterIP: None
   ports:
-  - port: 27017
-    targetPort: 27017
-    protocol: TCP
+    - port: 27017
+      targetPort: 27017
+      protocol: TCP
   selector:
     app: mongodb
 ---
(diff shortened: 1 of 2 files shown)


thjaeckle:
Hi Kai.

Thanks for your PR :)
We will have a look at it.

Regarding the helm setup: We already have a PR open for that: #327 
It would be awesome if you could have a look on that / verify that this is working - we did not yet find the time to validate that in the development team ourselves.

thjaeckle:
@kaizimmerm do you still have a valid Eclipse ECA? Seems IP-Validation somehow failed

kaizimmerm:
@thjaeckle I have, but it is broken it seems.

kaizimmerm:
I think you can verify this here https://www.eclipse.org/contribute/cla

thjaeckle:
Ah, indeed. Thx

PR #333, opened 2019-02-13
[WIP] Range-based paging for things-search

yufei-cai:
I'd like to add range-based paging for things-search so that Ditto supports paging through large numbers of things without putting excessive memory pressure on the persistence.

- Added a parameter "nextPageKey" to encode position of a search.
- TODO: Interpret "nextPageKey" in things-search so that paging is possible without large skip values.


thjaeckle:
Please ignore the Codacy check - I tried that out and it did "review" all open PRs

yufei-cai:
Closing this because an alternative implementation is under way.

PR #351, opened 2019-03-08
Add support for placeholder functions inside connections

### LINKED ISSUE #337: Provide placeholder functions for connectivity placeholder mechanism
In the connectivity microservice Ditto already makes use of placeholders, e.g. in order to determine a "target address" based on variables, e.g.:
```
{{ thing:id }}
{{ topic:full }}
{{ header:some-arbitrary-header }}
```

It would be very useful to also apply functions on these placeholders, e.g. in order to just select a substring of a variable, e.g.:
```
{{ thing:name | fn:substring-before('_') }}
```

Same applies for header mappings where placeholders may be used.

Current ideas for functions:
* substring-before
* substring-after
* default - provide a fallback value if for example a header was not set

thjaeckle:
Fixes #337


thjaeckle:
@joosdavid please sign the new Eclipse ECA

thjaeckle:
As @joosdavid has signed the ECA/CLA, going to proceed with merge

PR #360, opened 2019-03-19
Azure deployment guide and templates

kaizimmerm:
Hi,

I created a guide and a set of templates to support Ditto deployment on Microsoft Azure.

Let me know what you think or even getter give it a try 😄 

Kai

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ dependencies.txt
 ajcore.*.txt
 *.mdb
 .env
+.DS_Store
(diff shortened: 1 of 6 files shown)


ghandim:
> Hi,
> I created a guide and a set of templates to support Ditto deployment on Microsoft Azure.
> Let me know what you think or even getter give it a try smile
> Kai

@kaizimmerm I will give it a try in my Azure environment.

ghandim:
@kaizimmerm Worked like a charm in my subscription.

PR #364, opened 2019-03-27
Enable support for MongoDB persistence in K8s

kaizimmerm:
Hi,

small but practical improvement is to use persistent storage in the K8s setup for the MongoDB. 

So I:

- Migrated to the standard `stable/mongodb` chart which has this feature (and many more) already included.
- Adapt the helm guide to the new chart.
- Extended the Azure guide to leverage this using Azure storage.

Let me know what you think,

Kai


ghandim:
@kaizimmerm Why do you not connect a CosmosDB instance for a persistent storage example for ditto on Azure?

kaizimmerm:
@ghandim I intend to look into this soon.

However, I would like to keep the MongoDB embedded option for test/dev purposes and especially there it is really annoying if you lose your data all the time. 

Besides this patch will work in other K8s environments that have storage support as well. I hear there are other clouds next on Azure in existence, all balderdash of course 😈 

Kai

ghandim:
@kaizimmerm Good point :-) I will give it a try...

deployment/azure/README.md
```diff
@@ -44,19 +44,36 @@ aks_cluster_name+="dittoaks"
...
 
 ```bash
 k8s_namespace=dittons
 
 kubectl create namespace $k8s_namespace
+```
+
+...either with persistent storage
+
+```bash
```
ghandim: @kaizimmerm I think you have to add here `helm ../helm/eclipse-ditto/ dependency update`. Otherwise you get a weird message from helm...

deployment/azure/README.md
```diff
@@ -44,19 +44,36 @@ aks_cluster_name+="dittoaks"
...
+helm upgrade ditto ../helm/eclipse-ditto/ --namespace $k8s_namespace --set service.type=LoadBalancer,mongodb.persistence.enabled=true,mongodb.persistence.storageClass=managed-premium-retain --wait --install
```
ghandim: @kaizimmerm Whats the advantage to use an own storage class definition instead of the default one? If I spin up ditto with `helm upgrade ditto ../helm/eclipse-ditto/ --namespace $k8s_namespace --set service.type=LoadBalancer,mongodb.persistence.enabled=true --wait --install` K8s generates a default persistent volume with from my point of view the same configuration as configured in `managed-premium-retain`.

ghandim: @kaizimmerm Now I got it. The important difference is the `retain` police..

ghandim:
@kaizimmerm I think if we want to support the preservation of the mongodb data between complete reinstallations of ditto e.g. with `helm delete --purge ditto` than we have to introduce a vpc like
`apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: mypvc
  namespace: dittons
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi
  storageClassName: managed-premium-retain`

Then we can upgrade/install ditto without any data loss: `helm upgrade ditto ../helm/eclipse-ditto/ --namespace dittons --set service.type=LoadBalancer,mongodb.persistence.enabled=true,mongodb.persistence.existingClaim=mypvc --wait --install`

What do you think?

thjaeckle:
Please ignore the Codacy check - I tried that out and it did "review" all open PRs

kaizimmerm:
@ghandim yes, I guess that would work. I could add a paragraph on this to the helm guide. Agreed?

ghandim:
@kaizimmerm With this pvc I was able to create a persistent reusable (default) storage in Azure AKS/Minikube: 
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ditto-pvc
  namespace: dittons
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi
#  storageClassName: <YOUR_STORAGE_CLASS> 
```
To use the storage on Minikube in the namespace dittons you have to create ditto with 
```
helm upgrade ditto ./eclipse-ditto/ --namespace dittons --set mongodb.persistence.enabled=true,mongodb.persistence.existingClaim=ditto-pvc --wait --install
```
On Azure:
```
helm upgrade ditto ../helm/eclipse-ditto/ --namespace dittons --set service.type=LoadBalancer,mongodb.persistence.enabled=true,mongodb.persistence.existingClaim=ditto-vpc --wait --install
```
So I think we should align also the helm documentation and use a separate namespace dittons and move the ditto-pvc to helm. Then also for the developers it is more clear that there are not many differences between the deployments on different environments.
What do you think?

kaizimmerm:
@ghandim In Azure (and I assume other K8s setups as well) there are already PVCs pre setup. So I would make this an optional step in the helm guide.

kaizimmerm:
@ghandim forget it, I mixed something up 😄 So I agree, I will adapt accordingly

kaizimmerm:
@ghandim added it, let me know what you think

deployment/helm/README.md
```diff
@@ -44,8 +44,32 @@ helm init
 
 ```bash
 cd <DITTO_PATH>/deployment/helm/
+kubectl create namespace dittons
 helm update dependencies
```
ghandim: Path is missing here: helm dependencies update ./eclipse-ditto/

kaizimmerm: Fixed, good catch!

deployment/helm/ditto-mongodb-pvc.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: ditto-pvc
```
ghandim: @kaizimmerm Should we name the pvc also ditto-mongodb-pvc?

kaizimmerm: fixed

deployment/azure/README.md
```diff
@@ -44,19 +44,50 @@ aks_cluster_name+="dittoaks"
...
+```
+
+...or with a custom K8s [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) independent of the Helm release to ensure the data survives a helm delete:
+
+```bash
+echo "  storageClassName: managed-premium-retain" >> ../helm/ditto-mongodb-pvc.yaml
+kubectl apply -f ../helm/ditto-mongodb-pvc.yaml --namespace $k8s_namespace
+helm upgrade ditto ../helm/eclipse-ditto/ --namespace $k8s_namespace --set service.type=LoadBalancer,mongodb.persistence.enabled=true,mongodb.persistence.existingClaim=ditto-pvc --wait --install
```
ghandim: ditto-mongodb-pvc has now to be used here also for the existingClaim ;-)

kaizimmerm: done 😄

ghandim:
@kaizimmerm One last thing....
Are there any concerns to update the ditto tag to the current milestone 0.9.0-M1?

kaizimmerm:
@ghandim  sure, will do that

ghandim:
@kaizimmerm Last but not least I would reduce the values in Charts.yaml to 
```
apiVersion: v1
name: eclipse-ditto
version: 0.9.0-M1
description: A basic helm chart for Eclipse Ditto and K8s
```
What do you think?

ghandim:
@kaizimmerm LGTM :thumbsup: @thjaeckle Nothing more to add from my point of view at the moment.

thjaeckle:
Awesome, thank you both for the effort :+1:

Issue #366, opened 2019-03-28
Sending responses to Azure Service Bus via AMQP 1.0 fails after some time


thjaeckle:
When connecting to [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/) via AMQP 1.0 from Ditto's connectivity service, after some time the Azure Service Bus seems to close the AMQP link (the connection stays open) or seems to no longer accept new links.
From that point on no responses can be sent to that AMQP address.

This was observed for *responses*. The *commands* are still received via another AMQP address - a "reply-to" address was specified for those commands via AMQP header, to which the response is sent.

Following log occurs in the log:
```
org.eclipse.ditto.services.connectivity.messaging.amqp.AmqpPublisherActor: Failed to send JMS message: [IllegalStateException] The MessageProducer was closed due to an unrecoverable error.
```

Maybe the Ditto connectivity creates too much links (e.g. when each "reply-to" is a new address, e.g. a UUID) so that Azure Service Bus prevents adding new ones.

thjaeckle:
Fixed by #367

PR #367, opened 2019-03-28
Recover from closed JMS AMQP message producer

kaizimmerm:
Hi,

ditto is currently caching the MessageProducers in its AMQP 1.0 connector. However, producers might be closed by the server. In fact Azure Service Bus is doing so after 10 minutes of idle time. Other servers might have something similar.

I think there are multiple approaches to solve this:
* Put a static invalidation on the cache, e.g. 5 minutes.
* Recover from the IllegalStateException the closed producer is throwing when attempting to send through it.
* React to the clients connection listener which actually informs about a closed producer.

I decided to go for the later option as it is the cleanest one imho. However, for maximum robustness it might make sense to implement all three.

Be aware this is my first code contribution to Ditto or any Akka based app at all, so please let me know if I got it completely wrong here 😈 

Kai

Signed-off-by: Kai Zimmermann <kai.zimmermann@microsoft.com>

### DIFF
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpPublisherActor.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpPublisherActor.java
@@ -30,7 +30,6 @@ import javax.jms.JMSException;
 import javax.jms.Message;
 import javax.jms.MessageProducer;
 import javax.jms.Session;
-
 import org.apache.qpid.jms.JmsQueue;
 import org.apache.qpid.jms.message.JmsMessage;
 import org.apache.qpid.jms.message.facade.JmsMessageFacade;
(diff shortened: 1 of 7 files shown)


thjaeckle:
Did you just read our minds? see #366 :D

kaizimmerm:
I wish I could, I wish...

kaizimmerm:
btw. I needed a few additional debug messages to hunt this down. I think they make general sense. Let me know if you want me to remove them.

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpPublisherActor.java
```diff
@@ -108,8 +107,21 @@ public AmqpPublisherActor create() {
...
     @Override
     protected void preEnhancement(final ReceiveBuilder receiveBuilder) {
-        // noop
+        receiveBuilder.match(StatusReport.class, this::handleConnectionStatusReport);
     }
+    
+    private void handleConnectionStatusReport(final StatusReport report) {
+      report.getClosedProducer().ifPresent(producer -> {  
+              log.debug("Try to remove JMS producer '{}' from cache.", producer);
+              Optional<Map.Entry<Destination, MessageProducer>> remove = producerMap.entrySet().stream().
```
kaizimmerm: I know this is ugly btw. However, when closed MessageProducer.getDestination is rejecting access as well. So I had to do this brute-force.

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/StatusReport.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Optional;
+import javax.annotation.Nullable;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import org.eclipse.ditto.services.connectivity.messaging.internal.ConnectionFailure;
+
+/**
+ * Message for status reporter.
+ */
+public class StatusReport {
```
thjaeckle: could you please make this class package private and `final`?

kaizimmerm: done

thjaeckle:
@kaizimmerm did you sign all commits? the ECA check is not happy - the new one is quite reliable in our feeling.

kaizimmerm:
@thjaeckle merge commits usually do not have a sign off. Let me check how to fix this.

kaizimmerm:
@thjaeckle I got it fixed but had to force push.

PR #375, opened 2019-04-05
Fix Nginx connectivity after Helm update

kaizimmerm:
Hi,

I fixed an issue where the reverse proxy was no longer able the access the gateway after a (pod) update. The reason was that no ClusterIP was used and the pod IP changed as a result of the update.

In addition I improved the boolean flag for MongoDB SSL a bit, introduced replicaCount as a setting for the pods, a version setting for Swagger-UI and added a few missing labels.

Kai

### DIFF
--- a/deployment/helm/eclipse-ditto/templates/ditto-cluster.yaml
+++ b/deployment/helm/eclipse-ditto/templates/ditto-cluster.yaml
@@ -62,7 +62,7 @@ spec:
             - name: TZ
               value: "{{ .Values.global.timezone }}"
             - name: IBM_JAVA_OPTIONS
-              value: "-XX:MaxRAMPercentage=80 -XX:InitialRAMPercentage=40 -XX:+ExitOnOutOfMemoryError"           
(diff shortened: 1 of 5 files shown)


kaizimmerm:
I apologise but I had to change a detail (MongoDB uri secret) in the PR after opening it. Did not work always the way it was.

ghandim:
@kaizimmerm Will give it a try :smile:

ghandim:
@kaizimmerm LGTM :+1: . But I would also suggest to add the gateway cluster endpoint in nginx.conf as upstream, than developers can also re-deploy the gateway without the necessity to restart nginx.

```
...
  upstream gateway {
    server  gateway:8080;
  }

  server {
...
 proxy_pass                    http://gateway/api; // switch for all gateway proxy_pass configured in nginx.conf
```
What do you think?

kaizimmerm:
@ghandim sounds great. However, I will be on vacation for the next two weeks. So either you wait until then or merge and do it yourself. I am open to follow either way.

kai

ghandim:
@thjaeckle LGTM :+1: I will prepare a separate pull request for the nginx configuration improvement.

PR #380, opened 2019-04-10
added missing OSGi imports for "org.atteo.classindex"

thjaeckle:
The added registries require that import during runtime.


Yannic92:
Depends on https://github.com/atteo/classindex/pull/50

PR #382, opened 2019-04-12
Remove parboiled2 dependency.

yufei-cai:
Akka-Http includes an actively maintained copy of parboiled2.

### DIFF
--- a/model/rql-parser/pom.xml
+++ b/model/rql-parser/pom.xml
@@ -35,7 +35,13 @@
 
         <dependency>
             <groupId>com.typesafe.akka</groupId>
-            <artifactId>akka-http_${scala.version}</artifactId>
+            <artifactId>akka-parsing_${scala.version}</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.typesafe.akka</groupId>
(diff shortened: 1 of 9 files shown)


yufei-cai:
Closing because Akka-Http introduces undesirable dependencies to ditto-model-query.

model/rql-parser/pom.xml
```diff
@@ -35,7 +35,13 @@
 
         <dependency>
             <groupId>com.typesafe.akka</groupId>
-            <artifactId>akka-http_${scala.version}</artifactId>
+            <artifactId>akka-parsing_${scala.version}</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.typesafe.akka</groupId>
+                    <artifactId>akka-actor_${scala.version}</artifactId>
+                </exclusion>
```
thjaeckle: Isn't akka-actor a "provided" dependency for akka-parsing?
Looks that way here: https://mvnrepository.com/artifact/com.typesafe.akka/akka-parsing_2.12/10.1.8

yufei-cai: True; deleting the exclusion does not introduce akka-actor as dependency. Will amend.

PR #409, opened 2019-05-16
Forbid null character in strings to prevent JsonParseException

stmaute:
This PR fixes JsonParseException when a string contains null characters.
Using a string with '\\u0000' inside now causes an JsonInvalidException.


jufickel-b:
Did you guys check the performance of the new implementation with JMH? ditto-json is crucial for the overall performance of Ditto.

stmaute:
Not yet but I think it is a good idea to do this.
I will do this right away.

PR #410, opened 2019-05-17
Cleanup after TestConnection command for AMQP 1.0

dguggemos:
Connections created when testing an AMQP 1.0 connection were not closed and stayed open, potentially competing for messages with the actual connection. This PR also prevents execution of parallel test connection commands which might lead to strange behavior.


stmaute:
LGTM :+1:

PR #411, opened 2019-05-22
Make http request actor extensible

yufei-cai:
- Gave user complete freedom to choose HttpRequestActor implementations.
- Moved dynamic class loading from MappingStrategies to ditto-services-utils-akka as Scala object for Java API.
- Added the option of full HTTP logging to Gateway.
- Removed a tautology from CreateThingStrategy.

### DIFF
--- a/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/actors/AbstractHttpRequestActor.java
+++ b/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/actors/AbstractHttpRequestActor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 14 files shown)


services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/actors/AbstractHttpRequestActor.java
```diff
@@ -65,18 +64,13 @@
...
 import akka.util.ByteString;
 import scala.concurrent.duration.Duration;
 import scala.util.Either;
 
-/**
- * Every HTTP Request causes one new Actor instance of this one to be created.
- * It holds the original sender of an issued {@link Command} and tells this one the completed HttpResponse.
- */
-public final class HttpRequestActor extends AbstractActor {
+public abstract class AbstractHttpRequestActor extends AbstractActor {
```
thjaeckle: Could you please add a class javadoc here?

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/AbstractRoute.java
```diff
@@ -61,6 +64,9 @@
  */
 public abstract class AbstractRoute {
 
+    // TODO YC: refactor config key.
```
thjaeckle: what about that TODO? Is that for after the configuration refactoring?

yufei-cai: Yes.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/actors/AbstractHttpRequestActor.java
yufei-cai: Good point.

PR #416, opened 2019-05-29
Support for persistence without listCollections filter support

kaizimmerm:
CosmosDB does not support listCollections command with additional regex filters.

So I added a Java filter as addition. Will not make a difference for MongoDB but ensure that only the right collections are handled in CosmosDB case.

I added as well a few debug messages to hunt this down as well as a test which helped me to rule a driver config issue out.

Let me know what you think.

Signed-off-by: Kai Zimmermann <kai.zimmermann@microsoft.com>

### DIFF
--- a/services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/MongoClientWrapperTest.java
+++ b/services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/MongoClientWrapperTest.java
@@ -13,13 +13,11 @@
 package org.eclipse.ditto.services.utils.persistence.mongo;
 
 import static org.assertj.core.api.Assertions.assertThat;
-
 import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
-
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/MongoClientWrapperTest.java
```diff
@@ -54,6 +52,27 @@ private static String createUri(final boolean sslEnabled) {
         return connectionString.getConnectionString();
     }
 
+    @Test
+    public void createByUriWithExtraSettings() {
```
thjaeckle: This test is not related to the changes made to the `MongoReadJournal`, right?
We can of course keep it .. just wanted to make sure why you added the test :)

kaizimmerm: Yes, its a long story but in the beginning I was not sure if the config by URI worked correctly. So, yes, the test is not related but I think it makes sense to keep.

PR #420, opened 2019-06-05
Simple throttling for amqp 1.0 consumers

dguggemos:
This PR implements a simple throttling mechanism for amqp 1.0 consumers. The throttling is configurable by defining the number of messages allowed per time interval.


yufei-cai:
LGTM

PR #422, opened 2019-06-07
Bugfix/block namespace

yufei-cai:
Concierge cache loaders did not respect the distributed data of blocked namespaces. Combined with the over-eager invalidation of caches, it led to spurious messages into namespaces already blocked, preventing the recipient actors from shutting down properly.

This patch:
- introduces a delay before cache invalidation
- removes a no-op collection name suffix builder incompatible with AbstractNamespaceOpsActor
- improves logging of shutdown behavior and cache invalidation


ffendt:
LGTM

PR #423, opened 2019-06-11
Enable in-place update of secrets

yufei-cai:
Changes

1. Make docker secrets renamable for live secret rotation as a workaround of [moby #29882](https://github.com/moby/moby/issues/29882).

2. Fixed that Ditto did not read docker secrets in file-based host environments.

3. Added a separate credentials for overall status routes.


ffendt:
LGTM

Issue #318, opened 2019-01-18
Collect connectivity log entries and provide via devops command


thjaeckle:
Goal: 
In order to let a user analyze failures / unexpected behavior in his connections (e.g. to an MQTT or AMQP broker) without giving him full access to the log-files of Ditto, a *connection scoped* mechanism for retrieving connectivity logs should be added.

With a single devops command those logs can be retrieved (as JSON).

The logs should contain:
- infos
- warnings
- errors
- connection events (e.g.: changes in connection state, reconncetions, etc.)

thjaeckle:
Fixed by #430

PR #426, opened 2019-06-14
Enable connection scoped logs

ffendt:
With this PR, Ditto will be able to store logs for each connection and provides access via devops commands (solves issue #318). The logs will contain information on messages processed by connections and also some runtime information for the connection itself. This feature shall help users to get insights in what goes right or wrong in a connection without the need to access the service logs.

Connection logging is disabled by default and needs to be enabled via the new `connectivity.commands:enableConnectionLogs` devops command.
Afterwards it is possible to retrieve the logs for this connection via the command `connectivity.commands:retrieveConnectionLogs`.
The available log can be reset (but currently not disabled) by using `connectivity.commands:resetConnectionLogs`.
Further information and the exact structure of the commands can be found in the [documentation on managing connections](documentation/src/main/resources/pages/ditto/connectivity-manage-connections.md)

You can update the configuration of the logging feature with the following environment variables when starting the connectivity service:
* `CONNECTIVITY_LOGGER_SUCCESS_CAPACITY` (default: 10): The number of success logs stored for each log category and type. (resulting per connection in `max success logs = categories * types * <number>`)
* `CONNECTIVITY_LOGGER_FAILURE_CAPACITY` (default: 10): The number of failure logs stored for each log category and type. (resulting per connection in `max failure logs = categories * types * <number>`)
* `CONNECTIVITY_LOGGER_LOG_DURATION` (default: 1h): How long logging will stay enabled for a connection after enabling it.
* `CONNECTIVITY_LOGGER_ACTIVE_CHECK_INTERVAL` (default: 5m): Interval in which the service will check if the logging timeframe was exceeded for a connection and needs to be disabled again.


yufei-cai:
Superceded by #427.

PR #428, opened 2019-06-17
Feature/purge entities

Yannic92:
This pr provides a new command "PurgeEntities" which allows to really delete entities (events and snapshots) from the database instead of just marking an entity as deleted.


jufickel-b:
LGTM

PR #437, opened 2019-07-05
Fix amqp session recovery

dguggemos:
This PR fixes amqp session recovery and allows to configure it per connection.


ffendt:
Looks good to me. In a future PR we could also think about providing the same config value through the typesafe config and only use the specific connection config as an override over that value.

PR #438, opened 2019-07-08
Bugfix/resolve secret config

Yannic92:
the config needs to be resolved because in [SecrectsAsConfigSupplier](https://github.com/bsinno/ditto/blob/916b03118b119d5c691c892300492f87f97a382e/services/utils/config/src/main/java/org/eclipse/ditto/services/utils/config/raw/SecretsAsConfigSupplier.java#L82) the config is read.


stmaute:
LGTM :+1:

PR #445, opened 2019-07-16
Fix concierge memory leak

thjaeckle:
Substream creation for each namespace causes a memory leak when using a lot of different namespaces.
This premature optimization seems not to be necessary for overall performance.


ffendt:
Nice catch. LGTM

PR #453, opened 2019-07-26
Statistics-actor improvements

yufei-cai:
Changes
- Make shard regions queried by the statistics-actor configurable.
- `RetrieveStatistics` are now always answered by the cached value, updated at a configured interval.
- Filters by shards and by namespaces are added to `RetrieveStatisticsDetails`.
- Cache statistics-details for a configured duration to reduce cluster traffic.
- Fixed discarding of previous behavior before `getContext().unbecome()`.
- Fixed timeout of queries during update of statistics.

### DIFF
--- a/services/gateway/starter/src/main/resources/gateway.conf
+++ b/services/gateway/starter/src/main/resources/gateway.conf
@@ -144,6 +144,9 @@ ditto {
       // how often to update hot entities count for the public
       update-interval = 15s
 
+      // minimum pause between computations of statistics details
+      details-expire-after = 3s
+
       // CAUTION: no shard name should be a part of another shard name, because statistics-actor identifies
       // messages from shard regions by checking whether the sender's actor path contains the shard name.
       // This restriction does not apply to cluster role names; it is okay for example to have both the role "things"
--- a/services/gateway/starter/src/main/resources/gateway.conf
+++ b/services/gateway/starter/src/main/resources/gateway.conf
@@ -137,7 +137,14 @@ ditto {
     }
 
     statistics {
-      // CAUTION: no shard name should be the prefix of another shard name, because statistics-actor identifies
+
+      // how long to wait for messages from other cluster members
+      ask-timeout = 5s
+
+      // how often to update hot entities count for the public
+      update-interval = 15s
+
+      // CAUTION: no shard name should be a part of another shard name, because statistics-actor identifies
       // messages from shard regions by checking whether the sender's actor path contains the shard name.
       // This restriction does not apply to cluster role names; it is okay for example to have both the role "things"
       // and the role "things-search".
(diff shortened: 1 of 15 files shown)


signals/commands/devops/src/main/java/org/eclipse/ditto/signals/commands/devops/RetrieveStatisticsDetails.java
```diff
@@ -42,18 +49,37 @@
...
     public static final String TYPE = TYPE_PREFIX + NAME;
 
-    private RetrieveStatisticsDetails(final DittoHeaders dittoHeaders) {
+    private final List<String> shardRegions;
+    private final List<String> namespaces;
+
+    private RetrieveStatisticsDetails(final List<String> shardRegions, final List<String> namespaces,
+            final DittoHeaders dittoHeaders) {
         super(TYPE, null, null, dittoHeaders);
+        this.shardRegions = shardRegions;
```
thjaeckle: The class isn't immutable when we just assign the lists. Please use pattern `Collections.unmodifiableList(new ArrayList(shardRegions))`

signals/commands/devops/src/main/java/org/eclipse/ditto/signals/commands/devops/RetrieveStatisticsDetails.java
```diff
@@ -42,18 +49,37 @@
...
+        this.namespaces = namespaces;
```
thjaeckle: same as for the other list

services/gateway/starter/src/main/resources/gateway.conf
```diff
@@ -135,6 +135,40 @@ ditto {
...
+    statistics {
+
+      // how long to wait for messages from other cluster members
+      ask-timeout = 5s
+
+      // how often to update hot entities count for the public
+      update-interval = 15s
+
+      // minimum pause between computations of statistics details
+      details-expire-after = 3s
```
thjaeckle: Either reduce to less often or add an environment variable so that this may be configured.

services/gateway/starter/src/main/resources/gateway.conf
```diff
@@ -135,6 +135,40 @@ ditto {
...
       }
     }
+
+    statistics {
+
+      // how long to wait for messages from other cluster members
+      ask-timeout = 5s
+
+      // how often to update hot entities count for the public
+      update-interval = 15s
```
thjaeckle: Either reduce to less often or add an environment variable so that this may be configured.

signals/commands/devops/src/main/java/org/eclipse/ditto/signals/commands/devops/RetrieveStatisticsDetails.java
yufei-cai: good point

PR #469, opened 2019-08-12
Improve robustness of connections, especially of AMQP 1.0

yufei-cai:
- Fixed a bug where we relied on Akka FSM state timeout for connection handling. Actual duration of FSM state timeout is not deterministic.

- All client actors now react to `ConnectionFailure` in connected state by restarting the connection after exponential backoff.

- AMQP 1.0 consumers and publishers configured as sources and targets are recreated eagerly as the broker close them.

- There is a configurable limit on the number of open AMQP publishers created for `reply-to` addresses.

- Upgrade to Caffeine 2.7.0 due to [issue 301](https://github.com/ben-manes/caffeine/issues/301).

- When connected, client actors answer `OpenConnection` by `Status.Success` instead of ignoring it.

### DIFF
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/BaseClientActor.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/BaseClientActor.java
@@ -752,7 +752,7 @@ public abstract class BaseClientActor extends AbstractFSM<BaseClientState, BaseC
                         String.format("Connection failed due to: {0}. Will reconnect after %s.", nextBackoff);
                 connectionLogger.failure(errorMessage, event.getFailureDescription());
                 log.info("Connection failed: {}. Reconnect after {}.", event, nextBackoff);
-                return goTo(CONNECTING).forMax(reconnectTimeoutStrategy.getNextBackoff()).using(data.resetSession()
+                return goTo(CONNECTING).forMax(nextBackoff).using(data.resetSession()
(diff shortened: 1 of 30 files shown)


thjaeckle:
Requires an Eclipse CQ for the updated Caffeine version.

thjaeckle:
Created CQ for caffeine 2.8.0: 
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20651

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/BaseClientActor.java
```diff
@@ -1262,34 +1262,33 @@ private void increaseTimeoutAfterRecovery() {
...
-                    } else if (isLonger(durationBetweenTimeouts.dividedBy(3L), expectedDurationBetweenTimeouts)) {
-                        // safe for 3x the expected duration
-                        currentTries = Math.max(0, currentTries - 1);
-                    }
-                    currentTimeout = maxDuration(minTimeout, currentTimeout.minus(recovery));
-                    nextBackoff = maxDuration(minBackoff, nextBackoff.minus(recovery));
+                final Duration durationSinceLastTimeout = Duration.between(lastTimeoutIncrease, now);
+                final Duration resetThreshold = maxTimeout.plus(maxBackoff).multipliedBy(2L);
+                if (isLonger(durationSinceLastTimeout, resetThreshold)) {
+                    reset();
```
dguggemos: This looks much more maintainable now 👍

PR #471, opened 2019-08-15
Reduced TTL for ThingUpdater to 15m and switched to passivation.

stmaute:
ThingUpdater being a non-persistent actor with a cheap constructor, we can trade more restarts for less memory consumption. Instead of sending poison pill to self,it now passivates to avoid message loss.

Signed-off-by: Cai Yufei (INST/ECS1) <yufei.cai@bosch-si.com>


stmaute:
LGTM :+1:

PR #475, opened 2019-08-23
WIP: Feature/typed entity ids

Yannic92:
This PR introduces validated Java representations for entity IDs


services/concierge/actors/src/main/java/org/eclipse/ditto/services/concierge/actors/cleanup/EventSnapshotCleanupCoordinator.java
```diff
@@ -360,8 +362,10 @@ private void scheduleWakeUp(final Duration when) {
                         .match(ConnectionTag.class, connTag ->
                                 askShardRegionForCleanup(shardRegions.connections(), ConnectivityCommand.RESOURCE_TYPE, connTag))
                         .matchAny(e -> {
+                            final String msg = "Unexpected entity ID type: " + e.getClass().getSimpleName();
                             final CleanupPersistenceResponse error =
-                                    CleanupPersistenceResponse.failure("Unexpected entity ID type: ", DittoHeaders.empty());
+                                    //TODO: Dont build entity id of message.
```
thjaeckle: What about this TODO? Still open?

services/concierge/actors/src/main/java/org/eclipse/ditto/services/concierge/actors/cleanup/EventSnapshotCleanupCoordinator.java
```diff
@@ -388,7 +392,8 @@ private void scheduleWakeUp(final Duration when) {
                     } else {
                         final String msg = String.format("Unexpected response from shard <%s>: result=<%s> error=<%s>",
                                 resourceType, Objects.toString(result), Objects.toString(error));
-                        return CleanupPersistenceResponse.failure(msg, cleanupPersistence.getDittoHeaders());
+                        return //TODO: Dont build entity id of message.
```
thjaeckle: Open TODO.

services/connectivity/mapping/src/main/java/org/eclipse/ditto/services/connectivity/mapping/javascript/DefaultIncomingMapping.java
```diff
@@ -38,12 +37,7 @@ static DefaultIncomingMapping get() {
 
     @Override
     public Optional<Adaptable> apply(final ExternalMessage message) {
-        return Optional.ofNullable(
-                message.getTextPayload()
-                        .orElseGet(() -> message.getBytePayload()
-                                .map(b -> StandardCharsets.UTF_8.decode(b).toString())
-                                .orElse(null))
-        ).map(plainString -> DittoJsonException.wrapJsonRuntimeException(() -> {
+        return message.getTextPayload().map(plainString -> DittoJsonException.wrapJsonRuntimeException(() -> {
```
thjaeckle: Why did you do that change? And why in this PR?
Seems like we no longer do a fallback to bytePayload and decoding now ..

signals/base/src/main/java/org/eclipse/ditto/signals/base/WithIdButActuallyNot.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
thjaeckle: Year should be 2019.

model/base/src/main/java/org/eclipse/ditto/model/base/entity/id/RegexPatterns.java
```diff
@@ -0,0 +1,60 @@
...
+  *
+  * SPDX-License-Identifier: EPL-2.0
+  */
+ package org.eclipse.ditto.model.base.entity.id;
+
+ import java.util.regex.Pattern;
+
+ public final class RegexPatterns {
+
+     public static final String NAMESPACE_GROUP_NAME = "ns";
```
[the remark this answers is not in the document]
Yannic92: I'd like to keep them public and add javadoc for them as my intention was to define regex patterns that can be used all over the service or for users of our model. Do you agree with this?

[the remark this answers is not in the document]
thjaeckle: Jup, cool with me.

model/placeholders/src/main/java/org/eclipse/ditto/model/placeholders/EnforcementFactoryFactory.java
```diff
@@ -29,8 +29,8 @@
      * @param <I> the type from which the input values are resolved
      * @return the new {@link EnforcementFactoryFactory}
      */
-    public static <I> EnforcementFilterFactory<I, String> newEnforcementFilterFactory(final Enforcement enforcement,
-            final Placeholder<I> inputFilter, final Placeholder<String> filterPlaceholderResolver) {
+    private static <O> EnforcementFilterFactory<O, CharSequence > newEnforcementFilterFactory(final Enforcement enforcement,
```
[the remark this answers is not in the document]
Yannic92: Ah, I was confused about I(nput) and (O)utput there because this private method is called from the public method which used O. I'd revert this change back to use I again and change O in the public method to I aswell as currently O defines the type that is required for input.

services/connectivity/mapping/src/main/java/org/eclipse/ditto/services/connectivity/mapping/javascript/DefaultIncomingMapping.java
Yannic92: oh my .. Thank you for this catch. I trusted IntelliJ blindly .. I will revert this and try to restructure this statement so intelliJ won't recommend to remove this optional wrapping

services/concierge/actors/src/main/java/org/eclipse/ditto/services/concierge/actors/cleanup/EventSnapshotCleanupCoordinator.java
Yannic92: Yes this todo is still open. I "worked around" this right now by wrapping the mssage in a "DefaultEntityId" but this is obviously not the way we should use this ID. I'd suggest opening a new issue to provide a proper way for error responses to CleanupPersistence commands. Is this okay for you?

thjaeckle: Yes, ok for me

Yannic92: https://github.com/eclipse/ditto/issues/476

PR #478, opened 2019-08-28
Bugfix/testconnection exception handling

Yannic92:
Right now exceptions thrown during processing of TestConnection command in BaseClientActor are "swallowed" by ConnectionActor which leads to a timeout while waiting for the TestConnectionResponse. This PR wraps those exceptions into a failure status.


yufei-cai:
LGTM

PR #485, opened 2019-09-04
Improve secret retrieval for Azure Resource Manager based deployments

kaizimmerm:
Two small improvements:

- Extract variables directly from resource manager.
- Upgrade default k8s version.


ghandim:
Will give it a try today.

ghandim:
Looks good to me from a technical point of view. But the curl for verifying the deplyoment at the end could be misleading if you use the CosmosDB scenario. Then you get "DOWN" as result which is correct in this case because things-search cannot come up with CosmosDB. Think you should add a hint for this scenario.

kaizimmerm:
@ghandim good catch! I added a comment.

PR #493, opened 2019-09-23
Fix timing issues during client actor initialization

dguggemos:
Connection were able to consume messages from external systems before the actor hierarchy below client actors (mapper, publisher, cosumer) was ready to process messages. This should be fixed now.


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpClientActor.java
```diff
@@ -249,27 +251,50 @@ protected void doDisconnectClient(final Connection connection, @Nullable final A
...
-            // note: start order is important (publisher -> mapping -> consumer actor)
-            startPublisherActor();
-            startCommandConsumers(c.consumerList, jmsActor);
-            notifyConsumersReady();
         } else {
             log.info("ClientConnected was not JmsConnected as expected, ignoring as this probably was a reconnection");
         }
     }
 
+    protected CompletionStage<Status.Status> startPublisherActor() {
```
thjaeckle: is missing `@Override`

PR #506, opened 2019-10-04
Enhance docs of docker deployment (fixes #443)

### LINKED ISSUE #443: 'The supplied authentication is invalid' with custom devops user
I've setup a Ditto-Instance as in the [deployment/docker](https://github.com/eclipse/ditto/tree/master/deployment/docker) part of the project and changed the devops password to a custom one which I hashed using `openssl passwd --quiet` and updated in `nginx-devops.htpasswd`.

Unfortunately I get the error message `The supplied authentication is invalid` this way when trying to access any `/devops` resource. When using a random username and/or password it returns a "401 Unauthorized".

My nginx.conf is as follows

```
   # devops
    location /devops {
      include nginx-cors.conf;

      auth_basic                    "Authentication required";
      auth_basic_user_file          nginx-devops.htpasswd;

      proxy_pass                    http://gateway:8080/devops;
      proxy_http_version            1.1;
      proxy_set_header              Host                $http_host;
      proxy_set_header              X-Real-IP           $remote_addr;
      proxy_set_header              X-Forwarded-For     $proxy_add_x_forwarded_for;
      proxy_set_header              X-Forwarded-User    $remote_user;
    }
```

EDIT:

Ok, well, I got closer to an answer. I took the nginx-devops config from the subfolder sandbox because I was searching for a way to work with the devops API and couldn't find something in the docs. Apparently I have to at least configure the same passwort in nginx as well as in the gateway-container through the environment variable `DEVOPS_PASSWORD`. 

I'm wondering if there is a security flaw to only use the gateways-service authorization without nginx?

Is there a reason why this is documented nowhere?

w4tsn:
Add entrypoint section to compose file

To make it more obvious for the operator how a service may be
configured the entrypoint section used for the services is now
explicitly part of the compose file.

This way someone new to ditto and java has a better chance of
understanding how the services may be configured.

Add short summary on config to docker readme

To make it more clear this commit adds a section on configuration
to the docker deployment readme. This section handles two topics.

First the mechanics of a java application configuration are laid out.
The configuration is depicted via the entrypoint section of the
compose-file.
This approach is chosen, because it enables a fairly easy configuration
of the java service via the java VM parameters.
The chosen example is the devops command from the gateway service
due to it's practicability and importance when setting up an instance
for oneself.

Second follows a description on how to retrieve possible configuration
options either via HTTP REST API or via the given configuration files in
the repository. This should enable more people to faster learn how
to work with and configure each service.

Add a config section to operational docs

To enhance the understanding for configuration of the ditto
micro-services this commit adds a configuration section to the public docs
for operators. It contains a docker-specific example and hints on where
to find available configuration options.

This is necessary to give operators new to ditto and java applications
a better start on hosting and using their own instances of ditto.

Remove misleading configuration from sandbox

This will remove confusing and unecessary configuration from the sandbox
in the docker deployment.

Setting a devops user and password basic auth before the devops route is
not just uneccessary, because this is configured in the gateway service
already, it's also misleading to new operators of ditto.

This is an enhancement with regard to #443

### DIFF
--- a/deployment/docker/README.md
+++ b/deployment/docker/README.md
@@ -21,6 +21,30 @@ password in front like this:
 ditto:A6BgmB8IEtPTs
 ```
 
+## Configuration of the services
+
+You may configure each service via passing variables to the java VM in the entrypoint section for each
+service.
+
+```yml
+    ...
+    entrypoint:
+      - java
+      # Alternative approach for configuration of the service
+      - -Dditto.gateway.authentication.devops.password=foobar
+      - -jar
+      - starter.jar
+```
+
+To get a list of available configuration options you may retrieve them from a running instance via:
+
+```bash
+# Substitute gateway with the service you are interested in
+curl http://devops:foobar@localhost:8080/devops/config/gateway/?path=ditto
+```
+
+Or by going through the configuration files in this repository e.g. [/services/gateway/starter/src/main/resources/gateway.conf](https://github.com/eclipse/ditto/blob/master/services/gateway/starter/src/main/resources/gateway.conf).
+
 ## Start Eclipse Ditto
 
 ```bash
--- a/documentation/src/main/resources/pages/ditto/installation-operating.md
+++ b/documentation/src/main/resources/pages/ditto/installation-operating.md
@@ -11,6 +11,29 @@ Once you have successfully started Ditto, proceed with setting it up for continu
 
 This page shows the basics for operating Ditto.
 
+## Configuration
+
+Each service may be configured via it's java VM variables on service startup.
+
+The following example configures the devops password of the gateway-service started via docker-compose. In order
+to supply additional configuration one has to add the variable in the corresponding `entrypoint` section of the
+`docker-compose.yml` file.
+
+```yml
+    ...
+    entrypoint:
+      - java
+      # Alternative approach for configuration of the service
+      - -Dditto.gateway.authentication.devops.password=foobar
+      - -jar
+      - starter.jar
+```
+
+The executable for the microservice is called `starter.jar`. The configuration variables have to be set before
+the `-jar` option.
+
+For a list of available configuration options go to the respective configuration files under the services 
+directory. E.g. for the gateway service visit [/services/gateway/starter/src/main/resources/gateway.conf](https://github.com/eclipse/ditto/blob/master/services/gateway/starter/src/main/resources/gateway.conf).
 
 ## Logging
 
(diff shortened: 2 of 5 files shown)


deployment/docker/README.md
```diff
@@ -21,6 +21,30 @@ password in front like this:
...
+```
+
+To get a list of available configuration options you may retrieve them from a running instance via:
+
+```bash
+# Substitute gateway with the service you are interested in
+curl http://devops:foobar@localhost:8080/devops/config/gateway/1?path=ditto
+```
+
+Or by going through the configuration files in this repository e.g. [/services/gateway/starter/src/main/resources/gateway.conf](../../services/gateway/starter/src/main/resources/gateway.conf).
```
thjaeckle: Relative links to sourcecode won't work in the statically built website of Ditto.
Please replace with the "real" HTTP link to github.

deployment/docker/README.md
```diff
@@ -21,6 +21,30 @@ password in front like this:
...
+      - -Dditto.gateway.authentication.devops.password=foobar
+      - -jar
+      - starter.jar
+```
+
+To get a list of available configuration options you may retrieve them from a running instance via:
+
+```bash
+# Substitute gateway with the service you are interested in
+curl http://devops:foobar@localhost:8080/devops/config/gateway/1?path=ditto
```
thjaeckle: The `1`  in the URL path is only valid for docker-swarm based deployments.
In general this is the `instanceIndex` to put in there - for kubernetes based deployments that would be the Pod-ID instead.

documentation/src/main/resources/pages/ditto/installation-operating.md
```diff
@@ -11,6 +11,29 @@ Once you have successfully started Ditto, proceed with setting it up for continu
...
+      - -Dditto.gateway.authentication.devops.password=foobar
+      - -jar
+      - starter.jar
+```
+
+The executable for the microservice is called `starter.jar`. The configuration variables have to be set before
+the `-jar` option.
+
+For a list of available configuration options go to the respective configuration files under the services 
+directory. E.g. for the gateway service visit [/services/gateway/starter/src/main/resources/gateway.conf](../../../../../../services/gateway/starter/src/main/resources/gateway.conf).
```
thjaeckle: Relative links to sourcecode won't work in the statically built website of Ditto.
Please replace with the "real" HTTP link to github.

thjaeckle:
The PR has now a conflict with your previous changes :)
Could you please merge from the master branch and resolve those conflict.

PR #516, opened 2019-10-10
Implement reconnect back-off mechanism for targets in AMQP 1.0 connections

ffendt:
This is needed since there is a reconnect loop if the used JMS client gets an error after connecting successfully. This can be e.g. an authorization error which is thrown after the connection initially could be established. This will cause a retry for the target. Since the configuration seems to be wrong for that point in time, the connection will be looping through opening, getting disconnected and reopening the target.

As the target should get opened successfully when the endpoints accepts it, it makes sense to continue retrying the connection. That's why a back-off mechanism makes sense. On connection error, the PublisherActor will now back-off for some time until trying to reconnect. The back-off duration can be configured using the config keys ditto.connectivity.amqp10.backoff.timeout.[min-timeout, max-timeout]. The current implementation will double the previous timeout until max-timeout is reached.


thjaeckle:
Looks like there is indeed a test failure now related to the changes of this PR.

dguggemos:
Yes, I'm already in discussion with @ffendt.

PR #520, opened 2019-10-17
stabilize AmqpClientActorTest

thjaeckle:
by using another style for Mockito mocks
* it sometimes failed due to multi-threading


yufei-cai:
LGTM

PR #522, opened 2019-10-18
Fix state timeout in BaseClientActors

thjaeckle:
BaseClientActors sometimes did not receive the state timeout message reliably.
That was causing trouble for scaled connections (clientCount > 1) as the scaled connections did not connect to their backends like they should have.
This fixes that bug by simplifying the connection logic (by default always connect).


thjaeckle:
Changes LGTM :+1:

PR #527, opened 2019-10-22
Remove remnant of an unreleased module.

yufei-cai:
```
<groupId>org.eclipse.ditto</groupId>
<artifactId>ditto-services-topologies</artifactId>
<version>It was never meant to be.</version>
```


yufei-cai:
Does it look good still?

ffendt:
Yes, still looks good 👍

Issue #460, opened 2019-07-31
Add to documentation that AMQP connection target support topics as well


thjaeckle:
Ditto supports `targets` of connections to be either queues (being the default) or topics:
https://github.com/eclipse/ditto/blob/master/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpTarget.java#L44

We should add documentation for that and explain differences.

Issue #467, opened 2019-08-08
Document eventual consistency of the things-search


thjaeckle:
Is still missing from the documentation.
It is not documented that the search only guarantees eventual consistency.

PR #528, opened 2019-10-22
Fixing some minor documentation issues

thjaeckle:
Fixes: #460 #467


documentation/src/main/resources/pages/ditto/installation-operating.md
```diff
@@ -49,13 +61,19 @@ ditto.gateway.authentication {
 }
 ```
 
+In order to do this by specifying a Java system property, use the following:
+
+```
+-Dditto.gateway.authentication.oauth.openid-connect-issuers.myprovider=https://localhost:9000/
```
ffendt: With #535 we need to omit https:// in the issuer configuration

thjaeckle: indeed

PR #548, opened 2019-11-08
Add hivemq based mqtt 5 client implementation

### LINKED ISSUE #547: MQTT 5 Support
With #487 the HiveMQTT client was introduced which supports MQTT 5 protocol. A lot of the more advanced features like header mapping and source enforcement are only supported with headers which MQTT 5 provides.

Implementation
---

Not yet implemented MQTT 3 Features

- keepAlive
- willPublish

MQTT 5 client configuration options:

- [ ] sessionExpiryInterval
- [ ] restrictions
- [x] userProperties

Sub-tasks:

- [x] Update Documentation
  - [x] Basic Concepts / Connections
  - [x] Connectivity API / MQTT 5 Protocol Binding
  - [x] Document client-specific config (e.g. `clientId`)
- [x] Implement basic MQTT 5 client based off HiveMQTT 3
  - [x] Tests
- [x] Implement MQTT 5 userProperties
  - [x] Tests
- [x] Implement Header Mapping
  - [x] Tests
- [x] Implement Source Enforcement
  - [x] Tests
- [x] Add to DefaultClientActorPropsFactory.java
  - [x] Tests
- [x] Add mqtt 5 to MqttValidator
  - [x] Tests
- [ ] Implement Mqtt5Config (and Mqtt3Config?)
  - [ ] Tests

Discussion
---

**user properties and header mapping** are the way a client may provide additional header values in MQTT 5. Additional to the user properties, there are fixed and variable properties available. I had a look at the EMQX Broker for a reference of available properties. These are among others `Correlation Data`, `Response Topic` and `Content Type`. Those can be implemented as general available headers for header mapping.

User defined properties for targets can be defined via the specific Configs map in the connections configuration.

**Source Enforcement** for the time being I have no clear idea how to implement the source enforcement. MQTT 5 provides a property `Authentication Method` together with `Authentication Data`. This does not apply for publish and other message types however. 

Question: The clientid is already extracted and some brokers enforce the client id via certificate based authentication. Is this enough to have a first source enforcement?

Additionally to this, the broker would require the ability to set user defined properties per message right? Since the user defined properties are the only headers that are valid for all message types.

**Validator and MQTT5Config?** Is it necessary to have these classes separate from mqtt3? The validator currently has a hard check for `MQTT 3.1.1` in the protocol param. Should there be just an extension for `MQTT 5` or a separate validator? What do you think?

w4tsn:
This adds a first basic mqtt 5 client implementation based on the
already existing mqtt 3 implementation.

This introduces two new ConnectionType options 'mqtt-3' and 'mqtt-5' to
become more consistent with the already present format in amqp which
is also used in the docs. The old option 'mqtt' is kept for backwards 
compatibility reasons. I recommend to deprecate it.

The mqtt-5 connection type is implemented seperately to the mqtt-3
implementation to ensure both can be used in parallel, as it is with
amqp.

Tests are provided.

Docs are missing (yet).

Signed-off-by: Alexander Wellbrock <a.wellbrock@mailbox.org>

Closes #547


documentation/src/main/resources/pages/ditto/connectivity-protocol-bindings-mqtt5.md
```diff
@@ -0,0 +1,185 @@
...
+The common configuration for connections in [Connections > Sources](basic-connections.html#sources) and 
+[Connections > Targets](basic-connections.html#targets) applies here as well. 
+Following are some specifics for MQTT connections:
+
+### Source format
+
+For an MQTT connection:
+
+* Source `"addresses"` are MQTT topics to subscribe to. Wildcards `+` and `#` are allowed.
+* `"authorizationContext"` may _not_ contain placeholders `{%raw%}{{ header:<header-name> }}{%endraw%}` as MQTT 5
```
thjaeckle: That's not correct - for MQTT 5 we have the user properties which can be mapped to headers :)

documentation/src/main/resources/pages/ditto/connectivity-protocol-bindings-mqtt5.md
```diff
@@ -0,0 +1,185 @@
...
+  "authorizationContext": ["ditto:outbound-auth-subject", "..."],
+  "qos": 0
+}
+```
+
+#### Target header mapping
+
+MQTT 5 supports so-called user defined properties, which are defined for every message type. The [header mapping](connectivity-header-mapping.html) applies to the supported MQTT 5 specific headers as well as to the user defined properties.
+
+In order to provide user defined properties for the targets of the MQTT 5 connector, use the [specific configs](https://www.eclipse.org/ditto/basic-connections.html#specific-config) field of the connector configuration. The entries are interpreted as key-value-pairs and directly set as user properties.
```
thjaeckle: better use link target `basic-connections.html#specific-config`

documentation/src/main/resources/pages/ditto/connectivity-protocol-bindings-mqtt5.md
w4tsn: I removed the wrong statement and replaced it with a short description borrowed from amqp-1 docs.

w4tsn:
The latest commit adds the following:

Add MQTT5 header mapping and user props

This adds two things. The user properties of mqtt 5 publish messages
are now mapped to headers in ditto messages. Additionally this will
map mqtt 5 specific protocol headers for correlation_id, reply-to and
content-type.

The mapping also works for publishing messages, where the protocol
specific configuration map is mapped to mqtt 5 user properties.

Tests for the mqtt 5 consumer are missing.

This will also add additional headers for the test suite of the 
AbstractPublisherActorTest for correlation-id, reply-to and 
content-type. Those are then utilized in the mqtt 5 publisher 
test.

w4tsn:
I'm unsure about the tests for the mqtt 5 consumer. I've played around with the AbstractConsumerActorTest, but until now had no success in setting up and understanding the test cases.

thjaeckle:
@w4tsn hm, I don't know `AmqpConsumerActorTest` "personally" ;)
Did you try - based on the `RabbitMQConsumerActorTest` (which only overwrites the 2 required methods of the abstract base class) - to get a test running?

`AmqpConsumerActorTest` just defines some more tests in addition to the abstract base class.

thjaeckle:
hi @w4tsn 
You now have a conflict with the base branch - I hope/guess that is not too much.

What are your plans regarding the PR?
I'm asking because we release Ditto 1.0.0 on 12th of December and if we want to get the MQTT5 feature in, we should hurry ;)
Currently I would guess that this won't make it into 1.0.0, but then in 1.1.0.

What do you think?

w4tsn:
Hey there @thjaeckle ! Yes, I'd be happy to provide it in time, I have only one problem left though. I've successfully implemented the required features including enforcement and header mapping, but I'm currently stuck at the following problem:

The current approach is to have a topic placeholder + respective placeholder for the enforcement filter. I replaced that with the header enforcement filter like it's done in AMQP and RabittMQ classes. The problem however now is, that the client actor fails with timeouts in three tests and I can't figure out what's wrong just now. I suppose I somehow have to merge both topic and header enforcement filter, but had no time till now to figure out how to do this.

thjaeckle:
@w4tsn I had a look at it and added both header based and mqtt address based enforcement to the Mqtt5 implementation.
I took also the liberty to merge the latest changes from master and resolved 2 conflicts.

What do you think about the changes?

w4tsn:
So now it'll only work to have either one or the other. Is there a case where both are supplied and should be used in parallel? But still if there is, that can still be added in a later PR.

For the moment being I think that looks good and I want to thank you for helping me out there.

What is still missing are MQTT5 config and validator classes, but I'm not sure if they are actually required for this to work. In this case I think this PR is good to go.

thjaeckle:
Yes, using the combination of both seemed not plausible to me..

Could you already test the mqtt5 connectivity against a broker? 
I did not find the time yet myself and doing the smoke test would be a good idea before merging.

w4tsn:
Ok, I ran the manual test just now and unfortunately the connection type `mqtt-5` as well as `mqtt-3` is unknown, so I'll have to investigate. I think you were right, it's probably better to plan this feature for 1.1 instead of 1.0.

w4tsn:
@thjaeckle I can't get to run ditto after the merge anymore. It's complaining about

`Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project ditto-utils-jsr305: Fatal error compiling: invalid target release: 11`

Is there something broken on the current master branch? I'm also not able to build and run on the current master.

thjaeckle:
Hi @w4tsn
This week we merged the Java 11 support on master.
So you need a jdk 11 in order to compile Ditto now. That is also mentioned in the root Readme.md
Source levels for all of the api relevant modules is still on Java 8, so that they may be used in projects still on Java 8.

w4tsn:
Hi @thjaeckle,

I'm feeling a bit bad about asking you this. I'm kind of stuck at the point where ditto complains about the connection type being unknown and I'm a bit lost here. I've setup 'MQTT5' and 'MQTT3' with 'mqtt-5' and 'mqtt-3' as type config options everywhere, where also 'mqtt' is configured in the code.

What am I missing here? Do I have to mention the mqtt connection type also in some configuration file?

thjaeckle:
The `ConnectionType` enum should be the only place where it is defined.
Did you rebuild and restart all the services? Because the "gateway" where the "devops command" is sent to would also need to be aware that the enum was enhanced by an "mqtt-5" value.
That would be my guess.

thjaeckle:
Hi @w4tsn 
I had a look into it when starting in the IDE. It is the validator, namely the `org.eclipse.ditto.services.connectivity.messaging.mqtt.MqttValidator` which defines in its `type()` method to reflect the "mqtt" connectionType only.
That way "mqtt-3" and "mqtt-5" are not registered as validators.

`ConnectionPersistenceActor`, line 196 registers all known connection type validators, so that would also have to register one for "mqtt-3" and one for "mqtt-5".
I suggest making the MqttValidator an abstract class with 3 concrete implementations (potentially differing in the behavior of mqtt 3 vs. 5:
* MqttValidator
* Mqtt3Validator
* Mqtt5Validator

and registering all of those in the ConnectionPersistenceActor.

thjaeckle:
Ah, for example the `MqttValidator.validateSource()` checks if there is a header mapping present and throws an exception that this is not supported for MQTT (version 3).
So that is a difference which must be reflected in a Mqtt5Validator.

w4tsn:
Thank you Thomas, this will help me a lot.

thjaeckle:
Hi @w4tsn 
Do you think you find the time to get this PR ready for Ditto 1.1.0 (in roughly a month)?
Really looking forward to this :)

w4tsn:
I'll give it a another shot the following weekend and see if I'll be able to fix the remaining bugs. I'm currently a bit lost but I suppose I can tell next week if this PR will be ready for 1.1.0 or if not ;)

I'm still eager to finish this one as soon as possible ;)

w4tsn:
@thjaeckle this feature is almost done. The mqtt5 implementation is working and the tests for it are passing. I still have one failing test though with the mqtt3 compatibility `mqtt-3` I'm figuring out right now before merging the latest master and submitting the final state of the PR.

The failing test is `testValidSourceAddress` which is failing because of assigned headers, which should not be assigned. I'm still figuring out why this is not working for `mqtt-3` although it's working with `mqtt` which are both using the same classes underneath.

The error message is `Header mapping is not supported for MQTT sources.` in case you have a guess on what's wrong here. I think I can fix this soon.

thjaeckle:
Hi @w4tsn 
This error message is produced in the `MqttValidator`.
Did you create a new validator for MQTT5? Then you should delete the following checks (as MQTT5 does support headers):

In `validateSource(Source, DittoHeaders, Supplier<String>)`:
```
if (source.getHeaderMapping().isPresent()) {
  ...
```

In `validateTarget(Target, DittoHeaders, Supplier<String>)`:
```
if (target.getHeaderMapping().isPresent()) {
  ...
```

w4tsn:
@thjaeckle yes I did that. The strange thing is, that `mqtt-5` is working as intended. `mqtt-3` however is not, while `mqtt` is. Both last ones use the same Actor classes, tests and the same validator, but there still has to be a difference.

thjaeckle:
Ah, did you add a test for "mqtt-5" and "mqtt-3" connection types?
Maybe the test for "mqtt-3" tries to configure a source with header mapping?

w4tsn:
@thjaeckle After merging with the latest changes there are test-cases in AbstractConsumerActorTest which fail for a reason I'm not able to understand right now.

The test cases are `testInboundMessageWithHeaderMappingThrowsUnresolvedPlaceholderException` and `testInboundMessageFailsIfHeaderIsMissing`. They fail because of a 3-second timeout.

```
[ERROR] Failures: 
[ERROR]   HiveMqtt5ConsumerActorTest>AbstractConsumerActorTest.testInboundMessageFailsIfHeaderIsMissing:130->AbstractConsumerActorTest.testInboundMessage:175->AbstractConsumerActorTest.testInboundMessage:187 assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.eclipse.ditto.services.connectivity.messaging.BaseClientActor$PublishMappedMessage
[ERROR]   HiveMqtt5ConsumerActorTest>AbstractConsumerActorTest.testInboundMessageWithHeaderMappingThrowsUnresolvedPlaceholderException:157->AbstractConsumerActorTest.testInboundMessage:175->AbstractConsumerActorTest.testInboundMessage:187 assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.eclipse.ditto.services.connectivity.messaging.BaseClientActor$PublishMappedMessage
```

Any Idea what's wrong?

I've now verified in a local ditto instance, that I'm able to connect to an MQTT5 Broker with the `mqtt-5` connection type and that I'm able to add header mapping to that connection.

I've removed the `mqtt-3` connection type, because I want to get this done and honestly I'm not patient enough to pursuit what's wrong there right now :)

w4tsn:
And yup. I've screwed my branch and I probably have to make a new one :S

thjaeckle:
I'll have to look into the failing tests (maybe once you have the branch in a clean state again :D).
I guess we changed/added some tests in the meantime and this is not yet reflected here.

w4tsn:
@thjaeckle clean-up done.

thjaeckle:
@w4tsn I fixed the failing test - but did not resolve the merge conflict, so the build is not yet triggered in order to verify

w4tsn:
@thjaeckle nice. I reviewed your merge and it looks OK to me.

This PR is done for now from my side. Certain mqtt 5 features I mentioned in the issue are left out for more contributions. I think the basic functionality and reason for a mqtt 5 client - the header mapping - is provided now. What do you think? Is there anything open from your side?

What's missing is a full integration test with a real mqtt 5 broker, including header mapping for message from and to ditto. I just tested connection and header mapping definition to the public mqtt test server.

Finally shall I compress all commits into a single one as it is specified in the Eclipse Committers Agreement?

thjaeckle:
@w4tsn cool, IMO this is a great Mqtt5 support already, even without the certain Mqtt5 features you mentioned.
So nothing open from my side, but we did not yet fully review the code. 

I would suggest that you indeed squash into 1 commit (also because I changed my email address in the meantime which causes the ECA check to fail). 
Then, we'll have to file a CQ at Eclipse for an external contribution above 1000 LoC.
While this is reviewed by the Eclipse Foundation, we can do the final code review and maybe request some last small adjustments.

I think it will make it into 1.1.0 because we probably won't be able to deliver on the provided date, but a little bit later.

Is that OK for you?

w4tsn:
@thjaeckle that sound great to me. So I've already squashed my commits and updated the PR accordingly. Looking forward to the reviews and proposals!

thjaeckle:
@w4tsn during your squash a compile error found its way into the code:
```[ERROR] /home/runner/work/ditto/ditto/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/mqtt/AbstractMqttValidator.java:[135,38] cannot find symbol
  symbol:   method applyThingPlaceholderToAddresses(java.util.Set<java.lang.String>,org.eclipse.ditto.model.things.ThingId,(filter)->[...](); })
  location: class org.eclipse.ditto.model.placeholders.PlaceholderFilter
```

I guess you need to copy the 2 static methods which were added to the MqttValidator up to the AbstractMqttValidator:
https://github.com/eclipse/ditto/blob/master/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/mqtt/MqttValidator.java#L318

The name and signature of those methods changed.

thjaeckle:
Created  CQ for contribution above 1.000 LoC: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21809

w4tsn:
@thjaeckle pretty cool! Thanks for the guidance and improvements along the way, this means a lot to me :tada:

thjaeckle:
CQ approved that the code may be checked in

PR #555, opened 2019-11-25
use "/sbin/tini" as start command wrapping invocation of "java"

thjaeckle:
The reason is that otherwise "java" will be started with PID 1 which causes that PID 1 is not terminated when the docker daemon stops the container (e.g. during a rolling update or rebalancing, etc.).

"tini" forwards the SIGTERM command to the "java" process which then has the chance to shutdown itself gracefully.

### DIFF
--- a/deployment/helm/eclipse-ditto/Chart.yaml
+++ b/deployment/helm/eclipse-ditto/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.0.0-M3"
 description: A Helm chart for Eclipse Ditto
 name: eclipse-ditto
 version: 0.6.1
-home: www.eclipse.org/ditto
+home: https://www.eclipse.org/ditto
 sources:
 - https://github.com/eclipse/ditto
 - https://github.com/kiwigrid/helm-charts/tree/master/charts/eclipse-ditto
@@ -11,3 +11,5 @@ icon: https://www.eclipse.org/ditto/images/ditto.svg
 maintainers:
 - name: axdotl
   email: axel.koehler@kiwigrid.com
+- name: thjaeckle
+  email: thomas.jaeckle@bosch-si.com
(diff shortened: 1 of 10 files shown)


axdotl:
Instead of adjusting the start command within the k8s manifests, the Docker images itself could be adjusted (e.g. by using https://github.com/krallin/tini). As a result, the users don't have to take care about "starting containers properly". Other environments (e.g. OpenShift) might benefit from this as well.

thjaeckle:
Ah, ok .. so using the `ENTRYPOINT` is the magic sauce:

E.g.:
```
ENTRYPOINT ["/tini", "--"]

CMD ["/your/program", "-and", "-its", "arguments"]
```

We shall do that and adjust this PR accordingly.

deployment/helm/eclipse-ditto/templates/concierge-deployment.yaml
```diff
@@ -47,9 +47,8 @@ spec:
         - name: {{ .Chart.Name }}-concierge
           image: "{{ .Values.concierge.image.repository }}:{{ .Values.concierge.image.tag }}"
           imagePullPolicy: {{ .Values.concierge.image.imagePullPolicy }}
-          command: ["/sbin/tini"]
+          command: ["java"]
```
axdotl: Unfortunately this will override the Docker entrypoint. Property naming is misleading (see https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes)
`java` should be moved to `args` and `command` field can be removed.

thjaeckle: Thx for the hint - now it should finally be correct :D

deployment/helm/eclipse-ditto/Chart.yaml
```diff
@@ -3,11 +3,13 @@ appVersion: "1.0.0-M3"
 description: A Helm chart for Eclipse Ditto
 name: eclipse-ditto
 version: 0.6.1
```
axdotl: Version number should be increased to 0.7

thjaeckle:
Thx @axdotl for your input

PR #583, opened 2019-12-19
Connection with empty credentials

jokraehe:
Fixes an issue when creating a connection with empty credentials. Creating a connection with credentials without key and certificate resulted in a timeout.

### DIFF
--- a/model/connectivity/src/main/java/org/eclipse/ditto/model/connectivity/ClientCertificateCredentials.java
+++ b/model/connectivity/src/main/java/org/eclipse/ditto/model/connectivity/ClientCertificateCredentials.java
@@ -50,7 +50,7 @@ public final class ClientCertificateCredentials implements Credentials {
 
     @Override
     public <T> T accept(final CredentialsVisitor<T> visitor) {
-        return visitor.get(this);
+        return visitor.clientCertificate(this);
     }
 
     /**
--- a/model/connectivity/src/main/java/org/eclipse/ditto/model/connectivity/ClientCertificateCredentials.java
+++ b/model/connectivity/src/main/java/org/eclipse/ditto/model/connectivity/ClientCertificateCredentials.java
@@ -50,7 +50,7 @@ public final class ClientCertificateCredentials implements Credentials {
 
     @Override
     public <T> T accept(final CredentialsVisitor<T> visitor) {
-        return visitor.clientCertificate(this);
+        return visitor.get(this);
     }
 
     /**
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/internal/ssl/KeyManagerFactoryFactory.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/internal/ssl/KeyManagerFactoryFactory.java
@@ -32,6 +32,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.net.ssl.KeyManagerFactory;
 
 import org.eclipse.ditto.json.JsonPointer;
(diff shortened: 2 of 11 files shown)


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/internal/ssl/KeyManagerFactoryFactory.java
```diff
@@ -171,16 +172,27 @@ private static String pemRegex(final String label) {
...
         if (clientKeyPem != null && clientCertificatePem != null) {
             return newKeyManagerFactory(clientKeyPem, clientCertificatePem);
         } else {
-            throw exceptionMapper.fatalError("Either client key or certificate were missing")
-                    .build();
+            throw exceptionMapper.fatalError("Either client key or certificate were missing").build();
+        }
+    }
+
+    @Override
```
thjaeckle: Shouldn't we add `@Nullable` here?

model/connectivity/src/main/java/org/eclipse/ditto/model/connectivity/ClientCertificateCredentials.java
```diff
@@ -50,7 +50,7 @@ private ClientCertificateCredentials(@Nullable final String clientCertificate,
 
     @Override
     public <T> T accept(final CredentialsVisitor<T> visitor) {
-        return visitor.clientCertificate(this);
+        return visitor.get(this);
```
thjaeckle: Hm, as this `get(this)` is `@Nullable` - wouldn't that break the contract of `Credentials.accept(CredentialsVisitor)` which is `@NonNull`?

So instead of this, should we still use the `clientCertificate(this)` and "un-deprecate"?

services/connectivity/messaging/src/test/java/org/eclipse/ditto/services/connectivity/messaging/internal/ssl/AbstractSSLContextTest.java
```diff
@@ -245,7 +245,7 @@ private ServerSocket startServer(final boolean needClientAuth, final ClientCerti
 
         final SSLServerSocket serverSocket =
                 (SSLServerSocket) SSLContextCreator.of(Certificates.CA_CRT, null, null)
-                        .clientCertificate(credentials)
+                        .get(credentials)
```
thjaeckle: Potential NPE as `.get(credentials)` is `@Nullable`

services/connectivity/messaging/src/test/java/org/eclipse/ditto/services/connectivity/messaging/httppush/HttpPushClientActorTest.java
```diff
@@ -183,7 +183,7 @@ public void testTLSConnectionFails() {
                 .clientCertificate(TestConstants.Certificates.CLIENT_SELF_SIGNED_CRT)
                 .build();
         final SSLContext sslContext = SSLContextCreator.fromConnection(connection, DittoHeaders.empty())
-                .clientCertificate(credentials);
+                .get(credentials);
```
thjaeckle: Potential NPE in `ConnectionContext.https(SSLContext)` as `.get(credentials)` is `@Nullable`

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/internal/ssl/SSLContextCreator.java
```diff
@@ -136,7 +144,7 @@ public SSLContext clientCertificate(@Nonnull final ClientCertificateCredentials
      * @return the SSL context
      */
     public SSLContext withoutClientCertificate() {
-        return clientCertificate(ClientCertificateCredentials.empty());
+        return get(ClientCertificateCredentials.empty());
```
thjaeckle: Either Method must be `@Nullable` or use the `clientCertificate()` method

PR #587, opened 2020-01-08
Fix payload mapping for multiple targets

dguggemos:
If multiple targets of a connection defined different payload mappings, only the first mapping was applied.

### DIFF
--- a/services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignal.java
+++ b/services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignal.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
+Copyright (c) 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
--- a/services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignal.java
+++ b/services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignal.java
@@ -12,6 +12,9 @@
  */
 package org.eclipse.ditto.services.models.connectivity;
 
+import static org.eclipse.ditto.model.base.common.ConditionChecker.checkArgument;
+import static org.eclipse.ditto.model.base.common.ConditionChecker.checkNotNull;
+
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Predicate;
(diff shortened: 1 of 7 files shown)


services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignal.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
thjaeckle:
```suggestion
Copyright (c) 2019 Contributors to the Eclipse Foundation
```

services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/MappedOutboundSignal.java
```diff
@@ -68,27 +68,24 @@ public JsonObject toJson(final JsonSchemaVersion schemaVersion, final Predicate<
...
     @Override
     public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (!(o instanceof MappedOutboundSignal)) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
jufickel-b: The parentheses got lost somehow.

services/models/connectivity/src/main/java/org/eclipse/ditto/services/models/connectivity/OutboundSignal.java
```diff
@@ -60,6 +61,18 @@ default JsonObject toJson(final JsonSchemaVersion schemaVersion, final JsonField
         return toJson(schemaVersion, FieldType.notHidden()).get(fieldSelector);
     }
 
+    /**
+     * Extends the {@link OutboundSignal} by adding the payload mapping used to map the signal.
+     */
+    interface Mappable extends OutboundSignal {
```
jufickel-b: We should move the Mappable interface to its own file and rename it to MappableOutboundSignal. Nesting it in its super-type confused me at first.

services/models/connectivity/src/test/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignalTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.eclipse.ditto.model.connectivity.PayloadMapping;
+import org.junit.Test;
+import org.mutabilitydetector.unittesting.AllowedReason;
+import org.mutabilitydetector.unittesting.MutabilityAssert;
+import org.mutabilitydetector.unittesting.MutabilityMatchers;
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+
+public class MappableOutboundSignalTest {
```
jufickel-b: Is it intended that this class can be sub-classed? If so, it should be abstract or else final.

services/models/connectivity/src/test/java/org/eclipse/ditto/services/models/connectivity/MappableOutboundSignalTest.java
```diff
@@ -0,0 +1,38 @@
...
+                AllowedReason.provided(OutboundSignal.class, PayloadMapping.class).areAlsoImmutable());
+    }
+
+    @Test
+    public void testHashCodeAndEquals() {
+        EqualsVerifier.forClass(MappableOutboundSignal.class)
+                .usingGetClass()
+                .verify();
+    }
+
```
jufickel-b: I suggest to add a test case for getting the payload mapping as this is the core functionality of the tested class.

PR #596, opened 2020-01-16
Add a protocol header for the subject who caused an event.

yufei-cai:
Add the protocol header `ditto-originator` that is set for all outgoing messages to the subject who caused the sending of the message.

Motivation: The cause of any message is available for connections via the header mapping placeholder `{{ request:subjectId }}`. This information is not available over Websocket, where header (or payload) mapping is not possible.


yufei-cai:
TODO: merge master

thjaeckle:
Merged master and resolved test failure.

PR #598, opened 2020-01-20
Serialization improvements through CBOR

erikescher:
This pull request adds another serialization format and sets it as the default for internal communication.

Changing to this format increases throughput by about 10% in synthetic tests and improves response times in almost all test cases.

Dependencies:
This PR introduces a dependency on jackson-core and jackson-dataformats-cbor, but does not depend on jackson-databind (except in a Maven "provided" scope for internal tests by Jackson).

Upgrading:
Rolling upgrades are possible, if the new serializer is first declared in one upgrade, while maintaining the old serialization bindings and then bound and therefore enabled in a separate upgrade.


erikescher:
> 
> 
> Hmm, could we somehow make the Jackson dependencies optional to `ditto-json` (e.g. "provided" scope)?
> Because it would be of course helpful to have the CBOR serialization in the Ditto backend, but in the Ditto client (which also transitively pulls `ditto-json`) I don't see that we require them.

For Ditto, that should work. However anyone using the CBOR features in the client would then have to provide the dependencies manually or risk runtime errors.

This implementation also uses CBOR as the default for the internal SoftReference based caching in JsonObjects, which would trigger such runtime exceptions. However at least those could be overcome by making the default behaviour configurable. The risk of someone accidentally using CBOR features remains though.

To solve this we'd likely have to maintain two implementations of JsonValues, one that supports CBOR and one that does not.

thjaeckle:
> However at least those could be overcome by making the default behaviour configurable. The risk of someone accidentally using CBOR features remains though.
I think we should do that, yes.

E.g. by default when plain using the Ditto json, model, signals, etc. use the non-CBOR variant and by e.g. setting a defined system property using the CBOR based variant (which we would use by default in the Ditto backend services).
I don't think "accidentially" using the CBOR feature is a problem - if a user uses a CBOR class/method from the json or model it is ok that a `NoClassDefFoundError` is thrown, (s)he has to simply add the jackson dependencies then.

thjaeckle:
Created CQs:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21501
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21502

erikescher:
I implemented the changes requested above, however instead of a system property ditto-json now checks, if the necessary classes can be loaded and uses CBOR based on the result of that test.

This means that any JAR that provides the necessary dependencies will use CBOR, while others will continue to use JSON and run into Runtime Errors when attempting to use CBOR related functionality.

thjaeckle:
CQs were approved

thjaeckle:
Created CQ for external code contribution: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21533

thjaeckle:
Code contribution CQ was approved based on Parallel IP.
We can merge whenever we choose to.

services/utils/config/src/main/resources/ditto-akka-config.conf
```diff
@@ -102,7 +102,6 @@ akka {
       # Serialize Jsonifiable events with custom JSON serializer:
       "org.eclipse.ditto.model.base.json.Jsonifiable" = cbor
       "org.eclipse.ditto.model.base.exceptions.DittoRuntimeException" = cbor
-      "org.eclipse.ditto.signals.commands.devops.DevOpsCommandResponse" = json # to ensure readability
```
erikescher: @Yannic92 you´ll want to maintain this line as developers or applications using DevOpsCommands will otherwise receive CBOR responses to their commands.
This would make interacting with a running system a lot more difficult and will cause issues with internal tools that rely on the response being JSON.

thjaeckle: @erikescher will they? 
DevOpsCommand responses are always handled at HTTP level via gateway's `AbstractHttpRequestActor` which should use the JSON representation anyway, no matter how the message was serialized and deserialized in the cluster.

Yannic92: I also verified this by setting it to "cbor". Still receiving well formed JSON at the HTTP API.

PR #602, opened 2020-01-24
Bugfix websocket response + Update license headers + Minor improvement in connectivity service

Yannic92:
This PR fixes an issue where ditto sends a response to a command sent over websocket even if the "response-required" header was false.

In addition this PR updates the license headers to match 2020 and allows to start child actors in connectivity root actor.

### DIFF
--- a/services/base/src/main/java/org/eclipse/ditto/services/base/actors/DittoRootActor.java
+++ b/services/base/src/main/java/org/eclipse/ditto/services/base/actors/DittoRootActor.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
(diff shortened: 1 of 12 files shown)


services/base/src/main/java/org/eclipse/ditto/services/base/actors/DittoRootActor.java
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.ditto.services.base.actors;
+
+import org.eclipse.ditto.services.utils.akka.logging.DittoDiagnosticLoggingAdapter;
+import org.eclipse.ditto.services.utils.akka.logging.DittoLoggerFactory;
+
+import akka.actor.AbstractActor;
+import akka.actor.Props;
+import akka.japi.pf.ReceiveBuilder;
+
+public abstract class DittoRootActor extends AbstractActor {
```
jokraehe: Please document this class and use it where possible. Generalize as much as possible with this class, e.g. supervision.

Issue #561, opened 2019-12-03
Enrich outgoing events/messages with additional fields from Thing


thjaeckle:
Currently, whenever a [change notification](https://www.eclipse.org/ditto/basic-changenotifications.html) / event is published (e.g. via WebSocket or the connectivity - AMQP, MQTT, Kafka, HTTP), only the actually changed values are contained in that published event.

So e.g. when only a temperature is updated (this is the Ditto Protocol message):
```json
{
  "topic": "org.eclipse.ditto/my-fancy-car/things/twin/commands/modify",
  "path": "/features/temperature/properties/status/value",
  "value": 23.5
}
```
Then a change event will only include exactly that information, as Ditto Protocol message:
```json
{
  "topic": "org.eclipse.ditto/my-fancy-car/things/twin/events/modified",
  "path": "/features/temperature/properties/status/value",
  "value": 23.5
}
```

There are however cases when there should be more information added to that event, e.g. when some additional metadata is needed when the event is processed.

In such cases it would of course work to do a HTTP GET on the Ditto HTTP API in order to fetch that additional information, but that causes additional roundtrips and additional latency which could be avoided.

So we would like to add the possibility to specify a further [field selector](https://www.eclipse.org/ditto/httpapi-concepts.html#with-field-selector) of additional thing data to select for all channels we send change notifications on:
* WebSocket
* Server Sent Events (SSEs)
* all connectivity types (AMQP, MQTT, ...)

For WebSocket that could e.g. look like:
```
START-SEND-EVENTS?extraFields=attributes,features/location
```

The actual change should/could be "merged" with the additionally selected fields, so the resulting Ditto Protocol events could look like:
```json
{
"topic": "org.eclipse.ditto/my-fancy-car/things/twin/events/modified",
"path": "/features/temperature/properties/status/value",
"value": 23.5,
"extra": {
  "attributes": {
    "manufacturer": "ACME corp"
  },
  "features": {
    "location": { 
      "longitude": 42.123,
      "latitude": 3.54
    }
  }
}
}
```

Messages and live commands/events should also be able to be enriched.

thjaeckle:
One additional idea in order to prevent many remoting roundtrips in the Ditto cluster:

We could cache Things at the Ditto "edges" (gateway, connectivity services).
Apply a similar approach to the policy cache in concierge:
* use the Caffeine cache + cache loader in order to load non resolved Things
* the Caffeine cache has to be scoped to a single connection or session
[...]

PR #612, opened 2020-02-03
Signal enrichment when publishing events/messages

thjaeckle:
Fixes: #561 

This PR adds the ability to "enrich" events (twin & live) and (live) messages with additional `extraFields` from the related thing.
E.g. when just the temperature `value` of a feature inside a thing was modified, additional fields like for example the thing's `attributes` may be enriched - and those are included in the published message then.

This is supported for all channels capable of pushing out events/messages:
* WebSocket
* SSE
* Connections


BobClaerhout:
Nice! looking forward to this

PR #613, opened 2020-02-04
Specify parameter type of options for /search/things as string

Yannic92:
This is done because in ditto we don't support multiple headers with the same name


stmaute:
LGTM :+1:

PR #622, opened 2020-02-26
Revamp background sync for guaranteed eventual consistency

yufei-cai:
This patch replaces the timestamp-based background synchronization of the search index by an ID-based mechanism so that the search index will be consistent eventually no matter the starting database state.

- Replaced background sync. The new background sync actor has progress reporting and runtime config modification functions.

- Added negative feedback from search persistence to thing updaters.

- Added dissync-detection for query processing: Things known to be out-of-sync are reported at query time for update.

- Replaced manual updater by a DevOps command to force search index update for a thing.


thjaeckle:
@yufei-cai this fix can be applied without doing any DB migration, can't it?

yufei-cai:
Correct, no DB migration is needed.

However, the following colletions are not used any more:
```
searchDB.searchThingsSyncThings
searchDB.searchThingsSyncPolicies
```

PR #623, opened 2020-02-26
Adds policy management to Ditto Protocol

### LINKED ISSUE #554: Extend Ditto Protocol to support management of policies
The [Ditto Protocol](https://www.eclipse.org/ditto/protocol-overview.html) currently supports CRUD operations on things. We will extend the protocol to also support CRUD operations on policies and its entries and subjects. The protocol can then be use via various [bindings](https://www.eclipse.org/ditto/protocol-bindings.html) e.g. WebSocket, AMQP1.0, etc. to manipulate policies

### LINKED ISSUE #599: Overloaded protocol adapter methods put live signals in the TWIN channel
These methods produce twin signals even when given a live signal as argument:
```
ProtocolAdapter#toAdaptable(Event)
ProtocolAdapter#toAdaptable(Command)
```
There may be more. This is not intuitive.

Propose to deprecate the overloaded methods and leave only
```
ProtocolAdapter#toAdaptable(Signal)
```
All other `toAdaptable` methods should simply delegate to this method. `ProtocolAdapter#toAdaptable(Signal)` does put signals in the correct channel.

What do you guys think?

dguggemos:
Fixes #554.
Fixes #599.


ffendt:
I force-pushed multiple times since I failed setting my correct mail-address in my commits :pensive:

yufei-cai:
Fixes #599 also.

PR #624, opened 2020-02-27
Catch exception in WebSocket session

ffendt:
Catch exceptions (thrown by getAuthorizationContext) when refreshing the JWT authentication in WebSocket sessions.


stmaute:
LGTM :+1:

PR #646, opened 2020-04-07
Filter headers in all responses

stmaute:
To improve the security of ditto we need to filter headers in all of our responses.
Impact of this PR:
- no custom headers are included in a response.


DoktorDaveJoos:
Reviewed. Ready for merge.

PR #650, opened 2020-04-17
Add Content-Type header validation

jbartelh:
To improve ditto's overall security, the 'Content-Type' header should be validated. Therefore resources who only support JSON, should only accept requests with a 'application/json' media-type.
For CORS requests, this makes the necessary 'preflight' request obligatory.

- Check content-type on all PUT and POST API requests and reject others that the supported type (mainly "application/json")
- "Messages" HTTP API must still support arbitrary content types
- Additional media-types can be configured for compatibility
  - Also that requests without content-type can still be accepted (currently default)


services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
```diff
@@ -0,0 +1,111 @@
...
+             final DittoHeaders dittoHeaders,
+             final Supplier<Route> inner) {
+         return enhanceLogWithCorrelationId(dittoHeaders.getCorrelationId(), () -> {
+             final String requestsMediaType = extractMediaType(ctx.getRequest());
+
+             if(supportedMediaTypes.contains(requestsMediaType)){
+                 return inner.get();
+             } else {
+                 LOGGER.info("Request rejected: unsupported media-type: <{}>  request: <{}>",
+                         requestsMediaType, requestToLogString(ctx.getRequest()));
```
thjaeckle, line 62: the `requestToLogString` method is invoked even if the LOGGER does not log on `INFO` level.
In such cases it is better to put a `LOGGER.isInfoEnabled()` around the log statement.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
```diff
@@ -0,0 +1,111 @@
...
+
+     protected static Route completeWithMediaTypeNotSupported(final String mediaType,
+             final Set<String> supportedByResource){
+
+             final String msgPatten = "The Media-Type <{0}> is not supported for this endpoint. Allowed " +
+                     "Media-Types are: <{1}>";
+             final String responseMessage =
+                     MessageFormat.format(msgPatten, mediaType, supportedByResource);
+
+             return complete(StatusCodes.UNSUPPORTED_MEDIA_TYPE, responseMessage);
```
thjaeckle, line 76: Shouldn't we return a nice (new) DittoRuntimeException JSON here?

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
```diff
@@ -0,0 +1,111 @@
...
+      * @param inner route to wrap.
+      * @return the wrapped route.
+      */
+     public static Route ensureValidContentType(final Set<String> supportedMediaTypes, final RequestContext ctx,
+             final DittoHeaders dittoHeaders,
+             final Supplier<Route> inner) {
+         return enhanceLogWithCorrelationId(dittoHeaders.getCorrelationId(), () -> {
+             final String requestsMediaType = extractMediaType(ctx.getRequest());
+
+             if(supportedMediaTypes.contains(requestsMediaType)){
```
thjaeckle, line 58: Does this also work with content-types which define the charset?
I did not see at first sight any logic handling the charset part in a Content-Type..

jbartelh: The media-type is the content-type without the charset. That's how i omit the charset to avoid complications. Only in the case where an illegal/non-parsable content-type should be allowed, the configured whitelisted value has to match the raw-header equally with charset.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
jbartelh: Yes and the OpenAPI docs should then also be updated, with the status code and the new error code, right?

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
jbartelh: Good point, thanks.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
thjaeckle: Hm, unsure. That would be a massive duplication. Documenting this 405 status code seems not to have much value for the api user.
I think it's sufficient to have a good error response message.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/directives/ContentTypeValidationDirective.java
```diff
@@ -0,0 +1,111 @@
...
+      * and {@link akka.http.scaladsl.model.HttpEntity$}.
+      * @see
+      * <a href="https://doc.akka.io/docs/akka-http/current/common/http-model.html#http-headers">Akkas Header model</a>
+      *
+      * @param request the request where the media type will be extracted from.
+      * @return the extracted media-type.
+      */
+     protected static String extractMediaType(HttpRequest request){
+         final Optional<HttpHeader> rawContentType = StreamSupport.stream(request.getHeaders().spliterator(), false)
+                 .filter(header -> header.name().equals(DittoHeaderDefinition.CONTENT_TYPE.getKey()))
```
thjaeckle, line 104: hm, here you use the `rawContentType` as preferred one which could still contain a charset, doesn't it?

jbartelh: yes it does. But unfortunately akka's header handling is not quite obvious. If the content-type can be parsed by akka, then there is no 'content-type' raw-header. It tried to document it as good as i can and also linked to the akka documentation.

PR #654, opened 2020-04-22
Fixed misleading Error Messages when IP is rejected.

JulianFeinauer:
IPs are not only rejected due to being blacklisted but also for other reasons.


thjaeckle:
You're right, that is misleading.
The default will be changed in the upcoming Ditto 1.1.0 to not block "localhost" and other local IPs by-the-way.

Could you please sign an Eclipse ECA (Committer Agreement) and then also sign off your commit like mentioned in the [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md)

JulianFeinauer:
@thjaeckle fixed it, is green now : )

PR #660, opened 2020-04-27
Avoid 304 Not Modified when policy is in selected fields

joosdavid:
Ditto should not return 304 Not Modified when policy is in selected fields of a things request. Because things revision will not increment when policy is updated, things entity including policy could be updated but ditto is returning 304 not modified.


services/utils/conditional-headers/pom.xml
```diff
@@ -33,6 +33,16 @@
             <groupId>org.eclipse.ditto</groupId>
             <artifactId>ditto-signals-commands-base</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.ditto</groupId>
+            <artifactId>ditto-signals-commands-things</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
thjaeckle, line 41: Duplicated `<dependency>` block. Please remove one of them.

joosdavid: Done

services/utils/conditional-headers/src/main/java/org/eclipse/ditto/services/utils/headers/conditional/ConditionalHeadersValidator.java
```diff
@@ -95,13 +107,36 @@ public void checkConditionalHeaders(final Command command,
...
-                        Command.Category.QUERY.equals(command.getCategory()));
+                        Command.Category.QUERY.equals(command.getCategory())) || skipDueToRelation(command);
+    }
+
+    /**
+     * Skip precondition check if the selected fields contain exempted fields (e.g. {@code _policy} for things
+     * because the revision of a thing does not change if its policy is updated).
+     */
+    private boolean skipDueToRelation(final Command command) {
+        return command.toJson().getValue(SELECTED_FIELDS)
```
thjaeckle, line 125: Did you consider using `instanceof` here in order to avoid the `toJson()` and again the parsing of the field selector later.

E.g.: 
```
@Nullable final JsonFieldSelector selectedFields;
if (command instanceof RetrieveThing) {
  selectedFields = ((RetrieveThing) command).getSelectedFields().orElse(null);
} else if (command instanceof RetrieveThings) {
  selectedFields = ((RetrieveThings) command).getSelectedFields().orElse(null);
} else {
  selectedFields = null;
}
```

services/utils/conditional-headers/src/main/java/org/eclipse/ditto/services/utils/headers/conditional/ConditionalHeadersValidator.java
```diff
@@ -95,13 +107,36 @@ public void checkConditionalHeaders(final Command command,
...
         checkIfNoneMatch(command, currentETagValue);
+
     }
 
     private boolean skipPreconditionHeaderCheck(final Command command, @Nullable final EntityTag
             currentETagValue) {
         return currentETagValue == null &&
                 (Command.Category.DELETE.equals(command.getCategory()) ||
-                        Command.Category.QUERY.equals(command.getCategory()));
+                        Command.Category.QUERY.equals(command.getCategory())) || skipDueToRelation(command);
```
thjaeckle, line 117: I would appreciate making this statement clearer as now I have to know which binary operator has more weight here.
This would make it more clear:
```suggestion
        return (currentETagValue == null && (
                Command.Category.DELETE.equals(command.getCategory()) ||
                Command.Category.QUERY.equals(command.getCategory()))
        ) || skipDueToRelation(command);
```

services/utils/conditional-headers/src/main/java/org/eclipse/ditto/services/utils/headers/conditional/ConditionalHeadersValidator.java
```diff
@@ -95,13 +107,36 @@ public void checkConditionalHeaders(final Command command,
...
+                .map(str -> JsonFactory.newFieldSelector(str, JsonFactory.newParseOptionsBuilder()
+                        .withoutUrlDecoding()
+                        .build()))
+                .map(JsonFieldSelector::getPointers)
+                .map(this::containsExemptedField)
+                .orElse(false);
+    }
+
+    private boolean containsExemptedField(final Set<JsonPointer> selectedFields) {
+        final HashSet<JsonPointer> result = new HashSet<>(EXEMPTED_FIELDS);
```
thjaeckle, line 137: we don't use the concrete `HashSet` as type for variables, but `Set` instead. That way, the implementation can easily be modifed to e.g. another `Set` implementation.

services/utils/conditional-headers/src/main/java/org/eclipse/ditto/services/utils/headers/conditional/ConditionalHeadersValidator.java
thjaeckle: The current solution probably has a worse performance.
And as e.g. one "RetrieveThings" command with page size 200 multiplies to 200 single "RetrieveThing" commands the impact on 
* creating the json
* getting a field
* creating a new field selector from it
could really make performance worse than an `instanceof` check.

thjaeckle: @dguggemos what do you think?

dguggemos: that makes sense 👍

yufei-cai:
Replaced by #664 .

PR #664, opened 2020-04-28
Avoid 304 Not Modified when retrieving a thing with policy in selected fields

joosdavid:
Assume that you have a thing with an associated policy. When querying the Thing with

`GET /things/{thingId}?fields=_policy`

you will get the Thing containing its revision and associated policy.

If you now modify the associated policy, the revision of the Thing will not change! This could lead to an
inconsistent state if the Thing is getting refetched by using the `If-None-Match` header, because this would return a `304 Not Modified`, even if the policy has changed.

To tackle this, ditto bypasses the conditional header check when `policy` is in selectedFields of a command.


thjaeckle:
@joosdavid I'll handle the test errors - has to do with maven dependencies (to things model+commands) which we should not have added to the "ditto-service-utils-conditional-headers" module - that way they are also included in "policies" modules.

That's why the Policy registry tests expect to also have Thing classes on the classpath - which should be prevented.

PR #665, opened 2020-04-28
Replace insecure rubygems proxy. Fix #625

### LINKED ISSUE #625: Replace rubygems proxy for building Ditto's website
We should get rid of the http://rubygems-proxy.torquebox.org in the website build.

Automated PR #617 did find out that we use a "http" URL in our pom.xml - and that is the mentioned http://rubygems-proxy.torquebox.org
As there is no SSL version of this proxy, we should get rid of it completely.

The following issues provide more information on that:
jruby-gradle/jruby-gradle-plugin#349
and
jruby-gradle/jruby-gradle-plugin#364

MMD-Coding:
I replaced the insecure rubygems proxy by using `mavengem-wagon` project. See https://github.com/jruby/mavengem for details.

Fixes: #625

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -197,6 +197,13 @@
                 </executions>
             </plugin>
         </plugins>
+        <extensions>
+            <extension>
+                <groupId>org.torquebox.mojo</groupId>
+                <artifactId>mavengem-wagon</artifactId>
+                <version>1.0.3</version>
+            </extension>
+        </extensions>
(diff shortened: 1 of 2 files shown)


thjaeckle:
@MMD-Coding that's really awesome. Thank you for your contribution 👍 
Will look into it (and check if the site still is built as expected) ;)

MMD-Coding:
Happy to help! :)

PR #676, opened 2020-05-08
Handle unknown hosts in connections correctly

dguggemos:
An unknown/invalid host in a connection configuration caused an exception with an error message that did not indicate the actual cause. This is now fixed.


VadimGue:
LGTM

PR #689, opened 2020-06-10
Minor improvements to throughput/performance

thjaeckle:
By applying some modifications:
* Don't validate `DittoHeaders` when already built `DittoHeaders` are passed into the `DittoHeaderBuilder` - as they were already validated, that additional validation can be omitted
* Removed `DirectivesLogginUtils` and used `DittoLogger` in directives + routes
* optimization in `HeaderTranslator` making use of the `DittoHeaders.toBuilder()` improvement


thjaeckle:
Cherry-picked commits of this PR to the "release-1.1." branch

PR #720, opened 2020-07-07
Documented that text error codes are not scope of API.

joosdavid:
The Eclipse Ditto HTTP API uses response status codes (see RFC 7231) to indicate whether a specific request has been successfully completed, or not. However, the descriptions we provide additionally to the status code might change without advance notice. These are not be considered as official API, and must therefore not be applied in your applications or tests.


stmaute:
LGTM :+1:

PR #727, opened 2020-07-09
Query parameter 'fields' is ignored when retrieving feature properties

dguggemos:


thjaeckle:
@dguggemos what about the other "Retrieve" strategies?
Is it e.g. supported for `RetrieveFeaturePropertyStrategy`?

VadimGue:
I'm checking on that. 
And the missing unit test is a good point i will add them

VadimGue:
Unit test is added.

RetrieveAttributesStrategy looks good. 
For RetrieveFeaturePropertyStrategy & RetrieveAttributeStrategy a new task was created, because it would be too much for a BugFix

Issue #680, opened 2020-05-13
Support for adding arbitrary _metadata for features / properties


JulianFeinauer:
Hi all,
if considering using some kind of History API (see #545 ) it gets more and more relevant to know the timestamp of a specific value from a thing. Also in scenarios where you have no direct connection like HTTP / WS but have messaging involved like MQTT or generally Hono you can not rely that the timestamp when you receive an update message is roughly equivalent to the timestamp when it was measured. 

I thus suggest to enable explicit support for "last_modified" timestamp which could be either set explicitly in the message or implicitly with the "ingestion time" and which is then available for a possible history service and for users to see how "valid" their datasets are.

thjaeckle:
Indeed that would be useful to have as metainformation for each property.
As we can't simply store the timestamp or any additional meta data it in the "target JSON" without modifying the user/device provided JSON structure, that would have to be stored besides the "target JSON" structure.
E.g. under a `_meta` field:
```json
{
  "thingId": "org.eclipse.ditto:thing-1",
  "policyId": "...",
  "features": {
    "lamp": {
      "properties": {
        "on": true,
        "color": {
          "r": 0,
          "g": 255,          
          "b": 255,
        }
      }
    }
  },
  "_modified": "2020-06-09T14:30:00Z",
  "_revision": 42,
  "_meta": {
    "policyId": {
      "_modified": "2020-06-09T14:00:00Z",
      "_revision": 1
    },
    "features": {
      "lamp": {
        "properties": {
          "on": {
            "_modified": "2020-06-09T14:30:00Z",
            "_revision": 42
          },
          "color": {
            "r": {
              "_modified": "2020-06-09T14:15:00Z",
              "_revision": 23
            },
            "g": {
              "_modified": "2020-06-09T14:15:00Z",
              "_revision": 23
            },
            "b": {
              "_modified": "2020-06-09T14:15:00Z",
              "_revision": 23
            }
          }
        }
      }
    }
  }
}
```

Basically each "leaf" in the JSON gets an JsonObject containing additional meta information.
That could also be aggregated upwards, so in the provided example above you would know that the "lamp" feature is also of revision `42` last modified at the timestamp `14:30`.

I would also want to see the "revision" as meta information at a property level, so that you can e.g. see (like in Subversion) which revision a part of the subtree of a Thing or even a single property has.
But that may be a topic for another issue. :)

JulianFeinauer:
Hey @thjaeckle welcome back : )
[...]

PR #745, opened 2020-07-23
Feature/680 add metadata

JulianFeinauer:
Draft of current Implementation for feedback.

Fixes: #680


model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThing.java
```diff
@@ -532,7 +546,8 @@ public boolean equals(final Object obj) {
                 Objects.equals(features, other.features) &&
                 Objects.equals(lifecycle, other.lifecycle) &&
                 Objects.equals(revision, other.revision) &&
-                Objects.equals(modified, other.modified);
+                Objects.equals(modified, other.modified) &&
+                Objects.equals(metadata, other.metadata);
```
thjaeckle, line 550: `metadata` is still missing in the `toString()` of this class

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromCopyBuilder.java
```diff
@@ -690,6 +694,17 @@ public FromCopy setModified(final Predicate<Instant> existingModifiedPredicate,
         return this;
     }
 
+    @Override
+    public FromCopy setMetadata(JsonPointer resourcePath, Metadata metadata) {
+        fromScratchBuilder.setMetadata(resourcePath, metadata);
+        return this;
+    }
+
+    public FromCopy setMetadata(Metadata metadata) {
```
thjaeckle, line 703: What is this method for - not overriding an interface method?

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromScratchBuilder.java
```diff
@@ -336,6 +337,12 @@ public FromScratch setModified(@Nullable final Instant modified) {
         return this;
     }
 
+    @Override
+    public FromScratch setMetadata(JsonPointer resourcePath, Metadata metadata) {
```
thjaeckle, line 341: the params should be `final` (its our style for parameters)

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromCopyBuilder.java
```diff
@@ -690,6 +694,17 @@ public FromCopy setModified(final Predicate<Instant> existingModifiedPredicate,
         return this;
     }
 
+    @Override
+    public FromCopy setMetadata(JsonPointer resourcePath, Metadata metadata) {
```
thjaeckle, line 698: the params should be `final` (its our style for parameters)

model/things/src/main/java/org/eclipse/ditto/model/things/MetadataBuilder.java
```diff
@@ -0,0 +1,296 @@
...
+     * @throws NullPointerException if any argument is {@code null}.
+     * @throws IllegalArgumentException if {@code name} is empty.
+     */
+    @Override
+    default MetadataBuilder set(final CharSequence key, final JsonValue value) {
+        return set(key, value, field -> true);
+    }
+
+    @Override
+    <T> MetadataBuilder set(JsonFieldDefinition<T> fieldDefinition, @Nullable T value,
```
thjaeckle, line 196: Did you not want to add a `Metadata` specific Javadoc here?

model/things/src/main/java/org/eclipse/ditto/model/things/MetadataBuilder.java
```diff
@@ -0,0 +1,296 @@
...
+                              Predicate<JsonField> predicate);
+
+    @Override
+    default <T> MetadataBuilder set(final JsonFieldDefinition<T> fieldDefinition, @Nullable final T value) {
```
thjaeckle, line 200: Did you not want to add a `Metadata` specific Javadoc here?

model/things/src/main/java/org/eclipse/ditto/model/things/MetadataModelFactory.java
```diff
@@ -0,0 +1,115 @@
...
+
+import javax.annotation.concurrent.Immutable;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.eclipse.ditto.model.base.common.ConditionChecker.checkNotNull;
+import static org.eclipse.ditto.model.base.exceptions.DittoJsonException.wrapJsonRuntimeException;
+
+@Immutable
+public final class MetadataModelFactory {
```
thjaeckle, line 28: Please add Javadoc for a public class in the model

model/things/src/main/java/org/eclipse/ditto/model/things/Thing.java
```diff
@@ -653,6 +653,13 @@ default void validate(DittoHeaders headers) { }
                 JsonFactory.newJsonObjectFieldDefinition("features", FieldType.REGULAR, JsonSchemaVersion.V_1,
                         JsonSchemaVersion.V_2);
 
+        /**
+         * JSON field containing the Thing's metadata.
+         */
+        public static final JsonFieldDefinition<JsonObject> METADATA =
+            JsonFactory.newJsonObjectFieldDefinition("_metadata", FieldType.SPECIAL, FieldType.HIDDEN,
+                JsonSchemaVersion.V_1, JsonSchemaVersion.V_2);
+
```
thjaeckle, line 662: I would move this up below the "_modified" definition to have them all in one place.

services/things/persistence/src/main/java/org/eclipse/ditto/services/things/persistence/actors/strategies/events/AbstractThingEventStrategy.java
```diff
@@ -46,9 +46,12 @@ protected AbstractThingEventStrategy() {
     @Override
     public Thing handle(final T event, @Nullable final Thing thing, final long revision) {
         if (null != thing) {
+            final String issuedAt = event.getDittoHeaders().get("_issuedAt");
+
             ThingBuilder.FromCopy thingBuilder = thing.toBuilder()
                     .setRevision(revision)
                     .setModified(event.getTimestamp().orElse(null));
+                    // .setMetadata(event.getResourcePath(), event.getEntity());
```
thjaeckle, line 54: I think these changes can be removed, right?
We would need a mechanism to generically add metadata provided in the headers (with a defined prefix) to the thing's `Metadata`.

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableMetadataBuilder.java
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: Please change the year to 2020

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromCopyBuilder.java
JulianFeinauer: @thjaeckle I was unsure about this one. This method is a shorthand and is called in the `of(JsonObject)` method as FP. I dont know if we shuld then move it up to the interface. It could even be private here, in theory or we just call the other method `(JsonPointer, Metadata)` and pass empty Pointer explicitly. What would you prefer?

services/things/persistence/src/main/java/org/eclipse/ditto/services/things/persistence/actors/strategies/events/AbstractThingEventStrategy.java
JulianFeinauer: Yes, you are right, thats from me "playing around".

JulianFeinauer:
Thank you for the review @thjaeckle I agree with all points and added more specific questions where needed. I will resolve these things in the next time and wait for another review. In the meantime I will try to take care of the ditto headers. Although we could also do it in a separate PR then things stay a bit simpler. WDYT?

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromCopyBuilder.java
thjaeckle: I would prefer passing the empty Pointer explicitly to the already existing interface method.

thjaeckle:
> In the meantime I will try to take care of the ditto headers. Although we could also do it in a separate PR then things stay a bit simpler. WDYT?

I would add this to this PR as the `Metadata` is not usable without setting it via the ditto headers. :)

JulianFeinauer:
@thjaeckle I fixed all reported issues and tried the first implementation fo "header parsing" and metadata creation from it as you suggest it in the issue.

model/things/src/main/java/org/eclipse/ditto/model/things/Thing.java
```diff
@@ -572,6 +572,13 @@ default Thing setPolicyId(@Nullable String policyId) {
     @Deprecated
     default void validate(DittoHeaders headers) { }
 
+    /**
+     * Returns the Metadata of this Thing.
+     *
+     * @return the Metadata of this Thing.
```
thjaeckle, line 578: please add `@since 1.2.0` javadoc below `@return`

model/things/src/main/java/org/eclipse/ditto/model/things/MetadataModelFactory.java
```diff
@@ -24,6 +24,9 @@
 import static org.eclipse.ditto.model.base.common.ConditionChecker.checkNotNull;
 import static org.eclipse.ditto.model.base.exceptions.DittoJsonException.wrapJsonRuntimeException;
 
+/**
+ * Factory that creates new {@code metadata} objects.
```
thjaeckle, line 28: please add `@since 1.2.0` javadoc

services/things/persistence/src/main/java/org/eclipse/ditto/services/things/persistence/actors/strategies/events/AbstractThingEventStrategy.java
```diff
@@ -46,12 +51,23 @@ protected AbstractThingEventStrategy() {
     @Override
     public Thing handle(final T event, @Nullable final Thing thing, final long revision) {
         if (null != thing) {
-            final String issuedAt = event.getDittoHeaders().get("_issuedAt");
+            // Create Metadata from Header, see https://github.com/eclipse/ditto/issues/680#issuecomment-654165747
+            // Use existing Metadata or create a new Builder if none exists
+            MetadataBuilder metadataBuilder = thing.getMetadata()
```
thjaeckle, line 56: variable can be `final`

services/things/persistence/src/main/java/org/eclipse/ditto/services/things/persistence/actors/strategies/events/AbstractThingEventStrategy.java
```diff
@@ -46,12 +51,23 @@ protected AbstractThingEventStrategy() {
...
+                .map(Metadata::toBuilder)
+                .orElse(Metadata.newBuilder());
+            for (String headerKey : event.getDittoHeaders().keySet()) {
+                if (headerKey.startsWith(DITTO_HEADER_METADATA_PREFIX)) {
+                    String metadataKey = headerKey.substring(DITTO_HEADER_METADATA_PREFIX.length());
+                    String metadataValue = event.getDittoHeaders().get(headerKey);
+                    metadataBuilder.set(JsonPointer.of(metadataKey), metadataValue);
+                }
+            }
```
thjaeckle, line 65: I'm afraid that it's not that easy.
In https://github.com/eclipse/ditto/issues/680#issuecomment-654165747 I mentioned different syntaxes for the `ditto-metadata:`:
* "ditto-metadata:issuedAt": "2020-06-09T14:29:00Z"
   * the idea of this syntax is that each JSON "leaf" which was updated in the `event` shall receive the metadata `"issuedAt": "2020-06-09T14:29:00Z"`
   * therefore, the `event.getResourcePath()` has to be taken into account (as base path of which part of the thing was changed - this will be `/` when the complete thing was modified) , combined with all pathes of the JSON leafes which can be found out via `event.getEntity(event.getImplementedSchemaVersion())`
   * a JsonObject has to be created for each JSON leaf containing the set metadata
* "ditto-metadata:/features/lamp/properties/color/r/issuedAt": "2020-06-09T14:14:00Z"
   * the idea of this syntax is to specify "overwrites" for the syntax above

For example:
You modify the feature properties of a "Lamp" feature in this Thing:
```json
{
  "thingId": "org.eclipse.ditto:my-lamp",
  "policyId": "org.eclipse.ditto:my-policy",
  "features": {
    "Lamp": { 
      "properties": {
        "on": true,
        "dimmer": 0.8
      }
    }
  }
}
```

You modify `"/features/Lamp/properties"` to:
```json
{
  "on": false,
  "dimmer": 0.0
}
```

The event looks like:
* `getResourcePath()` returns a JsonPointer with `"/features/Lamp/properties"`
*  `event.getEntity(event.getImplementedSchemaVersion())` returns the above JsonObject

If now the header `"ditto-metadata:issuedAt": "2020-06-09T14:29:00Z"` was set, we want to have the following thing JSON (incl. `_metadata`) as a result:
```json
{
  "thingId": "org.eclipse.ditto:my-lamp",
  "policyId": "org.eclipse.ditto:my-policy",
  "features": {
    "Lamp": { 
      "properties": {
        "on": false,
        "dimmer": 0.0
      }
    }
  },
  "_metadata": {
    "features": {
      "Lamp": { 
        "properties": {
          "on": {
            "issuedAt": "2020-06-09T14:29:00Z"
          },
          "dimmer": {
            "issuedAt": "2020-06-09T14:29:00Z"
          }
        }
      }
    }
  }
}
```

So the mechanism is a lot more complex and I think we should extract it to another class in order to be able to also better unit test it.

Ah, and the idea would be to **always** add the following metadata for all affected JSON leafes:
* `_revision`: is passed in via the parameter `revision`
* `_modified`: can be obtained by `event.getTimestamp().orElse(null)`

JulianFeinauer: Thanks for the pointers. Indeed it was not meant as complete implementation but a first draft. But your comment helps me. I will try to extract that to another class and add some tests to get a better understanding :)

model/things/src/main/java/org/eclipse/ditto/model/things/ImmutableThingFromCopyBuilder.java
jufickel-b: On Thing level the Pointer is unused and  thus it should be removed IMHO.

model/things/src/test/java/org/eclipse/ditto/model/things/ImmutableThingTest.java
```diff
@@ -1115,10 +1124,22 @@ public void createThingWithInvalidCharactersInId() {
...
+                                MODIFIED,
+                                METADATA)
                 )
         );
 
     }
 
+    @Test
+    public void serializeMetadata() {
+        final JsonObject json = KNOWN_THING_V2.toJson(JsonSchemaVersion.V_2, field -> true);
```
jufickel-b, line 1136: I would prefer the following implementation of that method:

    final JsonObject json = KNOWN_THING_V2.toJson(JsonSchemaVersion.V_2, field -> true);

    final Optional<JsonObject> metadataOptional = json.getValue(Thing.JsonFields.METADATA);

    assertThat(metadataOptional).contains(JsonObject.newBuilder().set("_issuedAt", JsonValue.of(0)).build());

model/things/src/main/java/org/eclipse/ditto/model/things/NullMetadata.java
```diff
@@ -0,0 +1,304 @@
...
+     *
+     * @return the new NullMetadata object.
+     */
+    public static Metadata newInstance() {
+        return new NullMetadata();
+    }
+
+    @Override
+    public boolean isBoolean() {
+        return false;
```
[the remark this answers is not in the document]
thjaeckle: the same is done for `Attributes` so I would keep it that way

thjaeckle (approved):
I did a review of the complete PR.
I renamed the header key from "metadata" to "put-metadata" to be more precise and explicit what happens with the header value. Also I think that we will in the future need some header like "get-metadata" in order to e.g. retrieve metadata when doing a HTTP GET request for e.g. a single feature or property (to include the metadata as HTTP response header).

Afterwards I merged the latest master onto the PR branch and resolved the conflicts we got and I addressed the still open topics addressed in this PR discussion done by @jufickel-b 

So my approval from technical/code side 👍 
This is a great addition to Ditto - much looking forward to the possibilities it opens.

thjaeckle:
I marked PR as ready for review, so that it can make it into Ditto 1.2.0 - the documentation can be added in another PR.

PR #753, opened 2020-07-29
Append extra fields to selected fields for SSE

Yannic92:
* There was a bug where extra fields were filtered out of the json
  representation after beeing added before. Reason was that the filtering
  is based on the selected fields. To make our API easy to use we
  automatically add all extra fields to the selected fields.

### DIFF
--- a/json/src/main/java/org/eclipse/ditto/json/JsonFieldSelector.java
+++ b/json/src/main/java/org/eclipse/ditto/json/JsonFieldSelector.java
@@ -72,4 +72,13 @@ public interface JsonFieldSelector extends Iterable<JsonPointer> {
     @Override
     String toString();
 
+    /**
+     * Creates a new JsonFieldSelector which contains the json pointers of this instance and the pointers of the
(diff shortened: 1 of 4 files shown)


json/src/main/java/org/eclipse/ditto/json/JsonFieldSelector.java
```diff
@@ -72,4 +72,13 @@ static JsonFieldSelector newInstance(final CharSequence pointerString,
     @Override
     String toString();
 
+    /**
+     * Creates a new JsonFieldSelector which contains the json pointers of this instance and the pointers of the
```
jufickel-b, line 76: json --> JSON

Yannic92:
I withdraw this PR, as the implementation works as intended right now.
Fields that are selected in extraFields need to be selected by "fields" aswell. Only if "fields" is not specified at all, this is not required.
This was done because we want to provide "extraFields" for "filter" and remove these extra fields afterwards to save bandwidth.

PR #758, opened 2020-08-04
Validate JsonKeys in commands/responses that take a plain JsonPointer

dguggemos:
Some commands e.g. ModifyAttribute take a JsonPointer as an constructor argument. These pointers must be validated to fail early in case they contain illegal characters.

### DIFF
--- a/signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeResponseTest.java
+++ b/signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeResponseTest.java
@@ -12,12 +12,14 @@
  */
 package org.eclipse.ditto.signals.commands.things.modify;
 
+import static org.eclipse.ditto.signals.commands.things.TestConstants.Pointer.EMPTY_JSON_POINTER;
 import static org.eclipse.ditto.signals.commands.things.assertions.ThingCommandAssertions.assertThat;
 import static org.mutabilitydetector.unittesting.AllowedReason.provided;
 import static org.mutabilitydetector.unittesting.MutabilityAssert.assertInstancesOf;
 import static org.mutabilitydetector.unittesting.MutabilityMatchers.areImmutable;
 
 import org.eclipse.ditto.json.JsonFactory;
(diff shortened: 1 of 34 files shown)


signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeResponseTest.java
```diff
@@ -36,6 +39,8 @@
  */
 public final class DeleteAttributeResponseTest {
 
+    private static final JsonPointer INVALID_JSON_POINTER = JsonFactory.newPointer("key1/äöü/foo");
```
DerSchwilk, line 42: Shouldn't TestConstants.INVALID_JSON_POINTER be used instead? (Same value)

signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeTest.java
```diff
@@ -35,6 +38,7 @@
 public final class DeleteAttributeTest {
 
     private static final JsonPointer KNOWN_JSON_POINTER = JsonFactory.newPointer("key1");
+    private static final JsonPointer INVALID_JSON_POINTER = JsonFactory.newPointer("key1/äöü/foo");
```
DerSchwilk, line 41: Shouldn't TestConstants.INVALID_JSON_POINTER be used instead? (Same value)

signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeResponseTest.java
dguggemos: Yes!

signals/commands/things/src/test/java/org/eclipse/ditto/signals/commands/things/modify/DeleteAttributeTest.java
dguggemos: Yes!

PR #767, opened 2020-08-20
Logging and metric improvements for amqp

Yannic92:


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpConsumerActor.java
```diff
@@ -292,16 +295,14 @@ private void handleJmsMessage(final JmsMessage message) {
...
             inboundMonitor.success(externalMessage);
-
+            final Map<String, String> externalMessageHeaders = externalMessage.getHeaders();
             LogUtil.enhanceLogWithCorrelationId(log,
                     externalMessage.findHeader(DittoHeaderDefinition.CORRELATION_ID.getKey()));
-            if (log.isDebugEnabled()) {
-                log.debug("Received message from AMQP 1.0 ({}): {}", externalMessage.getHeaders(),
-                        externalMessage.getTextPayload().orElse("binary"));
-            }
+            log.info("Received message from AMQP 1.0 ({}): {}", externalMessageHeaders,
```
thjaeckle, line 301: @Yannic92 was this done on purpose? logging the complete received textPayload of each incoming AMQP 1.0 message on `info` level?

Yannic92: Mh, switching the log level from debug to info was on purpose, but I think message headers would be sufficient. The message id or something to correlate the incoming message log with the acknowledged message log should be part of the headers.
I think we can have an additional debug log and move the text payload there. What do you think?

thjaeckle: Sounds good, yes 👍 
I'll make it so.

thjaeckle (approved):
Good to merge

PR #769, opened 2020-08-24
Use rql-parser and thingssearch-parser with Java 8

ffendt:


thjaeckle (approved):
LGTM

I will cherry-pick the commit to the release-1.1 branch - that way we could provide a Ditto (and Ditto Client) 1.1.4 if needed

PR #781, opened 2020-08-28
Allow configuring max pool size

ffendt:
The pool is used for mapping inbound and outbound messages in the connectivity service. It is configured
per connection in the attribute `processorPoolSize`.


thjaeckle (approved):
LGTM

Issue #770, opened 2020-08-24
Connectivity service does not consume message after reconnect to AMQP (0.9.1) 


qthuy:
Ditto connectivity service does not consume message after reconnect to AMQP (0.9.1).

Ditto version: 1.1.3
AMQP: rabbitmq:3.8.5
Environment: kubernetes 1.14

Step to reproduce:
1. Configure ditto connectivity service to consume from AMQP.
2. Confirm that ditto can consume and process ditto message from AMQP.
3. Delete rabbitmq cluster PODs (simulate rolling upgrade on rabbitmq).
4. Confirm that rabbitmq cluster re-created ok.
5. Confirm that ditto connectivity service re-connect to AMQP ok.
6. Send ditto message on AMQP.
7. Expect ditto connectivity service to consume and process the message, but it does not until I restart the ditto connectivity POD.

I collected log for ditto connectivity service below.
[ditto-connectivity.log](https://github.com/eclipse/ditto/files/5120169/ditto-connectivity.log)

thjaeckle:
Thank you for your bug report.

From the logs, do I understand correctly that:
* you have 2 AMQP 0.9.1 connections 
   * `rabbit-ditto-connection-source-3` with both sources (consuming from RabbitMQ) and targets (publishing to RabbitMQ)?
   * `rabbit-ditto-connection-target-3` only with a target (publishing to RabbitMQ)?
* the connectiond are successfully re-established
   `2020-08-24 22:24:15,222 INFO  ... Established RMQ connection: amqp://ditto-user@172.20.204.247:5672/`
* however neither consuming messages (via the sources) or publishing messages (via the targets) works?

qthuy:
I setup rabbit-ditto-connection-source-3 as source only (send message to ditto) and rabbit-ditto-connection-target-3 as target only (get notifications from ditto).

Yes, connectivity service stop consuming message from rabbit-ditto-connection-source-3 with this bug. I usually only get notification after I send a modify message to ditto using rabbit-ditto-connection-source-3, so not sure if ditto still continue to send to rabbit-ditto-connection-target-3 or not.

rabbit-ditto-connection-source-3:

{
    "targetActorSelection": "/system/sharding/connection",
    "headers": {
        "aggregate": false
    },
    "piggybackCommand": {
        "type": "connectivity.commands:modifyConnection",
        "connection": {
            "id": "rabbit-ditto-connection-source-3",
            "connectionType": "amqp-091",
            "connectionStatus": "open",
[...]

PR #782, opened 2020-08-31
#770 recreate consumer actors after rabbitmq client reconnect

dguggemos:
The consumers of the RabbitMQ connection were not recreated with the new channel after the client reconnected.

Fixes: #770


thjaeckle (approved):
Cool, LGTM 👍

PR #784, opened 2020-09-02
encode binary payload when for HTTP push publisher with base64

thjaeckle:
* fixed that text payload was handled as "null"

@yufei-cai could you please review?
FMPOV the `else if` I adjusted was wrong - has a negation where it should not have one. This should have caused that e.g. "text/plain" payload was handled as `null` instead of hitting the `else` statement.
For `binary()` ContentTypes it makes sense to Base64 encode the payload and put it into a JsonString - I added this.


yufei-cai (approved):
LGTM

PR #786, opened 2020-09-04
Fix that blank `ca` in connection leads to an empty truststore

ffendt:
Instead of falling back to the JDK truststore, all certificates were rejected. With this PR, a blank `ca` field is handled as if it wouldn't exist.


joosdavid (approved):
Looks good to me

PR #787, opened 2020-09-04
add logging of failures during connectivity message publishing on "info" level

thjaeckle:
* same/similar as/to the consumer connection log entries
* adding correlationId + connectionId to MDC


yufei-cai (approved):
LGTM

PR #789, opened 2020-09-04
prevent "correlation-id" DittoHeader to be ever empty

thjaeckle:
by validating its value with a new NonEmptyValueValidator
* that ensures that even when the DittoHeaders are being built from a JsonObject or Map, non-emptiness is ensured


yufei-cai (approved):
LGTM

Issue #790, opened 2020-09-04
Import package configuration line breaks OSGi compatibility of Ditto JSON


kaloyanrradev:
Please check this line in the maven-bundle-plugin config of the Ditto JSON client bundle:

https://github.com/eclipse/ditto/blob/1ba630b5f07086c9a5127ef9a5b3a28650bc8343/json/pom.xml#L176

It needs to be cleared or extended with a star in the end, otherwise it will hide (not import) the Jackson packages used:

`<Import-Package>com.eclipsesource.json,*</Import-Package>`

Currently the single imported package actually allows the bundle to be resolved in OSGi environments, but it will actually have a ClassNotFound exception once it tries to uses most of its functionality (unless the class is loaded by the caller code).

thjaeckle:
Hi.
Jackson should not be a required dependency for Ditto-json. It's only used in Ditto's backend for doing CBOR serialization.
When not available on the classpath, Ditto-json should still work. 
Do you have a stack trace where the Jackson class is being loaded?

thjaeckle:
Probably we did something wrong here.

ditto-json in fact has imports to `com.fasterxml.jackson.core` and `com.fasterxml.jackson.dataformat.cbor` - however catches a `NoClassDefFoundError` when they can't be loaded during runtime.

See https://github.com/eclipse/ditto/blob/master/json/src/main/java/org/eclipse/ditto/json/CborAvailabilityChecker.java#L38-L48

This `CborAvailabilityChecker` is used in both JsonObject and JsonArray implementations.
We somehow need to dynamically determine and load the CBOR classes - but in OSGi environments that could get tricky. Will have to investigate ...

thjaeckle:
We're working on it.
Goal is to ship "ditto-json" with no dependency to any Jackson library, also not in scope of OSGi "imports".

thjaeckle:
@kaloyanrradev how exactly did you produce this error?
We're wondering as this change is already in ditto-json since Ditto 1.1.0 which was released in April this year.
And we just tried out using the ditto-client 1.1.0 and 1.2.0 (which uses ditto-json as well) in a Felix container which worked.

desislava-marinova:
@thjaeckle, as we used ditto 1.0.0 and we had no such issues, in ditto 1.1.5 the problem is also experienced. 

Is the following stack trace helpful? 
[...]

PR #793, opened 2020-09-08
Issue #790 split up ditto-json and ditto-json-cbor

thjaeckle:
* in order to get ditto-json w/o dependencies to Jackson running in OSGi environments, it is required to do a split and move CBOR related functionalities to another module
* abstraction is done via a java.util.ServiceLoader
* application of the ServiceLoader in OSGi is not ensured or tested as we don't assume that someone wants to use CBOR backed serialization when e.g. using the Java Ditto-Client
* some "breaking" changes to ditto-json can be tolerated as the affected APIs can be treated as "Ditto backend internal" APIs required to apply CBOR serialization

Fixes: #790


dguggemos (approved):
LGTM 👍

PR #794, opened 2020-09-09
Add a payload mapper to convert directly between message command payload and external messages

### LINKED ISSUE #777: Add a payload mapper to convert directly between message command payload and external messages
The payload mapper should behave similarly to the messages HTTP API in that it converts between the payload of message commands and the payload of messages over protocols such as AMQP, MQTT and Kafka. For incoming messages, the mapper may optimize the presentation of the payload for the Ditto protocol JSON format according to content type.

Yannic92:
This PR fixes #777


thjaeckle (approved):
LGTM 👍

PR #796, opened 2020-09-10
Add warnings to health of BackgroundSyncActor

ffendt:
The `BackgroundSyncActor` synchronizes things from the twin persistence to the search index. It's current health is provided at the `/status/health` route.
To be more aware of inconsistent states in the health response, this PR changes the level of two state messages of the actor to `WARN` (instead of `INFO`):
* when the stream between things and the `BackgroundSyncActor` sync actor fails with an error
* when the same thing with the same revision is processed twice


DerSchwilk (approved):
LGTM :+1:

PR #802, opened 2020-09-15
Missing JsonKey Validation for ModifyAttribute and ModifyFeatureProperty

joosdavid:
Due to the missing validation of the JsonValue in the constructor of ModifyAttribute and ModifyFeatureProperty, it was possible to populate the thing with not valid JsonKeys.


jufickel-b:
Why no validate in the model classes?

joosdavid:
In both cases the command is built with a JsonValue.

PR #804, opened 2020-09-16
Move OWASP Security response headers out of code

ffendt:
Move OWASP Security response headers out of the code as they should be handled by a reverse proxy in front of Ditto


jbartelh (approved):
LGTM

PR #807, opened 2020-09-17
Send error responses for implicit thing creation

jokraehe:
This PR fixes that error responses will be sent back to the gateway device.


DerSchwilk (approved):
LGTM

PR #817, opened 2020-09-24
made BaseClientState enum Jsonifiable

thjaeckle:
in order to use as message sent in the cluster w/o the need for using the Java Serializer

Required as prerequisite for using Akka 2.6 without Java Serializer (which is now the default) when connections have configured a `clientCount > 1`.


Yannic92 (approved):

Yannic92:
I'm going to add a unit test for this. Looks good to me. Will merge it after my unit test.

PR #819, opened 2020-09-25
Changes topic path documentation to use entity name instead of id

DerSchwilk:
For the topic path an entity name should be used instead of a namespaced entity id. The documentation in this regard was misleading.

### DIFF
--- a/documentation/src/main/resources/pages/ditto/protocol-specification-topic.md
+++ b/documentation/src/main/resources/pages/ditto/protocol-specification-topic.md
@@ -7,7 +7,7 @@ permalink: protocol-specification-topic.html
 
 The Ditto Protocol defines a **Topic** for each Protocol message having following structure:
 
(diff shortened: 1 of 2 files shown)


documentation/src/main/resources/pages/ditto/protocol-specification-topic.md
```diff
@@ -21,9 +21,9 @@ Examples for valid topic paths are:
 
 The entity's namespace in which the entity is located.
 
-## Entity ID
+## Entity Name
```
thjaeckle, line 24: When changing this heading, the link to it also changes.
Please check that no other page referenced to `protocol-specification-topic.html#entity-id` (adjust link otherwise).

Please also adjust in Line 10 the link + the text of the overall topic.

DerSchwilk: You're right. I'll have a check :+1:

PR #820, opened 2020-09-25
Enriched logging with Exceptions message which are not DittoRuntimeExceptions

joosdavid:
Messages from exceptions which are not DittoRuntimeExceptions were ignored.


stmaute:
LGTM 👍

PR #821, opened 2020-09-25
Add header to optionally allow policy creation without WRITE access

dguggemos:
By default, it is not possible create a policy for which the authorized subject has no `WRITE` permission. This avoids that the caller is locked out (cannot change it after creation) from the created policy. For some scenarios however this might be reasonable, e.g. if the caller creates the policy for someone else and does not want to modify the policy later. Therefore the header `allow-policy-lockout` was added to skip this "policy lockout" check. It can be used when creating a policy (directly or via the `_copyPolicyFrom` field when creating a thing).


stmaute:
LGTM 👍

Issue #823, opened 2020-09-28
Metadata isn't yet persisted correctly for event journal entries


thjaeckle:
Current, when putting metadata, the added metadata is not persisted in the Akka event journal.

It is only applied to the in-memory state of the `thing`. 
When the snapshot mechanism of our persistence stores a snapshot, the metadata contained in the thing is persisted in the snapshot. 

However, if no snapshot is done and e.g. Ditto shuts down, the in-memory metadata information is lost and can't be recovered.

The only correct way to solve that is to add the `metadata` to the persisted event (e.g. similar to `_timestamp` of the event).
By doing that, the ThingPersistenceActor can access this information when replaying events and apply it to the in-memory state.

PR #825, opened 2020-09-29
Fix persisting _metadata for thing journal entries

thjaeckle:
Fixes: #823 

With the fix, the `_metadata` field is added when persisting the `Event` (e.g. the `ThingEvent`) to the Akka event journal.
This lead to some refactorings in all `ThingEvent`s as the `metadata` had to be added to the event model classes.
I deprecated the old `of` factory methods and added a new one taking in the `metadata`.


signals/events/base/src/main/java/org/eclipse/ditto/signals/events/base/Event.java
```diff
@@ -77,6 +78,14 @@ default JsonSchemaVersion getImplementedSchemaVersion() {
      */
     Optional<Instant> getTimestamp();
 
+    /**
+     * Returns the event's metadata.
+     *
+     * @return the metadata.
+     * @since 1.3.0
+     */
+    Optional<Metadata> getMetadata();
```
yufei-cai, line 87: This method fits better in ThingEvent. No other event has metadata.

services/things/persistence/src/main/java/org/eclipse/ditto/services/things/persistence/actors/strategies/events/AbstractThingEventStrategy.java
```diff
@@ -48,20 +47,14 @@ public Thing handle(final T event, @Nullable final Thing thing, final long revis
         if (null != thing) {
             ThingBuilder.FromCopy thingBuilder = thing.toBuilder()
                     .setRevision(revision)
-                    .setModified(event.getTimestamp().orElse(null));
-            thingBuilder = setMetadata(event, thing, thingBuilder);
+                    .setModified(event.getTimestamp().orElse(null))
+                    .setMetadata(event.getMetadata().orElse(null));
```
yufei-cai: Isn't the event metadata relativised to the resource path? Is it correct to replace the thing metadata by the event metadata?

yufei-cai: Nevermind, fixed in subsequent commit
https://github.com/eclipse/ditto/pull/825/commits/f58e8ee5454f7e9d80ee447f2bf968deb9852776

signals/events/base/src/main/java/org/eclipse/ditto/signals/events/base/Event.java
thjaeckle: Yes, however we also added getMetadata() to Entity so that eg also Policy could have metadata.
However policy also only provides empty metadata currently.
So this is analog to Entity

yufei-cai (approved):
LGTM. Just deleted unused class and added and fixed unit tests.

PR #826, opened 2020-09-29
Added metrics about acknowledgements 

thjaeckle:
(whether they were a success and redeliver status)
* enhanced existing metrics about message mapping with connectionType


yufei-cai (approved):
LGTM

PR #828, opened 2020-09-30
Use maxQueryTime config setting

ffendt:
Actually set the max query time in MongoClientWrapper when built from a MongoDbConfig


thjaeckle (approved):
LGTM

Issue #773, opened 2020-08-26
Logging might cause NPE


jufickel-b:
`ThingCommandEnforcement` uses a `DittoDiagnosticLoggingAdapter`.
As all of Ditto's loggers, this logger is not thread-safe. However, `ThingCommandEnforcement` happens to access the logger in parallel. This leads to a `NullPointerException` where in `DittoDiagnosticLoggingAdapter` the correlation ID is made a String.

A first, defensive, approach to circumvent that NPE is to not call `toString` on the correlation ID but passing it down to LogUtil as CharSequence. Furthermore, `ThingCommandEnforcement` should not use its parent's logger but its own (which should be a `DittoLogger`).
Both measures do not remedy the problem of tangled correlation IDs due to parallel calls to logger. To address this it would be necessary to introduce a synchronized wrapper for `DittoLogger` and `DittoDiagnosticLoggingAdapter` which could be used where needed.

thjaeckle:
> introduce a synchronized wrapper

Veto, I think that is performance wise a really bad idea. Better to have some wrongly correlated log-statements then IMO.

jufickel-b:
Veto is veto.

How about a new type like `ThreadSafeDittoLogger` which differs from the already existing `DittoLogger` in a way that it only provides the `withCorrelationId` methods as well as the `discardCorrelationId` method.

An immutable implementation of that type could return a new logger instance for each of those methods which guarantees thread safety.

Furthermore a different type with an obvious name might prevent unnecessary usage.

PR #830, opened 2020-10-01
Bugfix/npe logging command enforcement

jufickel-b:
Improved Ditto's logging framework.

* There are now thread-safe variants of Ditto loggers. They are intended to be used in situations where multiple threads use the same logger instance to ensure that the expected MDC is used.
* Additional to correlation ID, Ditto loggers now allow to set and remove any diagnostic value.

Where it was apparent, the loggers were already switched with their thread-safe variant.
Besides, some refactoring happened in `BasePublisherActor`.

Fixes: #773


thjaeckle (approved):
LGTM 👍

Issue #822, opened 2020-09-28
SSE thing search 


Geonjae-Joo:
I changed nginx-cors.conf  Acess-Control-Allow-Origin wild card to $http_origin in order to perform SSE, and It worked successfully.

but when I try to SSE search/things, the CORS issue pops up again.
anyone knows how to fix it?

![image](https://user-images.githubusercontent.com/62494715/94382947-8e007f80-0179-11eb-86cf-4fc51f112d4c.png)
![image](https://user-images.githubusercontent.com/62494715/94382970-a4a6d680-0179-11eb-87ed-6fabb4ec66d9.png)

thjaeckle:
Can you inspect via browser which HTTP Headers are sent back on the SSE search call?
CORS should work for SSE search as well as for other HTTP calls.

Geonjae-Joo:
@thjaeckle oh, thank you for your reply, i will check on that right after vacation. It's a long holiday week in South Korea :)

Geonjae-Joo:
@thjaeckle 
this is my http header for normal SSE:
![image](https://user-images.githubusercontent.com/62494715/95033345-d29b9600-06f8-11eb-84a0-ec1e7bfa373a.png)

and this is http header for thing search SSE:
![image](https://user-images.githubusercontent.com/62494715/95033341-cdd6e200-06f8-11eb-8abd-d8008d4f8893.png)

I still have no idea what is wrong here. maybe the content-type?

thjaeckle:
I think you found a bug - I can't successfully create a SSE for search, regardless of the CORS headers.
I even wonder why you run into the CORS problem as Ditto should report a HTTP status 501.

I created PR #831 which should fix that ..

thjaeckle:
The `Content-Type` is another potential problem - but as long as the SSE for search is not working at all, I can't repdroduce it.
Probably the `nginx-cors.conf` has to be adjusted that it doesn't statically set the content-type to `text/plain` - at least that's my feeling.

Geonjae-Joo:
Yeap! That's my feeling too.
Thanks for your help :)

PR #831, opened 2020-10-05
fixed that search via SSE was broken

thjaeckle:
* the "proxyActor" was not configured for the ThingsSseRouteBuilder

Fixes: #822


yufei-cai (approved):
LGTM

PR #848, opened 2020-10-21
fixes #844: "location" header was not read when provided by external headers

### LINKED ISSUE #844: Lower-case location header is removed by ditto via websocket response
I'm on ditto 1.3 using a websocket connection to process a ditto live message. The response receives a status of 303 and a location header. If the location header is all-lower-case `location` it is not returned to the client of the original ditto message. Only if the header is `Location` it is returned (and all-lowercase again at the client, at least with curl).

I've verified (via log) that the location header is present until the point right before it's returned via the websocket send operation. I've also verified that this happens from ditto v1.1 to 1.3. For < 1.1 I don't know.

According to the HTTP RFCs 2616 and 7230 HTTP headers should be case-**in**sensitive.

How to reproduce
---

1. Connect via websocket to ditto's live channel
2. Send an arbitrary live message to the websocket listener
3. Send a `location` or `Location` header with the response

thjaeckle:
fixes: #844

### DIFF
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
@@ -230,7 +230,7 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
      *
      * @since 1.1.0
      */
-    WWW_AUTHENTICATE("www-authenticate", String.class, true, true,
+    WWW_AUTHENTICATE("www-authenticate", String.class, false, true,
             HeaderValueValidators.getNoOpValidator()),
 
     /**
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
@@ -230,7 +230,7 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
      *
      * @since 1.1.0
      */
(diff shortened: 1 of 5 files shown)


model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
```diff
@@ -230,7 +230,7 @@
      *
      * @since 1.1.0
      */
-    WWW_AUTHENTICATE("www-authenticate", String.class, false, true,
+    WWW_AUTHENTICATE("www-authenticate", String.class, true, true,
```
Yannic92, line 233: I think it could result in strange behaviours when we read this from external headers. I also don't understand what the benefit could be.

Yannic92 (changes requested):
This doesn't completely fix #844 as this issue also mentioned that we should handle http headers in a case insensitive way. I'm however not sure if we can simply change our behaviour in this case.
Have you already thought about it?

model/base/src/main/java/org/eclipse/ditto/model/base/headers/DittoHeaderDefinition.java
thjaeckle: Yes, you're right .. I'll undo this one.

thjaeckle:
@Yannic92 yes, this PR indeed only fixes the "location" issue - which was wrongly defined in our `DittoHeaderDefinition` to be not readable from "external" headers.
What is still missing is the case insensitivity - I'll have a look on that generally in our `HeaderTranslator`.

Yannic92 (approved):

PR #849, opened 2020-10-21
Make protocol of OAuth providers configurable.

yufei-cai:
Added an OAuth configuration to set the protocol with which to access all OAuth endpoints, to enforce HTTPS for production and to use unencrypted HTTP for development and test.


stmaute (approved):

stmaute:
LGTM 👍

PR #850, opened 2020-10-22
Bugfix/response required for events and responses

Yannic92:
Responses to events and responses are not supported in ditto. Therefore it makes sense to enforce that a published response or event does always have the header value of response-required set to false.


yufei-cai (approved):
LGTM

PR #851, opened 2020-10-23
Declare unique ack labels

### LINKED ISSUE #792: Make custom acknowledgement labels unique for each subscriber
As a follow up of #611, #661 and #757, a uniqueness constraint shall be imposed on acknowledgement labels each subscriber (websocket or connection) may issue. This is to prevent racing in fulfillment of acknowledgement requests and to detect misconfiguration early.

- Require each subscriber to declare a list of acknowledgement labels it may issue.
- Prevent subscribers to send acknowledgements whose labels do not match any in the declared list.
- Add a distributed data to check for global uniqueness at subscription time.

yufei-cai:
Enforce that acknowledgement labels are globally unique and identifying of the subscriber.

Fix #792.


thjaeckle (approved):
LGTM 👍

Yannic92 (approved):

PR #865, opened 2020-10-30
Enhanced model and APIs with "desiredProperties" for Features

### LINKED ISSUE #697: Enhance model and APIs with "desiredProperties" for Features
Actual issue for implementing the model discussed first in #125 and now being part of #696 

The latest proposal is the following JSON structure containing the "desiredProperties" needed for #696:
```json
{
  "thingId": "org.eclipse.ditto:my-thermostat-1",
  "policyId": "org.eclipse.ditto:my-thermostat-1",
  "features": {
    "Thermostat": {
      "definition": [ "org.eclipse.ditto:Thermostat:1.0.0" ],
      "properties": {
        "status": {
          "measured-temperature": 20
        },
        "configuration": {
          "target-temperature": 20.0
        }
      },
      "desiredProperties": {
        "configuration": {
          "target-temperature": 22.0
        }
      }
    }
  }
}
```

Explanation:
* aside to `properties` the new Ditto managed field `desiredProperties` is added
* the same `definition` applies for both `properties` and `desiredProperties`
   * however, inside `desiredProperties` all mandatory properties defined by the definition are optional
* once a `desiredProperties` property is set (e.g. by a backend application or mobile app), the actual device has the task to apply that and reflect the applied `desiredProperties` property in the same one below the `properties` object
* when `desiredProperties` is not existing or is empty, all desired property changes were applied, none are pending

The pros/cons of this approach (other formats were already discussed in #125) are:

`+`:
* same definition is shared for properties and desired
   * (future topic relevant for schema validation): however, for desired the "required" properties defined in the Vorto model probably have all to be optional
* defines fine grained which deltas should be sent downlink
* policy may be written in a way which defines different access rights for properties/desiredProperties
* search can use desiredProperties and/or reported properties (even in combination)
* reduced unneeded payload duplication (besides properties hierarchy structure in JSON)
* no need for a `desired` channel in addition to `twin/live` - that would be "just" another JsonPointer path in the twin

`-`:
* is currently NOT possible by convention before implementing the model enhancement

## Interim format until model is implemented

In order to mitigate the negative point, that this format is not usable in the meantime until the Ditto model was enhanced, the following interim format is suggested:

<details>
  <summary>Click to see the interim format</summary>
  
  ```json
{
  "thingId": "org.eclipse.ditto:my-thermostat-1",
  "policyId": "org.eclipse.ditto:my-thermostat-1",
  "features": {
    "Thermostat": {
      "definition": [ "org.eclipse.ditto:Thermostat:1.0.0" ],
      "properties": {
        "~properties": {
          "status": {
            "measured-temperature": 20
          },
          "configuration": {
            "target-temperature": 20.0
         }
        },
        "~desiredProperties": {
          "configuration": {
            "target-temperature": 22.0
          }
        }
      }
    }
  }
}
  ```
</details>

All API calls would work very similar to the future structure. Partial GET/PUT semantics are similarly supported.

The character prefix `"~"` is used to mark that this is the preliminary convention and this allows a simpler migration towards the target format.

Migration is later possible by:
* rewriting all paths: "properties/~properties" → "properties" and "properties/~desiredProperties" → "desiredProperties"

A script could optionally migrate existing things/policies from the interim format to the actual one once implemented.

joosdavid:
Fixes #697 

Desired properties were added to the Thing Model and APIs.


documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/createthingresponse.md
```diff
@@ -22,7 +22,8 @@
           "y": 2.718,
           "z": 1,
           "unit": "g"
-        }
+        },
+        "desiredProperties": {}
```
thjaeckle, line 26: That looks wrong - in the related `creatething.md`, there was a `desiredProperty` "x" set to "4".
That must also be in the response of the created Thing.

documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/modifyfeature.md
```diff
@@ -11,6 +11,12 @@
       "y": 2.718,
       "z": 1,
       "unit": "g"
+    },
+    "desiredProperties": {
```
thjaeckle, line 15: I don't see the file `modifyfeaturersponsecreated.md` also being changed in this PR - but it should be as the created JSON content is returned when the feature gets created.

thjaeckle:
@joosdavid and @DerSchwilk: please ensure that by splitting the OpenAPI doc in several files that all places which use the API doc still work:
* the Ditto website at https://www.eclipse.org/ditto/http-api-doc.html
* the kubernetes deployment
* the Docker (compose) deployment
* can it be still used to import it into Postman?

documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/modifyfeature.md
joosdavid: The file is created and should be displayed as before.. 
[documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/modifyfeatureresponsecreated.md](url)

model/things/src/main/java/org/eclipse/ditto/model/things/ThingBuilder.java
```diff
@@ -992,7 +1054,7 @@ default FromCopy setFeature(final Feature feature) {
         /**
          * Sets the given Feature to this builder. A previously set Feature with the same ID is replaced.
          *
-         * @param existingFeaturesPredicate a predicate to decide whether the given features are set. The predicate
+         * @param existingFeaturesPredicate a predicate to decide whether the given features exist. The predicate
```
jufickel-b, line 1057: Descriptions of predicates are not yet correct (in other classes, too).
The given predicate determines whether the provided arguments are set to the builder (depending on the Features in this case).
This should be transported by the Javadoc comment.

DerSchwilk: Changed in commit: https://github.com/eclipse/ditto/pull/865/commits/41a3a6d494a67cd4e4a76e7103d147d8d0e35e05
Can you have a check?

documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/createthingresponse.md
thjaeckle: still missing as far as I see

documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/modifyfeature.md
thjaeckle: Yes, you're right

documentation/src/main/resources/pages/ditto/protocol/examples/things/generated/commands/modify/createthingresponse.md
DerSchwilk: Yes you're right... Thought this was already fixed.
https://github.com/eclipse/ditto/pull/865/commits/1af9fc6397b3edb1b5c4de67e2a810a651e17291

PR #868, opened 2020-11-02
Fix 2 sonar issues.

yufei-cai:
1. Duplicate subexpression in `DefaultAmqp10Config#equals(Object)`.

2. Synchronous HTTP call in `DittoPublicKeyProvider`.


thjaeckle (approved):
LGTM 👍

PR #870, opened 2020-11-02
Have AmqpPublisherActor wait for message settlement in stream.

yufei-cai:
This prevents message sending when the number of unsettled messages
grows too large, in addition to when messages are not settled.
The error message is adjusted to reflect the fact.


thjaeckle (approved):
LGTM 👍

PR #872, opened 2020-11-04
Don't send error response for exceptions during implicit thing creation

jokraehe:


thjaeckle (approved):
Looks good to me 👍

PR #873, opened 2020-11-05
fixed exception handling in AbstractEnforcement

thjaeckle:
* simplifies enforcement
* should fix timeouts for gathering metrics because exceptions were handled somewhere else than the default metric closing


Yannic92 (approved):
LGTM.
I took this as an opportunity to unify the handling of asking an actor and handle the response properly.

PR #875, opened 2020-11-06
Allow sending empty "Messages" via the OpenAPI documentation

ffendt:
Add a content-type `none/none` as a means to send empty messages with OpenAPI (as OpenAPI doesn't support requests without any content-type).

The type `none/none` is actually used by akka-http as a fallback if a request without any content-type information reaches an endpoint. By explicitly using the content-type `none/none`, we therefore forward a similar message to/from a Thing (via OpenAPI) as if it were sent without a content-type:
```
{
	"topic": "org.eclipse.ditto:thing",
	"headers": {
		"response-required": false,
		"ditto-originator": "nginx:ditto",
		"correlation-id": "009fc0ec-0282-4fc4-a3d8-84e85fef7514",
		"content-type": "none/none",
		"version": 2,
		"timeout": "10",
		"accept": "*/*",
		"timestamp": "2020-11-06T13:45:01.664126345+01:00"
	},
	"path": "/inbox/messages/anyop",
	"value": ""
}

```


thjaeckle:
Hm, I guess what we really wanted here is that "value" is not present at all in the Ditto Protocol message created from this call.

DerSchwilk (approved):
LGTM :+1:

PR #877, opened 2020-11-09
Fix error mapping in ImplicitThingCreationMessageMapper

dguggemos:


services/connectivity/mapping/src/main/java/org/eclipse/ditto/services/connectivity/mapping/ImplicitThingCreationMessageMapper.java
```diff
@@ -227,9 +227,12 @@ private static String getCopyPolicyFrom(final JsonObject thingJson) {
     @Override
     public List<ExternalMessage> map(final Adaptable adaptable) {
         if (TopicPath.Criterion.ERRORS.equals(adaptable.getTopicPath().getCriterion())) {
-            throw GlobalErrorRegistry.getInstance().parse(adaptable.getPayload().toJson(), adaptable.getDittoHeaders());
+            adaptable.getPayload().getValue().filter(JsonValue::isObject).map(JsonValue::asObject)
```
Yannic92, line 230: Refactoring looks correct to me. I'm however not 100% sure if it's correct to just drop "error" adaptables here in case the value is not present or if present but not a JsonObject.
I think we could at least add a WARN log here, as it is something we would not expect here, but its not a critical issue (no ERROR log required).
What do you think?

Yannic92 (approved):

PR #879, opened 2020-11-11
fix regression about (expected) exceptions thrown in AmqpPublisherActor CompletionStage

thjaeckle:
backing the SourceQueue of AmqpPublisherActor

regression was introduced in #870 by not handling the "exceptionally" part which caused the queue to break


dguggemos (approved):
LGTM :+1:

PR #883, opened 2020-11-17
Send weak acknowledgements for dropped signals

### LINKED ISSUE #852: Prevent endless redelivery for filtered out events requesting acknowledgements
After #611, #661, #757, #792, Ditto has enough information to detect misconfigured publisher-subscriber pairs, for example:
- The subscriber that declared an ack label requested by the publisher is not authorized to receive the published signal, or
- The subscriber that declared an ack label requested by the publisher discards the published signal due to namespace or RQL filtering.

Resending the signal will not help in these situations. Ditto should rather emit a "weak acknowledgement" that does not trigger redelivery.

yufei-cai:
This patch adds "weak acknowledgements" issued when the requested subscriber is not authorized for a signal or when the requested subscriber drops the signal intentionally via filtering according to RQL, namespace, signal type or payload mapping.

Fixes #852


thjaeckle (approved):
Approved 👍

PR #888, opened 2020-11-19
Reduce amount of / consolidate config files

Yannic92:
Before this PR we supported the following different "hosting environments":
- docker
- cloud
- filebased
- dev
Which expected config in different files. For example: The docker hosting environment expected the config to be present in `servicename-docker.conf`

This lead to confusion and duplication in configuration files. This PR now only supports the following "hosting environments":

- docker (just uses the default servicename.conf)
- filebased (uses a configurable file as config)
- dev (uses the servicename-dev.conf)

So no `-cloud` and no `-docker` files are required anymore.

VCAP and docker secret support is now available in all kind of config files not only in the `-cloud` config file.


services/gateway/starter/src/main/resources/gateway.conf
```diff
@@ -264,21 +265,22 @@ ditto {
...
 
-      // how long to wait for messages from other cluster members
+      # how long to wait for messages from other cluster members
       ask-timeout = 5s
+      ask-timeout = ${?STATISTICS_UPDATE_INTERVAL}
 
-      // how often to update hot entities count for the public
-      update-interval = 30s
+      # how often to update hot entities count for the public
+      update-interval = 15s
```
jbartelh, line 273: Is this a breaking change in ditto?

services/gateway/starter/src/main/resources/gateway.conf
```diff
@@ -264,21 +265,22 @@ ditto {
...
       update-interval = ${?STATISTICS_UPDATE_INTERVAL}
 
-      // minimum pause between computations of statistics details
-      details-expire-after = 5s
+      # minimum pause between computations of statistics details
+      details-expire-after = 3s
```
jbartelh, line 277: Is this a breaking change in ditto?

services/gateway/starter/src/main/resources/gateway.conf
```diff
@@ -372,7 +379,7 @@ akka {
         #
         # `relaxed`: all visible 7-Bit ASCII chars are allowed
         #
-        uri-parsing-mode = relaxed
+        uri-parsing-mode = strict
```
jbartelh, line 382: Is this a breaking change in ditto?

jbartelh (approved):
Not sure if the three mentioned config value changes are categorized as 'breaking changes' or not, but we should at least mention them in the upcoming release notes so that ditto user can adapt if needed. What do you say @thjaeckle ?

The rest LGTM

services/gateway/starter/src/main/resources/gateway.conf
thjaeckle: no

services/gateway/starter/src/main/resources/gateway.conf
thjaeckle: no

services/gateway/starter/src/main/resources/gateway.conf
thjaeckle: hmmh, I think we once configured `strict` for a very good reason. (that I can't think of any more)
why was this changed to `relaxed` again?

Yannic92: I think you're right. Let's revert this change and set it to relaxed again.

thjaeckle (approved):

PR #891, opened 2020-11-23
increase jms client version to 0.54.0

Yannic92:
There were a few bugfixes in qpid jms client and this PR aims to take advantage of these fixes.


dguggemos (approved):
LGTM 👍

PR #892, opened 2020-11-24
Adapt connectivity config handling and connection restrictions

VadimGue:
Add a more fine granular way for restrict connections and a possibility for connectivity config changes without service restart


json/src/main/java/org/eclipse/ditto/json/JsonFactory.java
```diff
@@ -284,6 +284,17 @@ public static JsonObject newObject(final Iterable<JsonField> jsonFields) {
         return newObjectBuilder(jsonFields).build();
     }
 
+    /**
+     * Merge two JSON objects into one JSON object.
+     *
+     * @param jsonObject1 the json object to merge, overrides conflicting fields.
+     * @param jsonObject2 the json object to merge.
+     * @return returns a new object merged the given {code jsonObject1} and {code jsonObject2}.
```
thjaeckle, line 292: This is missing the `@since 1.5.0`

thjaeckle (changes requested):
Please add the missing mentioned `@since` in the public javadoc in `JsonFactory`

json/src/main/java/org/eclipse/ditto/json/JsonFactory.java
stmaute: I have add the missing annotation.

thjaeckle (approved):

PR #893, opened 2020-11-24
Improve the container build script

ctron:
This change makes the build script more flexible:

* Allow to run from outside the script's directory
* Allow to override most of the parameters
* Allow to change the container registry
* Allow to override the version

### DIFF
--- a/services/build-images.sh
+++ b/services/build-images.sh
@@ -16,6 +16,9 @@
 # specified services (SERVICES).                                               #
 ################################################################################
 
+# Directory of the script
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
 # Array of the services to build Docker images for.
 # The pattern is MODULE_NAME:IMAGE_NAME as both can differ from each other.
 SERVICES=(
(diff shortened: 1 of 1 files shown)


services/build-images.sh
```diff
@@ -45,18 +50,18 @@ print_used_proxies() {
...
-    $DOCKER_BIN build --pull -f $DOCKERFILE \
+    $DOCKER_BIN build --pull -f $SCRIPTDIR/$DOCKERFILE \
       --build-arg HTTP_PROXY="$HTTP_PROXY_LOCAL" \
       --build-arg HTTPS_PROXY="$HTTPS_PROXY_LOCAL" \
       --build-arg TARGET_DIR="$module_name_base"/starter/target \
       --build-arg SERVICE_STARTER="$module_name"-starter \
       --build-arg SERVICE_VERSION=$SERVICE_VERSION \
-      -t "$image_tag":$SERVICE_VERSION \
+      -t "$image_tag":$IMAGE_VERSION \
       $DOCKER_BUILD_CONTEXT
```
thjaeckle, line 65: You removed the `DOCKER_BUILD_CONTEXT` variable - can this reference `$DOCKER_BUILD_CONTEXT` then also be removed?

ctron: Indeed. I replaced it with `$SCRIPTDIR` now.

thjaeckle (approved):
Awesome, I didn't know yet about this syntax:
```
: "${HTTP_PROXY_LOCAL:=$HTTP_PROXY}"
```

Very useful, thx 👍

PR #894, opened 2020-11-24
Speed up the container build

ctron:
Assuming the locally built artifacts changes more
frequent than the package dependencies, it makes more sense
to add the local artifacts as a last step to the container.

This will more often use cached build steps, and thus
speed up the container build process and, as image layers
can be re-use, requires less bytes to be transmitted when
pushing.


thjaeckle (approved):
LGTM 👍

Issue #889, opened 2020-11-20
Support receiving cloud events using HTTP endpoint


ctron:
I think it makes sense for Ditto to support receiving Ditto Protocol JSON messages also via [Cloud Events](https://cloudevents.io/).

While cloud events provide bindings for Kafka, MQTT, … they also have an [HTTP endpoint binding](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md), which can easily be used in the combination with [Knative](https://knative.dev/). So it would be possible, to easily map incoming messages from any [Knative eventing source](https://knative.dev/docs/eventing/), to Ditto, acting as a Knative eventing sink.

The payload of these events would be expected to be "Ditto Protocol payload (JSON)".

I would be volunteering to work on this.

thjaeckle:
Cool 👍 
As discussed in Gitter.IM we would appreciate such a contribution.
Implementation hints:
* HTTP handling in Ditto is done with [Akka HTTP](https://doc.akka.io/docs/akka-http/current/index.html)
* the Ditto service responsible for terminating HTTP requests is the [gateway service](https://www.eclipse.org/ditto/architecture-services-gateway.html), see also the [architecture](https://www.eclipse.org/ditto/architecture-overview.html)
* [`RootRoute.buildApiSubRoutes(RequestContext, DittoHeaders)`](https://github.com/eclipse/ditto/blob/master/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/RootRoute.java#L264) would be the place to add a new top-level route below `/api/2/`

Regarding naming and verb of the new endpoint I would suggest:
* using `POST` as verb
* regarding the "noun":
** I am not yet sure .. `/events` for example would not fit IMO as Ditto Protocol JSON messages are not always events, but also commands
** `/protocolMessages` could be an option, however I would like to prevent using camel case or a dash in the endpoint
** maybe `/signals` as all Ditto Protocol JSON messages represent Signals: https://www.eclipse.org/ditto/basic-signals.html

thjaeckle:
I would close this as the basic functionality is available via #895

PR #895, opened 2020-11-24
[#889]: Implement cloud events HTTP endpoint

ctron:
This change adds support for receiving cloud events through the HTTP
binding.

The event's payload must be in the Ditto Protocol JSON.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -56,6 +56,7 @@
         <amqp-client.version>5.7.3</amqp-client.version>
         <reactive-streams.version>1.0.3</reactive-streams.version>
         <netty.version>4.1.50.Final</netty.version>
+        <cloundevents.version>2.0.0.RC1</cloundevents.version>
 
         <slf4j.version>1.7.28</slf4j.version>
         <logback.version>1.2.3</logback.version>
@@ -324,6 +325,12 @@
                 <groupId>io.jsonwebtoken</groupId>
                 <artifactId>jjwt-impl</artifactId>
                 <version>${jjwt.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>io.cloudevents</groupId>
+                <artifactId>cloudevents-http-basic</artifactId>
+                <version>${cloundevents.version}</version>
             </dependency>
 
             <!-- ### Indirect "runtime" dependencies we want to pin to a common version -->
--- a/documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
+++ b/documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
@@ -1,0 +1,105 @@
+---
+title: Cloud Events HTTP protocol binding
+keywords: binding, protocol, http, cloudevents
+tags: [binding, protocol, http]
+permalink: httpapi-protocol-bindings-cloudevents.html
+---
+
+Implements the [HTTP Protocol Binding for CloudEvents - Version 1.0](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md).
+
+Unless mentioned otherwise, the endpoint following the Cloud Events specification for the HTTP binding in version 1.0.
+
+## Cloud Events features
+
+The Cloud Events endpoint provides an alternative to the other connectivity APIs to stream data into your instance.
+
+## Cloud Events endpoint
+
+The Cloud Events endpoint is accessible at these URLs (depending on which API version to use):
+
+```
+http://localhost:8080/api/1/cloudevents
+http://localhost:8080/api/2/cloudevents
+```
+
+### Authentication
+
+A user who connects to the Cloud Events endpoint can be authenticated by using
+
+* HTTP BASIC Authentication by providing a username and the password of a user managed within nginx or
+* a JSON Web Token (JWT) issued by an OpenID connect provider.
+
+See [Authenticate](basic-auth.html) for more details.
+
+## Cloud Events protocol format
+
+The source must be a Cloud Event, encoded in the format for the HTTP binding. It may be encoded in *binary content mode*
+or in *structural content mode*.
+
+The *data content type* of the event must be `application/json`.
+
+The *data schema* must start with `ditto:`, for example `ditto:some-schema`.
+
+The events *payload* must in the Ditto Protocol JSON format as defined in the
+[Protocol specification](protocol-specification.html).
+
+## Publishing events to the endpoint
+
+Publishing events to the endpoint can be done by directly sending HTTP requests, conforming to the Cloud Events
+HTTP binding specification. Or by using other technologies that have adopted Cloud Events.
+
+### Knative eventing
+
+The endpoint can directly be configured as a [Knative eventing](https://knative.dev/docs/eventing/) destination.
+
+In the following example, a Knative eventing flow is configured to normalize the payload with a Vorto converter
+and send the result to Ditto's cloud events endpoint:
+
+~~~yaml
+apiVersion: flows.knative.dev/v1
+kind: Sequence
+metadata:
+ name: digital-twin
+spec:
+  channelTemplate:
+    apiVersion: messaging.knative.dev/v1alpha1
+    kind: KafkaChannel
+    spec:
+      numPartitions: 1
+      replicationFactor: 1
+  steps:
+  - ref:
+      # Convert incoming payload to the Ditto format
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: vorto-converter
+      namespace: digital-twin
+  reply:
+    # Deliver to Ditto Cloud Events endpoint
+    uri: http://ditto:ditto@ditto-nginx.digital-twin.svc.cluster.local:8080/api/2/cloudevents
+~~~
+
+This sequence itself can again be the target of another operation.
+
+### Direct invocation
+
+Of course, it is also possible to directly access the Cloud Events endpoint through HTTP:
+
+An example HTTP request could look like this:
+
+~~~
+POST /api/2/cloudevents HTTP/1.1
+ce-specversion: 1.0
+ce-type: my.ditto.event
+ce-time: 2020-11-24T14:35:00Z
+ce-id: f7b197fe-2e59-11eb-a8f4-d45d6455d2cc
+ce-source: /my/source
+ce-dataschema: ditto:some-schema
+Content-Type: application/json; charset=utf-8
+
+{
+    ... Ditto Protocol JSON ...
+}
+~~~
+
+For more information, see [HTTP Protocol Binding for CloudEvents - Version 1.0](https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md).
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventNotParsableException.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventNotParsableException.java
@@ -1,0 +1,128 @@
+ /*
+  * Copyright (c) 2020 Contributors to the Eclipse Foundation
+  *
+  * See the NOTICE file(s) distributed with this work for additional
+  * information regarding copyright ownership.
+  *
+  * This program and the accompanying materials are made available under the
+  * terms of the Eclipse Public License 2.0 which is available at
+  * http://www.eclipse.org/legal/epl-2.0
+  *
+  * SPDX-License-Identifier: EPL-2.0
+  */
+ package org.eclipse.ditto.model.base.exceptions;
+
+ import java.net.URI;
+ import java.text.MessageFormat;
+
+ import javax.annotation.Nullable;
+ import javax.annotation.concurrent.Immutable;
+ import javax.annotation.concurrent.NotThreadSafe;
+
+ import org.eclipse.ditto.json.JsonObject;
+ import org.eclipse.ditto.model.base.common.HttpStatusCode;
+ import org.eclipse.ditto.model.base.headers.DittoHeaders;
+ import org.eclipse.ditto.model.base.json.JsonParsableException;
+
+ /**
+  * Thrown if a request with an unparsable cloud event is made.
+  *
+  * @since 1.5.0
+  */
+ @Immutable
+ @JsonParsableException(errorCode = CloudEventNotParsableException.ERROR_CODE)
+ public final class CloudEventNotParsableException extends DittoRuntimeException {
+
+     /**
+      * Error code of this exception.
+      */
+     public static final String ERROR_CODE = "cloudevents.unparsable";
+
+     private static final String DEFAULT_MESSAGE = "The event could not be parsed.";
+     private static final String MESSAGE_PATTERN = "The system was unable to parse the event: {0}";
+     private static final String DESCRIPTION = "Ensure that the event is being transmitted according to the Cloud Events HTTP binding specification v1.0.";
+     private static final URI DEFAULT_URI = URI.create("https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md");
+
+     private static final HttpStatusCode STATUS_CODE = HttpStatusCode.BAD_REQUEST;
+
+     /**
+      * Constructs a new {@code CloudEventNotParsableException} object.
+      *
+      * @param dittoHeaders the headers with which this Exception should be reported back to the user.
+      * @param message the detail message for later retrieval with {@link #getMessage()}.
+      * @param description a description with further information about the exception.
+      * @param cause the cause of the exception for later retrieval with {@link #getCause()}.
+      * @param href a link to a resource which provides further information about the exception.
+      * @throws NullPointerException if {@code dittoHeaders} is {@code null}.
+      */
+     private CloudEventNotParsableException(final DittoHeaders dittoHeaders,
+                                            @Nullable final String message,
+                                            @Nullable final String description,
+                                            @Nullable final Throwable cause,
+                                            @Nullable final URI href) {
+         super(ERROR_CODE, STATUS_CODE, dittoHeaders, message, description, cause, href == null ? DEFAULT_URI : href);
+     }
+
+     /**
+      * A mutable builder for a {@code CloudEventNotParsableException} where the message contains detailed information
+      * about the parse error.
+      *
+      * @param details the unsupported data schema used in the call.
+      * @return the new CloudEventNotParsableException.
+      */
+     public static DittoRuntimeExceptionBuilder<CloudEventNotParsableException> withDetailedInformationBuilder(
+             final String details) {
+
+         final String msgPattern = MessageFormat.format(MESSAGE_PATTERN, details);
+         return new Builder().message(msgPattern).description(DESCRIPTION);
+     }
+
+     /**
+      * Constructs a new {@code CloudEventNotParsableException} object with the exception message extracted from the
+      * given JSON object.
+      *
+      * @param jsonObject the JSON to read the {@link JsonFields#MESSAGE} field from.
+      * @param dittoHeaders the headers of the command which resulted in this exception.
+      * @return the new CloudEventNotParsableException.
+      * @throws NullPointerException if any argument is {@code null}.
+      * @throws org.eclipse.ditto.json.JsonMissingFieldException if this JsonObject did not contain an error message.
+      * @throws org.eclipse.ditto.json.JsonParseException if the passed in {@code jsonObject} was not in the expected
+      * format.
+      */
+     public static CloudEventNotParsableException fromJson(final JsonObject jsonObject, final DittoHeaders dittoHeaders) {
+         return DittoRuntimeException.fromJson(jsonObject, dittoHeaders, new Builder());
+     }
+
+     @Override
+    public DittoRuntimeException setDittoHeaders(final DittoHeaders dittoHeaders) {
+         return new Builder()
+                 .message(getMessage())
+                 .description(getDescription().orElse(null))
+                 .cause(getCause())
+                 .href(getHref().orElse(null))
+                 .dittoHeaders(dittoHeaders)
+                 .build();
+     }
+
+     /**
+      * A mutable builder with a fluent API for a {@link CloudEventNotParsableException}.
+      */
+     @NotThreadSafe
+     public static final class Builder extends DittoRuntimeExceptionBuilder<CloudEventNotParsableException> {
+
+         private Builder() {
+             message(DEFAULT_MESSAGE);
+         }
+
+         @Override
+         protected CloudEventNotParsableException doBuild(final DittoHeaders dittoHeaders,
+                                                          @Nullable final String message,
+                                                          @Nullable final String description,
+                                                          @Nullable final Throwable cause,
+                                                          @Nullable final URI href) {
+
+             return new CloudEventNotParsableException(dittoHeaders, message, description, cause, href);
+         }
+     }
+
+ }
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventNotParsableException.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventNotParsableException.java
@@ -25,7 +25,7 @@
  import org.eclipse.ditto.model.base.json.JsonParsableException;
 
  /**
-  * Thrown if a request with an unparsable cloud event is made.
+  * Thrown if a request with an unparsable CloudEvent is made.
   *
   * @since 1.5.0
   */
@@ -36,12 +36,14 @@
      /**
       * Error code of this exception.
       */
-     public static final String ERROR_CODE = "cloudevents.unparsable";
+     public static final String ERROR_CODE = "cloudevent.unparsable";
 
      private static final String DEFAULT_MESSAGE = "The event could not be parsed.";
      private static final String MESSAGE_PATTERN = "The system was unable to parse the event: {0}";
-     private static final String DESCRIPTION = "Ensure that the event is being transmitted according to the Cloud Events HTTP binding specification v1.0.";
-     private static final URI DEFAULT_URI = URI.create("https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md");
+     private static final String DESCRIPTION =
+             "Ensure that the event is being transmitted according to the Cloud Events HTTP binding specification v1.0.";
+     private static final URI DEFAULT_URI =
+             URI.create("https://github.com/cloudevents/spec/blob/v1.0/http-protocol-binding.md");
 
      private static final HttpStatusCode STATUS_CODE = HttpStatusCode.BAD_REQUEST;
 
@@ -56,10 +58,10 @@
       * @throws NullPointerException if {@code dittoHeaders} is {@code null}.
       */
      private CloudEventNotParsableException(final DittoHeaders dittoHeaders,
-                                            @Nullable final String message,
-                                            @Nullable final String description,
-                                            @Nullable final Throwable cause,
-                                            @Nullable final URI href) {
+             @Nullable final String message,
+             @Nullable final String description,
+             @Nullable final Throwable cause,
+             @Nullable final URI href) {
          super(ERROR_CODE, STATUS_CODE, dittoHeaders, message, description, cause, href == null ? DEFAULT_URI : href);
      }
 
@@ -67,7 +69,7 @@
       * A mutable builder for a {@code CloudEventNotParsableException} where the message contains detailed information
       * about the parse error.
       *
-      * @param details the unsupported data schema used in the call.
+      * @param details the details about why the CloudEvent could not be parsed.
       * @return the new CloudEventNotParsableException.
       */
      public static DittoRuntimeExceptionBuilder<CloudEventNotParsableException> withDetailedInformationBuilder(
@@ -89,12 +91,13 @@
       * @throws org.eclipse.ditto.json.JsonParseException if the passed in {@code jsonObject} was not in the expected
       * format.
       */
-     public static CloudEventNotParsableException fromJson(final JsonObject jsonObject, final DittoHeaders dittoHeaders) {
+     public static CloudEventNotParsableException fromJson(final JsonObject jsonObject,
+             final DittoHeaders dittoHeaders) {
          return DittoRuntimeException.fromJson(jsonObject, dittoHeaders, new Builder());
      }
 
      @Override
-    public DittoRuntimeException setDittoHeaders(final DittoHeaders dittoHeaders) {
+     public DittoRuntimeException setDittoHeaders(final DittoHeaders dittoHeaders) {
          return new Builder()
                  .message(getMessage())
                  .description(getDescription().orElse(null))
@@ -116,10 +119,10 @@
 
          @Override
          protected CloudEventNotParsableException doBuild(final DittoHeaders dittoHeaders,
-                                                          @Nullable final String message,
-                                                          @Nullable final String description,
-                                                          @Nullable final Throwable cause,
-                                                          @Nullable final URI href) {
+                 @Nullable final String message,
+                 @Nullable final String description,
+                 @Nullable final Throwable cause,
+                 @Nullable final URI href) {
 
              return new CloudEventNotParsableException(dittoHeaders, message, description, cause, href);
          }
--- a/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventUnsupportedDataSchemaException.java
+++ b/model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventUnsupportedDataSchemaException.java
@@ -1,0 +1,128 @@
+ /*
+  * Copyright (c) 2020 Contributors to the Eclipse Foundation
+  *
+  * See the NOTICE file(s) distributed with this work for additional
+  * information regarding copyright ownership.
+  *
+  * This program and the accompanying materials are made available under the
+  * terms of the Eclipse Public License 2.0 which is available at
+  * http://www.eclipse.org/legal/epl-2.0
+  *
+  * SPDX-License-Identifier: EPL-2.0
+  */
+ package org.eclipse.ditto.model.base.exceptions;
+
+ import java.net.URI;
+ import java.text.MessageFormat;
+
+ import javax.annotation.Nullable;
+ import javax.annotation.concurrent.Immutable;
+ import javax.annotation.concurrent.NotThreadSafe;
+
+ import org.eclipse.ditto.json.JsonObject;
+ import org.eclipse.ditto.model.base.common.HttpStatusCode;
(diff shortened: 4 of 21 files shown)


services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+
+    /**
+     * Convert the request and payload to a cloud event.
+     *
+     * @param ctx The request context.
+     * @param payload The binary payload, this may contain the cloud event payload, or the fully encoded cloud
+     *         event structure.
+     * @return The cloud event
+     */
+    private @Nullable
```
thjaeckle, line 162: please remove the `@Nullable` - this may not be `@Nullable` as `null` is not allowed as element in Akka Streams where this method is invoked in.
I also don't see at the `MessageReader.toEvent()` javadoc that `null` could be returned.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+            if (!sawContentType.get()) {
+                // we didn't see the content type in the header, so extract it from akka's request
+                acceptor.accept(DittoHeaderDefinition.CONTENT_TYPE.getKey(), ctx.getRequest().entity().getContentType().mediaType().toString());
+            }
+        }, payload.toArray());
+
+        return reader.toEvent();
+    }
+
+    private CloudEvent validateCloudEvent(@Nullable final CloudEvent cloudEvent, final RequestContext ctx, final DittoHeaders dittoHeaders) {
```
thjaeckle, line 196: I don't think that the `cloudEvent` can be `@Nullable` here, can it?

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+
+        if (cloudEvent == null || cloudEvent.getData() == null) {
+            throw MissingPayloadException
+                    .withDetailedInformationBuilder()
+                    .dittoHeaders(dittoHeaders)
+                    .build();
+        }
+
+        LOGGER.debug("Cloud event: {}", cloudEvent);
```
thjaeckle, line 205: Please make use of `LOGGER.withCorrelationId(dittoHeaders).trace(...)` in order to log the correlation-id of the DittoHeaders to the MDC.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/MissingPayloadException.java
```diff
@@ -0,0 +1,122 @@
...
+ import org.eclipse.ditto.model.base.exceptions.DittoRuntimeExceptionBuilder;
+ import org.eclipse.ditto.model.base.headers.DittoHeaders;
+ import org.eclipse.ditto.model.base.json.JsonParsableException;
+
+ /**
+  * Thrown if a request with missing payload is being made.
+  */
+ @Immutable
+ @JsonParsableException(errorCode = MissingPayloadException.ERROR_CODE)
+ public final class MissingPayloadException extends DittoRuntimeException {
```
thjaeckle, line 33: Please move the exception to package `org.eclipse.ditto.model.base.exceptions` where also the `UnsupportedMediaTypeException` is.
The reason why unit tests fail is because of added exceptions in this module.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/UnsupportedDataSchemaException.java
```diff
@@ -0,0 +1,129 @@
...
+ import org.eclipse.ditto.model.base.exceptions.DittoRuntimeExceptionBuilder;
+ import org.eclipse.ditto.model.base.headers.DittoHeaders;
+ import org.eclipse.ditto.model.base.json.JsonParsableException;
+
+ /**
+  * Thrown if a request with an unsupported data schema is made.
+  */
+ @Immutable
+ @JsonParsableException(errorCode = UnsupportedDataSchemaException.ERROR_CODE)
+ public final class UnsupportedDataSchemaException extends DittoRuntimeException {
```
thjaeckle, line 35: Please move the exception to package `org.eclipse.ditto.model.base.exceptions` where also the `UnsupportedMediaTypeException` is.
The reason why unit tests fail is because of added exceptions in this module.

Please also rename to `CloudEventUnsupportedDataSchemaException` as this is a "cloud events" specific error.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/UnsupportedDataSchemaException.java
```diff
@@ -0,0 +1,129 @@
...
+
+     /**
+      * Error code of this exception.
+      */
+     public static final String ERROR_CODE = "dataschema.unsupported";
```
thjaeckle, line 40:
```suggestion
     public static final String ERROR_CODE = "cloudevents.dataschema.unsupported";
```

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+        if (data == null) {
+            return Optional.empty();
+        }
+        final byte[] payload = data.toBytes();
+        if (payload == null || payload.length == 0) {
+            return Optional.empty();
+        }
+
+        JsonObject jsonObject = JsonObject.of(payload);
+        LOGGER.debug("JSON: {}", jsonObject);
```
thjaeckle, line 223: Please make use of `LOGGER.withCorrelationId(dittoHeaders).debug(...`

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+    private Optional<Signal<?>> jsonToDittoSignal(@Nullable final CloudEventData data, final DittoHeaders dittoHeaders) {
+        if (data == null) {
+            return Optional.empty();
+        }
+        final byte[] payload = data.toBytes();
+        if (payload == null || payload.length == 0) {
+            return Optional.empty();
+        }
+
+        JsonObject jsonObject = JsonObject.of(payload);
```
thjaeckle, line 222: This jsonObject may be final.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+        LOGGER.debug("JSON: {}", jsonObject);
+
+        final JsonifiableAdaptable jsonifiableAdaptable = ProtocolFactory.jsonifiableAdaptableFromJson(jsonObject);
+        // use empty header translator to pass along the ditto-auth-context information
+        final DittoProtocolAdapter dittoProtocolAdapter = DittoProtocolAdapter.of(HeaderTranslator.empty());
```
thjaeckle, line 227: Please instantiate the `DittoProtocolAdapter` only once as static field in this class.

thjaeckle (changes requested):
Thank you very much for a great contribution 👍 

I had a first look, found some things and mentioned them inline.

Other stuff which I could not add as inline-comments:
* Please adjust the documentation `documentation/src/main/resources/_data/sidebars/ditto_sidebar.yaml` and add the new page about the cloud events binding
   * I suggest below the "title: WebSocket protocol binding"
* please adjust the "Copyright holders" in `legal/NOTICE.md`
* a unit test would be great, but I suggest to add this in a follow-up PR to not reach the 1.000 LoC ;)
* please adjust the OpenAPI documentation with the new added endpoint: https://github.com/eclipse/ditto/tree/master/documentation/src/main/resources/openapi (cloud also be done in a follow-up PR)

A few "open" questions to me:
* Does the Eclipse Foundation release process allow to use "milestone" dependencies when releasing a non-milestone release? Did not find anything about that in the handbook
   * according to ClearlyDefined the cloudevents depencencies have a good enough score, so CQs are not required
* Regarding the "cloud events" fields:
   * Do we need to specify which values `ce-type` and `ce-data-schema` can take?

documentation/src/main/resources/pages/ditto/connectivity-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,92 @@
...
+
+~~~
+POST /api/2/cloudevents HTTP/1.1
+ce-specversion: 1.0
+ce-type: my.ditto.event
+ce-time: 2020-11-24T14:35:00Z
+ce-id: f7b197fe-2e59-11eb-a8f4-d45d6455d2cc
+ce-source: /my/source
+ce-data-schema: ditto:some-schema
+Content-Type: application/json; charset=utf-8
```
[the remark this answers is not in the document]
ctron: I am not so sure about that … as the content is JSON. And the schema of the JSON is "ditto". I guess it could be OK to simply accept both.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,281 @@
...
+
+        // create a reader for the message
+        final MessageReader reader = HttpMessageFactory.createReader(acceptor -> {
+
+            // NOTE: this acceptor may be run multiple times by the message reader
+
+            // record if we saw the content type header
+            final AtomicBoolean sawContentType = new AtomicBoolean();
+            // consume the HTTP request headers
+            ctx.getRequest().getHeaders().forEach(header -> {
```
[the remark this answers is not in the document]
ctron: I don't think that will work. The `acceptor` requires **all** HTTP headers. As the content type may be removed from the original set, that ensures that the content type is passed to the reader.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
ctron: It should not, but there is nothing that ensures that. But I can remove it.

ctron:
> A few "open" questions to me:
> 
>    * Does the Eclipse Foundation release process allow to use "milestone" dependencies when releasing a non-milestone release? Did not find anything about that in the handbook

It is a released version on Maven central. So that is stable enough. I am not aware of any limitations from the Eclipse Foundation side. Their focus is on availability and license compatibility.

>       
>    * according to ClearlyDefined the cloudevents depencencies have a good enough score, so CQs are not required

Correct.

> 
>   * Regarding the "cloud events" fields:
>       
>     * Do we need to specify which values `ce-type` and `ce-data-schema` can take?

Hm good question. The `ca-data-schema` already has the expectation in code of starting with `ditto:`. And that is part of the documentation. For the type, I am not sure and open for suggestions.

ctron:
I amended the PR with changes to your feedback. I will now take a look at the documentation part of that.

ctron:
I re-iterated over the documentation. Maybe have a second look.

documentation/src/main/resources/pages/ditto/connectivity-protocol-bindings-cloudevents.md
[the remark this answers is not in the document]
thjaeckle: Accepting both sounds good, yes - the implementation however would also need to accept the Ditto Protocol content-type -  if I looked correctly it does not yet.

thjaeckle:
The file `connectivity-protocol-bindings-cloudevents.md` must now be removed and with it all links to `connectivity-protocol-bindings-cloudevents.html`

model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventUnsupportedDataSchemaException.java
```diff
@@ -0,0 +1,126 @@
...
+ import javax.annotation.concurrent.Immutable;
+ import javax.annotation.concurrent.NotThreadSafe;
+
+ import org.eclipse.ditto.json.JsonObject;
+ import org.eclipse.ditto.model.base.common.HttpStatusCode;
+ import org.eclipse.ditto.model.base.headers.DittoHeaders;
+ import org.eclipse.ditto.model.base.json.JsonParsableException;
+
+ /**
+  * Thrown if a request with an unsupported data schema is made.
```
thjaeckle, line 28: Missing `@since 1.5.0`, sorry .. I missed this in the first iteration

model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/MissingPayloadException.java
```diff
@@ -0,0 +1,120 @@
...
+ import javax.annotation.concurrent.Immutable;
+ import javax.annotation.concurrent.NotThreadSafe;
+
+ import org.eclipse.ditto.json.JsonObject;
+ import org.eclipse.ditto.model.base.common.HttpStatusCode;
+ import org.eclipse.ditto.model.base.headers.DittoHeaders;
+ import org.eclipse.ditto.model.base.json.JsonParsableException;
+
+ /**
+  * Thrown if a request with missing payload is being made.
```
thjaeckle, line 27: Missing `@since 1.5.0`, sorry .. I missed this in the first iteration

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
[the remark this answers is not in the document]
thjaeckle: Ah, now I get it .. nevermind and thx for clarification :)

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,105 @@
...
+* a JSON Web Token (JWT) issued by an OpenID connect provider.
+
+See [Authenticate](basic-auth.html) for more details.
+
+## Cloud Events protocol format
+
+The source must be a Cloud Event, encoded in the format for the HTTP binding. It may be encoded in *binary content mode*
+or in *structural content mode*.
+
+The *data type* of the event must be `application/json`.
```
thjaeckle, line 39: If I understood the CloudEvent spec correctly, this is the `datacontenttype`, not the `data type`.
Which - if omitted in "binary content mode" - is populated from the `Content-Type` HTTP header.

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,105 @@
...
+
+The *data schema* must start with `ditto:`, for example `ditto:org.eclipse.vorta:TestDevice:1.0.0`.
```
thjaeckle, line 41: If I understood the CloudEvent spec correctly, this is the `dataschema`, not the `data schema`.
I would like to omit the reference to Vorto here as this information (the Vorto model) can currently not be used to do anything meaningful.

Thinking about it: do we even need to specify this? It is optional in the spec and we don't do anything meaningful with that ..
So adding it to a cloud-event brings no value.

thjaeckle:
> Hm good question. The `ca-data-schema` already has the expectation in code of starting with `ditto:`. And that is part of the documentation. For the type, I am not sure and open for suggestions.

Like mentioned inline:
* I would suggest to remove specifying and validating the `ca-dataschema` again as this is optional in the CloudEvents spec
* I read again: the `type` is defined by the event producer, so we can't specify it

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
ctron: Changed to `data content type` … I would leave it as is, since the HTTP binding spec of Cloud Events defined how this is encoded. Either from the Content-Type, from a Header or the wrapping JSON structure.

ctron:
> The file `connectivity-protocol-bindings-cloudevents.md` must now be removed and with it all links to `connectivity-protocol-bindings-cloudevents.html`

I thought I had done that. Then I had some merge issues. Will re-check.

ctron:
> Like mentioned inline:
> 
>     * I would suggest to remove specifying and validating the `ca-dataschema` again as this is optional in the CloudEvents spec

My idea was to do some kind of validation to ensure that the data is actually suitable for Ditto. The field is optional, but still Ditto could require it.

But I don't have a strong opinion here. If you think it is better to remove this, then I will do that.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,296 @@
...
+        return Optional.of(PROTOCOL_ADAPTER
+                .fromAdaptable(jsonifiableAdaptable)
+                .setDittoHeaders(dittoHeaders));
+    }
+
+    private void ensureDataContentType(@Nullable final String dataContentType,
+                                       final RequestContext ctx,
+                                       final DittoHeaders dittoHeaders) {
+
+        if (dataContentType == null || !mediaTypeJsonWithFallbacks.contains(dataContentType)) {
```
thjaeckle, line 252: I think that this "contains" check is problematic as the `dataContentType` could also contain a charset.
The javadoc of `getDataContenType()` says that it MUST follow RFC2046 which can contain the optional charset.

ctron: Yes, but I think Ditto can define what it expects in there. Same a the standard HTTP `Content-Type` header … so requiring `application/json` from a Ditto side is fine I think.

thjaeckle:
> My idea was to do some kind of validation to ensure that the data is actually suitable for Ditto. The field is optional, but still Ditto could require it.
> 
> But I don't have a strong opinion here. If you think it is better to remove this, then I will do that.

Ok, I would suggest to at least make it optional (so that it may be null), but if it is provided, it must start with `ditto:` (like it is currently implemented.

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,105 @@
...
+  steps:
+  - ref:
+      # Convert incoming payload to the Ditto format
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: vorto-converter
+      namespace: drogue-iot
+  reply:
+    # Deliver to Ditto Cloud Events endpoint
+    uri: http://ditto:ditto@ditto-nginx.drogue-iot.svc.cluster.local:8080/api/2/cloudevents
```
thjaeckle, line 79: Are you sure that you want `....drogue-iot...` in here? ;)

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
thjaeckle: Hm, some tools may always include the charset even if not explicitly specified, so I think that this would also be true for backend applications emitting cloud-events.

I could also work on this part (also supporting the Ditto Protocol content-type) if you like.

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
ctron: True.

thjaeckle:
Cloudevents Java SDK 2.0.0.RC1 was just released: https://github.com/cloudevents/sdk-java/releases/tag/2.0.0.RC1

ctron:
> Cloudevents Java SDK 2.0.0.RC1 was just released: https://github.com/cloudevents/sdk-java/releases/tag/2.0.0.RC1
> Bleeding edge ;)

Cool, let me trigger the clearlydefine harvester.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -0,0 +1,296 @@
...
+                acceptor.accept(header.name(), header.value());
+            });
+
+            if (!sawContentType.get()) {
+                // we didn't see the content type in the header, so extract it from akka's request
+                acceptor.accept(DittoHeaderDefinition.CONTENT_TYPE.getKey(), ctx.getRequest().entity().getContentType().mediaType().toString());
+            }
+        }, payload.toArray());
+
+        return reader.toEvent();
```
thjaeckle, line 206: I just tried out invoking the Route manually.
This definitely needs to catch `CloudEventRWException` and `IllegalStateException`. `IllegalStateException` is for example thrown if invoked without "cloudevent" headers.
That currently causes an 500 "internal server" error because of the unhandled exception.

I suggest catching the exceptions here and throwing a new DittoRuntimeException instead, e.g. `CloudEventNotParsableException` with status code 400 "bad request" giving a hint that a HTTP cloudevent was expected and e.g. pointing to the spec in the "href" of the exception.

thjaeckle:
I just manually invoked the `/cloudevents` resource.
When doing this successfully e.g. with a "create thing" Ditto Protocol message, the returned response is the same as invoking e.g. the `POST /api/2/things` resource: when creating a thing with 201 "created" outcome, the created thing JSON is returned.

I ask me whether this is an expected outcome.
I guess that the expected outcome would be also to get a response in Ditto Protocol JSON as I also sent this in.
Any thoughts about that, @ctron ?

Yannic92:
> 
> 
> I just manually invoked the `/cloudevents` resource.
> When doing this successfully e.g. with a "create thing" Ditto Protocol message, the returned response is the same as invoking e.g. the `POST /api/2/things` resource: when creating a thing with 201 "created" outcome, the created thing JSON is returned.
> 
> I ask me whether this is an expected outcome.
> I guess that the expected outcome would be also to get a response in Ditto Protocol JSON as I also sent this in.
> Any thoughts about that, @ctron ?

> I just manually invoked the `/cloudevents` resource.
> When doing this successfully e.g. with a "create thing" Ditto Protocol message, the returned response is the same as invoking e.g. the `POST /api/2/things` resource: when creating a thing with 201 "created" outcome, the created thing JSON is returned.

Thought about this as well. Content negotiation would be awesome for this endpoint.
I'm however not sure if this is required for this first pull request. It's already pretty powerful. Thank you @ctron!
I think Ditto Protocol would be a reasonable default, because it's the same content-type as used for the command payload.
Maybe we could support content negotiation later?

thjaeckle:
The [spec for HTTP](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#3-http-message-mapping) says on that:
> The event binding is identical for both HTTP request and response messages.

So for HTTP responses there are some open issues:
* the content-mode must be preserved (when invocation did happen with "structured content mode", the response must be also in "structured content mode")
* the passed-in cloudevent headers have to be preserved and returned in the HTTP response
* IMO the response payload must also be in Ditto Protocol JSON

ctron:
> So for HTTP responses there are some open issues:

I would propose to handle responses in a new PR. There is no requirement to respond with a cloud event, Ditto can be seen as a "sink", only accepting messages without a full reply. "Full", in the sense of Cloud event response, it still would respond with an HTTP response that it accepted or rejected the HTTP request.

ctron:
> > I just manually invoked the `/cloudevents` resource.
> > When doing this successfully e.g. with a "create thing" Ditto Protocol message, the returned response is the same as invoking e.g. the `POST /api/2/things` resource: when creating a thing with 201 "created" outcome, the created thing JSON is returned.
> > I ask me whether this is an expected outcome.
> > I guess that the expected outcome would be also to get a response in Ditto Protocol JSON as I also sent this in.
> > Any thoughts about that, @ctron ?
> 
> > I just manually invoked the `/cloudevents` resource.
> > When doing this successfully e.g. with a "create thing" Ditto Protocol message, the returned response is the same as invoking e.g. the `POST /api/2/things` resource: when creating a thing with 201 "created" outcome, the created thing JSON is returned.
> 
> Thought about this as well. Content negotiation would be awesome for this endpoint.
> I'm however not sure if this is required for this first pull request. It's already pretty powerful. Thank you @ctron!
> I think Ditto Protocol would be a reasonable default, because it's the same content-type as used for the command payload.
> Maybe we could support content negotiation later?

I am not sure what you have in mind with "content negotiation". From the HTTP binding spec's perspective there are two (or three) modes available. Which get selected through the `Content-Type` header. "Structured" and "binary". That is why I was referring the docs to "data content type" and not to "Content-Type", because the latter is an implementation detail of the binary encoding.

I think it is ok for Ditto to enforce the "data content type", but that would not be done through the HTTP Content-Type header, but through the parsed Cloud Event "data content type" field.

The current implementation already support both binary and structured mode. Only "batch" is not supported, but that should require some extra understanding of both communication parties anyway:

> The batched mode MUST NOT be used unless solicited, and the gesture SHOULD allow the receiver to choose the maximum size of a batch.

Yannic92:
> I am not sure what you have in mind with "content negotiation". From the HTTP binding spec's perspective there are two (or three) modes available. Which get selected through the `Content-Type` header. "Structured" and "binary". That is why I was referring the docs to "data content type" and not to "Content-Type", because the latter is an implementation detail of the binary encoding.

What I had in mind was to define the content-type returned in the response by setting the `Accept`-header in the request, like describe [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation).

Based on this, ditto could either return the `application/json` which it returns for the other HTTP endpoints or it could return `application/vnd.eclipse.ditto+json`.

But as already mentioned: I think its not required to be part of this pull request. I however think that the default content-type should be `application/vnd.eclipse.ditto+json`, because that's what's provided in the request. 
It seems to be intuitive for me to get the same structure of payload in the response that is provided in the request.

ctron:
> > I am not sure what you have in mind with "content negotiation". From the HTTP binding spec's perspective there are two (or three) modes available. Which get selected through the `Content-Type` header. "Structured" and "binary". That is why I was referring the docs to "data content type" and not to "Content-Type", because the latter is an implementation detail of the binary encoding.
> 
> What I had in mind was to define the content-type returned in the response by setting the `Accept`-header in the request, like describe [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation).
> 
> Based on this, ditto could either return the `application/json` which it returns for the other HTTP endpoints or it could return `application/vnd.eclipse.ditto+json`.
> 
> But as already mentioned: I think its not required to be part of this pull request. I however think that the default content-type should be `application/vnd.eclipse.ditto+json`, because that's what's provided in the request.
> It seems to be intuitive for me to get the same structure of payload in the response that is provided in the request.

I am not sure what the response would be? But maybe we discuss this in a new issue, which could focus in the response, or there is one. Right now, it simply accepts incoming cloud events. And the response is, that it accepted/rejected the event.

I am not sure how the `Accept` header maps to the cloud events spec though. Or if that is concept which should be used in this context. But we can discuss this in a new issue/PR.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
ctron: I updated the PR, checking and handling the error. Please take a look, I am not sure I was following "best practice" of Ditto error handling.

model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/CloudEventNotParsableException.java
```diff
@@ -0,0 +1,129 @@
...
+                                            @Nullable final URI href) {
+         super(ERROR_CODE, STATUS_CODE, dittoHeaders, message, description, cause, href == null ? DEFAULT_URI : href);
+     }
+
+     /**
+      * A mutable builder for a {@code CloudEventNotParsableException} where the message contains detailed information
+      * about the actual used data schema and the description information about data schemas are supported for the
+      * requested resource.
+      *
+      * @param details the unsupported data schema used in the call.
```
thjaeckle, line 71: Please fix the javadoc copy/paste, this mentions "data schema".

model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/MissingPayloadException.java
```diff
@@ -0,0 +1,122 @@
...
+  *
+  * @since 1.5.0
+  */
+ @Immutable
+ @JsonParsableException(errorCode = MissingPayloadException.ERROR_CODE)
+ public final class MissingPayloadException extends DittoRuntimeException {
```
thjaeckle, line 33: I suggest to also rename this exception with the "CloudEvent" prefix to: `CloudEventMissingPayloadException`

model/base/src/main/java/org/eclipse/ditto/model/base/exceptions/MissingPayloadException.java
```diff
@@ -0,0 +1,122 @@
...
+
+     /**
+      * Error code of this exception.
+      */
+     public static final String ERROR_CODE = "payload.missing";
```
thjaeckle, line 38: Please use suffix `"cloudevents."` here as well.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
thjaeckle: Cool, looks good 👍

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,105 @@
...
+An example HTTP request could look like this:
+
+~~~
+POST /api/2/cloudevents HTTP/1.1
+ce-specversion: 1.0
+ce-type: my.ditto.event
+ce-time: 2020-11-24T14:35:00Z
+ce-id: f7b197fe-2e59-11eb-a8f4-d45d6455d2cc
+ce-source: /my/source
+ce-data-schema: ditto:some-schema
```
thjaeckle, line 97: This header must be `ce-dataschema`. I just verified - when setting to `ce-data-schema`, the cloudevent is not processed, as the dataSchema is not available in the CloudEvent object.

documentation/src/main/resources/pages/ditto/httpapi-protocol-bindings-cloudevents.md
```diff
@@ -0,0 +1,105 @@
...
+See [Authenticate](basic-auth.html) for more details.
+
+## Cloud Events protocol format
+
+The source must be a Cloud Event, encoded in the format for the HTTP binding. It may be encoded in *binary content mode*
+or in *structural content mode*.
+
+The *data content type* of the event must be `application/json`.
+
+The *data schema* must start with `ditto:`, for example `ditto:org.eclipse.vorto:TestDevice:1.0.0`.
```
thjaeckle, line 41: Please adjust this example to e.g. `"ditto:some-schema"`.

bom/pom.xml
```diff
@@ -56,6 +56,7 @@
         <amqp-client.version>5.7.3</amqp-client.version>
         <reactive-streams.version>1.0.3</reactive-streams.version>
         <netty.version>4.1.50.Final</netty.version>
+        <cloundevents.version>2.0.0-milestone4</cloundevents.version>
```
thjaeckle, line 59: 2.0.0.RC1 may now be used instead

thjaeckle:
I added some (hopefully last) comments to the latest adjustments.
FMPOV we would be good to merge then and follow up (or create issues) with the discussed other stuff:
* unit tests
* OpenAPI documentation
* responding with Ditto Protocol JSON
   * potentially with what @Yannic92 suggested (negotiation): interpreting the `Accept` header
* accepting Ditto Protocol JSON specific content-type

What I think would be useful until the response-handling is clarified:
* statically set "response-required" to false for all incoming cloudevents
* however, use the requested acknowledgement "twin-persisted" in order to block responding to the HTTP request until successfully persisted in Ditto
   * the benefit would be that
      * in a success case, the HTTP response will be 202 (and empty payload)
      * in an error case, the HTTP response will reflect the error which occured (e.g. missing permissions, bad formatted JSON, etc.)
   * WDYT, @Yannic92 - should that work?

ctron:
I amended the PR based on your feedback. Please double check the addition of the header, I am not 100% this is what you had in mind.

thjaeckle:
I pushed one commit to your branch and hope you don't mind.
That was easier for me instead of starting commenting all over again.

I changed the header-thing I mentioned (response-required=false and requested-acks=twin-persisted) and ensure that it works that way - your variant was the right direction, however the invoked `.setDittoHeaders(DittoHeaders)` did overwrite the headers again.
For that reason, we also don't need the `DittoProtocolAdapter` to be initialized in a special way.

I also had to fix a bug in `AbstractShardedPersistenceActor` which dropped errors when no response was required, but "twin-persisted" acknowledgement was requested.

thjaeckle (approved):

ctron:
> I pushed one commit to your branch and hope you don't mind.

Not at all! :+1:

thjaeckle:
Added mandatory follow-ups in #903

PR #896, opened 2020-11-24
Support integration testing against mongo 4.2

Yannic92:
This PR starts and stops a MongoDB docker container instead of using the embedded MongoDB approach of flapdoodle.

The main reason for this change is, that we want to support MongoDB  4.2 and therefore also want to test against this MongoDB  version. Unfortunately flapdoodle does still have [an open issue](https://www.google.com/search?client=firefox-b-e&q=flapdoodle+mongo+4.2) for this topic.

So with this PR we can now verify that our integration tests are working with a MongoDB  of version 4.2.


services/utils/test/src/test/java/org/eclipse/ditto/services/utils/test/mongo/MongoContainerFactory.java
```diff
@@ -14,6 +14,7 @@
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
```
Yannic92, line 17: Why adding an import for "Map" when using "List"?

jbartelh (approved):
Naming the container would be nice for better identification, but found no simple way without losing the capability of starting multiple container or introducing a naming-generator. The docker implementation could be more general but out of scope, can be added later on when needed.
Therefore -> LGTM

Issue #903, opened 2020-12-01
CloudEvents route follow up tasks


thjaeckle:
After the first part of adding a `/cloudevents` route done in #889 and PR #895 there are some follow ups:
- [ ] in addition to `application/json` content-type, also accept the Ditto Protocol content type `application/vnd.eclipse.ditto+json`
- [ ] also accept content-types suffixed with a charset
- [ ] make the `ce-dataschema` optional - but if it is provided, it must start with `ditto:`
- [ ] provide unit tests for HTTP route
- [ ] enhance OpenAPI docs adding the new `/cloudevents` route and providing the Ditto Protocol JSON schema inline in the OpenAPI doc

ctron:
I created PR https://github.com/eclipse/ditto/pull/905 … following up in content-type suffix , optional data schema and additional content type.

thjaeckle:
@ctron I added OpenAPI documetation in #911 (wanted to have that in the release and the docs for the release).

PR #905, opened 2020-12-04
[#903]: Make schema optional, allow alternate content type

ctron:
This change allows to make the schema optional and also allows to
configure the data content type.

### DIFF
--- a/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
+++ b/services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
@@ -36,6 +36,7 @@ import org.eclipse.ditto.protocoladapter.JsonifiableAdaptable;
 import org.eclipse.ditto.protocoladapter.ProtocolFactory;
 import org.eclipse.ditto.services.gateway.endpoints.actors.AbstractHttpRequestActor;
 import org.eclipse.ditto.services.gateway.endpoints.routes.AbstractRoute;
+import org.eclipse.ditto.services.gateway.util.config.endpoints.CloudEventsConfig;
 import org.eclipse.ditto.services.gateway.util.config.endpoints.CommandConfig;
 import org.eclipse.ditto.services.gateway.util.config.endpoints.HttpConfig;
 import org.eclipse.ditto.services.utils.akka.logging.DittoLoggerFactory;
@@ -46,6 +47,8 @@ import org.eclipse.ditto.signals.commands.base.CommandNotSupportedException;
 import akka.actor.ActorRef;
 import akka.actor.ActorSystem;
 import akka.actor.Status;
+import akka.http.javadsl.model.ContentType;
+import akka.http.javadsl.model.ContentTypes;
 import akka.http.javadsl.model.HttpRequest;
 import akka.http.javadsl.model.HttpResponse;
 import akka.http.javadsl.server.RequestContext;
@@ -73,6 +76,8 @@ public final class CloudEventsRoute extends AbstractRoute {
     private static final DittoProtocolAdapter PROTOCOL_ADAPTER = DittoProtocolAdapter.newInstance();
     private static final String DATA_SCHEMA_SCHEME = "ditto";
 
+    private final CloudEventsConfig cloudEventsConfig;
+
     /**
      * Constructs the cloud events route builder.
      *
@@ -81,6 +86,7 @@ public final class CloudEventsRoute extends AbstractRoute {
      * @param httpConfig the configuration settings of the Gateway service's HTTP endpoint.
      * @param commandConfig the configuration settings for incoming commands (via HTTP requests) in the gateway.
      * @param headerTranslator translates headers from external sources or to external sources.
+     * @param cloudEventsConfig the configuration settings for cloud events.
      * @throws NullPointerException if any argument is {@code null}.
      */
     public CloudEventsRoute(
@@ -88,9 +94,11 @@ public final class CloudEventsRoute extends AbstractRoute {
             final ActorSystem actorSystem,
             final HttpConfig httpConfig,
             final CommandConfig commandConfig,
-            final HeaderTranslator headerTranslator
+            final HeaderTranslator headerTranslator,
+            final CloudEventsConfig cloudEventsConfig
     ) {
         super(proxyActor, actorSystem, httpConfig, commandConfig, headerTranslator);
+        this.cloudEventsConfig = cloudEventsConfig;
     }
 
 
@@ -251,7 +259,7 @@ public final class CloudEventsRoute extends AbstractRoute {
         final JsonifiableAdaptable jsonifiableAdaptable = ProtocolFactory.jsonifiableAdaptableFromJson(jsonObject);
         final Signal<?> signal = PROTOCOL_ADAPTER.fromAdaptable(jsonifiableAdaptable);
         final Signal<?> signalWithAdjustedHeaders = signal.setDittoHeaders(
-                        signal.getDittoHeaders().toBuilder().putHeaders(adjustedHeaders).build());
+                signal.getDittoHeaders().toBuilder().putHeaders(adjustedHeaders).build());
         return Optional.of(signalWithAdjustedHeaders);
     }
 
@@ -259,7 +267,7 @@ public final class CloudEventsRoute extends AbstractRoute {
             final RequestContext ctx,
             final DittoHeaders dittoHeaders) {
 
-        if (dataContentType == null || !mediaTypeJsonWithFallbacks.contains(dataContentType)) {
+        if (!isCorrectDataType(dataContentType)) {
             if (LOGGER.isInfoEnabled()) {
                 LOGGER.withCorrelationId(dittoHeaders)
                         .info("CloudEvent request rejected: unsupported data-content-type: <{}>  request: <{}>",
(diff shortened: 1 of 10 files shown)


ctron:
Hm, looks like you have a CI check in place that doesn't allow to properly specify the copyright year range.

thjaeckle:
> Hm, looks like you have a CI check in place that doesn't allow to properly specify the copyright year range.

Yes, copyright header for us is only the creation year as this is sufficient according to the Eclipse Foundation Handbook.

ctron:
> > Hm, looks like you have a CI check in place that doesn't allow to properly specify the copyright year range.
> 
> Yes, copyright header for us is only the creation year as this is sufficient according to the Eclipse Foundation Handbook.

Works for me :+1:

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -251,28 +260,52 @@ private CloudEvent validateCloudEvent(final CloudEvent cloudEvent, final Request
...
+     * @param dataContentType The content type to check.
+     * @return {@code true} if the content type is acceptable, {@code false} otherwise.
+     */
+    private boolean isCorrectDataType(@Nullable final String dataContentType) {
+        if (dataContentType == null) {
+            // no content type
+            return false;
+        }
+
+        final var result = MediaType.parse(dataContentType);
```
thjaeckle, line 299: Please use `akka.http.javadsl.model.ContentTypes.parse(String)` instead (whenever possible, use Akka's "javadsl" packages which contain the Java APIs - using the Scala APIs also work but often looks "strange").
The returned `ContentType` has a `mediaType()` accessor.
Please also don't use `var` keyword here - we did not yet adopt this ;)

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/endpoints/DefaultCloudEventsConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.ditto.services.utils.config.ConfigWithFallback;
+import org.eclipse.ditto.services.utils.config.ScopedConfig;
+
+import com.typesafe.config.Config;
+
+/**
+ * This class is the default implementation of the cloud events endpoint config.
+ */
+@Immutable
+public class DefaultCloudEventsConfig implements CloudEventsConfig {
```
thjaeckle, line 29: Please make this class `final`

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/endpoints/DefaultCloudEventsConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        DefaultCloudEventsConfig that = (DefaultCloudEventsConfig) o;
```
thjaeckle, line 72: please add `final`

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/endpoints/DefaultCloudEventsConfig.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(emptySchemaAllowed, dataTypes);
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + " [" +
+                "emptySchemaAllowed=" + emptySchemaAllowed +
+                "dataTypes=" + dataTypes +
```
thjaeckle, line 85:
```suggestion
                ", dataTypes=" + dataTypes +
```

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/endpoints/CloudEventsConfig.java
```diff
@@ -0,0 +1,76 @@
...
+
+        /**
+         * Flag if an empty data schema is allowed.
+         */
+        EMPTY_SCHEMA_ALLOWED("empty-schema-allowed", true),
+
+        /**
+         * Set of allowed data types
+         */
+        DATA_TYPES("data-types", Set.of("application/json", "application/vnd.eclipse.ditto+json"));
```
thjaeckle, line 55: Please use `org.eclipse.ditto.model.base.common.DittoConstants.DITTO_PROTOCOL_CONTENT_TYPE` for the Ditto Protocol content-type.
`application/json` could also be referenced e.g. via `MediaTypes.APPLICATION_JSON.toString()`.

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/endpoints/DefaultCloudEventsConfig.java
```diff
@@ -0,0 +1,89 @@
...
+        return emptySchemaAllowed;
+    }
+
+    @Override
+    public Set<String> getDataTypes() {
+        return dataTypes;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
thjaeckle, line 65: please add `final` for `o`

thjaeckle (changes requested):
Thx @ctron, already looks very promising :)
Added a few requests..

One more thing: 
Could you please enhance the `services/gateway/starter/src/main/resources/gateway.conf` with the `cloud-events` section (e.g. below `public-health { ... }`)  and the same defaults as in the code, so:
```
cloud-events {
  empty-schema-allowed = true
  data-types = [
    "application/json"
    "application/vnd.eclipse.ditto+json"
  ]
}
```
That way Ditto's configuration files act as documentation of what can be configured overall.

ctron:
I just updated the PR.

services/gateway/endpoints/src/main/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsRoute.java
```diff
@@ -20,6 +20,7 @@
 
 import javax.annotation.Nullable;
 
+import akka.http.javadsl.model.ContentTypes;
```
thjaeckle, line 23: Double import can be removed

services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public final class CloudEventsTest extends EndpointTestBase {
+
+    @Test
+    public void testDefaults() {
```
thjaeckle, line 27: IMO this test does not make much sense as it just ensures what already is added as defaults in the `CloudEventsConfigValue`. If you enhance the default config, you must also enhance the test accordingly.

services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.eclipse.ditto.services.gateway.endpoints.EndpointTestBase;
+import org.eclipse.ditto.services.gateway.util.config.endpoints.CloudEventsConfig;
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public final class CloudEventsTest extends EndpointTestBase {
```
thjaeckle, line 24: What we normally add to our endpoint tests can be seen e.g. in: https://github.com/eclipse/ditto/blob/master/services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/things/ThingsRouteTest.java

* build the route
* wrap it in Akka HTTP testkit's `testRoute` (as seen in `setUp()`)
* run `POST ` requests against the endpoint providing `ce-*` headers or omitting them in order to test for the exception which may be returned

thjaeckle (changes requested):
The `CloudEventsTests` would need meaningful tests, but I'm not sure if you want to address all mentioned follow-up tasks of #903 ;)
I would rather delete the current `CloudEventsTests` class and the one not so useful test to be honest.

services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsTest.java
ctron: Well it was for me, testing of the constants that I used actually provide the expected content. They do. But I thought I would leave the test in there, to ensure that is true in the future.

But I can also remove them if you prefer that.

services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsTest.java
ctron: I can do that, but would tackle the problem in another PR.

ctron:
> The `CloudEventsTests` would need meaningful tests, but I'm not sure if you want to address all mentioned follow-up tasks of #903 ;)
> I would rather delete the current `CloudEventsTests` class and the one not so useful test to be honest.

I wanted to keep the scope of this PR as is, if possible. And work on the other items in a new PR. Just to keep PRs small.

I tried to explain the idea behind the current state of the unit test above. But as mentioned, I am open to dropping it altogether for this PR.

services/gateway/endpoints/src/test/java/org/eclipse/ditto/services/gateway/endpoints/routes/cloudevents/CloudEventsTest.java
thjaeckle: Yes, I would prefer to delete that again 👍

thjaeckle (approved):

thjaeckle:
Thx for the contribution, really appreciated 👍

PR #907, opened 2020-12-07
Set failure-without-redelivery AMQP disposition to REJECTED 

yufei-cai:
Instead of MODIFIED[undeliverableHere].

Reason: Incoming messages with client errors should not be delivered
to any other clients, including other client actors of the same
connection.

### DIFF
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpConsumerActor.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpConsumerActor.java
@@ -14,7 +14,7 @@ package org.eclipse.ditto.services.connectivity.messaging.amqp;
 
 import static org.apache.qpid.jms.message.JmsMessageSupport.ACCEPTED;
 import static org.apache.qpid.jms.message.JmsMessageSupport.MODIFIED_FAILED;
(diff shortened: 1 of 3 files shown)


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpConsumerActor.java
```diff
@@ -366,7 +366,7 @@ private void acknowledge(final JmsMessage message, final boolean isSuccess, fina
                 ackType = ACCEPTED;
                 ackTypeName = "accepted";
             } else {
-                ackType = redeliver ? MODIFIED_FAILED : MODIFIED_FAILED_UNDELIVERABLE;
+                ackType = redeliver ? MODIFIED_FAILED : REJECTED;
                 ackTypeName = redeliver ? "modified[delivery-failed]" : "modified[delivery-failed,undeliverable-here]";
```
thjaeckle, line 370: @yufei-cai please also adjust the `ackTypeName` accordingly to "rejected" - this is used only for logging but it would be confusing if that would differ.

thjaeckle (changes requested):
Please adjust the `ackTypeName` as well

yufei-cai:
Fixed. Also adjusted documentation.

thjaeckle (approved):
LGTM 👍

PR #911, opened 2020-12-10
[#903] provide OpenAPI documentation for /cloudevents on API 2

thjaeckle:
Did not add explicitly for API 1 as this one already is deprecated and not so well integrated in Ditto's OpenAPI building based on several source files.


yufei-cai (approved):
LGTM

PR #919, opened 2020-12-16
fix implicit thing creation with "If-None-Match: *" header

thjaeckle:
caused that a referenced policy (via _copyPolicyFrom) kept the precondition headers
* this fails the "retrieve policy" as the condition most likely will not match


dguggemos (approved):
LGTM 👍

PR #920, opened 2020-12-16
Fix pubsub incremental update and connectivity backoff

yufei-cai:
- Fix that re-subscription by the last subscriber did not trigger distributed data update.
- Reset backoff strategy whenever a client actor is connected.
- Prevent AmqpClientActor from entering CONNECTED state before message producers and consumers are created.


thjaeckle (approved):
LGTM

PR #938, opened 2021-01-13
Add possibility for search query validation extensions

DerSchwilk:
This PR adds the possibility to use custom search query filter validations via Akka extension points.
The validator to be used can be adjusted via the `query-criteria-validator.implementation` config option.


DerSchwilk:
Already reviewed by @ffendt and me

PR #939, opened 2021-01-13
Replaced enum HttpStatusCode with class HttpStatus to allow custom status codes

jufickel-b:
Deprecated enum `HttpStatusCode` and introduced class `HttpStatus` as replacement.
This makes it possible to work with custom HTTP status codes as long as they are in the valid range (100-599).


thjaeckle:
@jufickel-b could you please adjust the PR title to be usable in some kind of changelog / release notes?

DerSchwilk (approved):
LGTM, minor changes in commit 652f88a

PR #941, opened 2021-01-18
Fix non-thread-safe calls to actor-starting methods in client actors.

yufei-cai:
Moved method calls with potential actor creation to the client actor's message handling thread.


thjaeckle (approved):
LGTM 👍

Issue #926, opened 2020-12-21
Support injection of temporary policy subjects from JWTs


yufei-cai:
After #890, temporary authorization subjects based on expiring JWT is possible. However, adding such subjects require WRITE permission on the policy. It would be better to inject the temporary subjects when the user presents a JWT without elevated privilege.

One way to achieve it:
- Add HTTP API to inject authorization subjects based on the JWT in the request
- Add a new permission EXECUTE to authorize subject injection

w4tsn:
Just to check back about the new situation with #512 being implemented now. If it's not already implemented, should there be a follow up for this issue to also provide ability to parse temporary subjects from JWT claims?

thjaeckle:
It is already implemented as part of this PR, see also the blogpost (and documentation) about it: 
https://www.eclipse.org/ditto/2021-01-22-policy-subject-activate-token-integration.html

PR #945, opened 2021-01-21
#926: inject configurable policy subjects into policies via new added JWT evaluating policy action

thjaeckle:
Fixes: #926 

A blogpost contained in the PR describes the feature.


yufei-cai (approved):
LGTM

PR #948, opened 2021-01-28
Improve rolling update behaviour

Yannic92:
This PR makes use of the rolling update improvements that [came with akka 2.6.10](https://doc.akka.io/docs/akka/2.6/additional/rolling-updates.html#cluster-sharding).

Furthermore akka-http version and akka-persistence-mongo version were lifted to the latest versions


thjaeckle:
@Yannic92 Akka released once again in order to fix a memory leak in streams: https://akka.io/blog/news/2021/01/28/akka-2.6.12-released
we should update :)

jbartelh (approved):

PR #949, opened 2021-02-02
Bugfix/cleanup

yufei-cai:
Add request-based event and snapshot cleanup to support the scenario that a small number of things generate the majority of events.


thjaeckle (approved):
LGTM

PR #951, opened 2021-02-04
Fix requesting "search-persisted" as ACK did not guarantee search consistency

thjaeckle:
The search consistency depends on the configuration of MongoDB writeConcerns, adjusted them accordingly or made possible to configure them for various DB access points in Ditto.
* added metrics for search consistency lag
* improved througput of search updates
* parallelized search index updates requesting "search-persisted" ACK and search index updates not requesting them
* cleaned up MongoDB config which partially was not even "pulled in"


yufei-cai (approved):
LGTM

PR #952, opened 2021-02-04
Make ClientPropsFactory extensible.

yufei-cai:
- Make ClientPropsFactory an extension point.
- Extract AMQP specific config handing into its own class.
- Change default jms.sendTimeout to 60s; 2s is too short.


thjaeckle (approved):
LGTM 👍

PR #953, opened 2021-02-04
Update deployment section

stmaute:
With this PR it is possible to deploy Ditto to a k3s cluster.
Updated kubernetes descriptor files to newer apiVersion and extend descriptor files.


thjaeckle (approved):
LGTM 👍

PR #955, opened 2021-02-05
fix default value for CONNECTING_MAX_TIMEOUT

stmaute:
Fix some connectivity client config and add subscription-manager-timeout key to connectivity config.


dguggemos (approved):
LGTM 👍

PR #959, opened 2021-02-08
Fix live-response acknowledgement handling

Yannic92:
This PR 

* removes the special handling of live-response acknowledgements to decide if an acknowledgement was successful or not
* Unwraps the Message for a live-response acknowledgement so the value of the live-response acknowledgement is now the payload of the message instead of the JSON representation of a message


jbartelh (approved):

jbartelh (approved):
LGTM

PR #965, opened 2021-02-12
re-added MongoDbUriSupplier in order to parse options

thjaeckle:
from a given/configured/injected MongoDB URI and apply additional options based on that base URI config

### DIFF
--- a/services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/config/DefaultOptionsConfigTest.java
+++ b/services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/config/DefaultOptionsConfigTest.java
@@ -12,6 +12,7 @@
  */
 package org.eclipse.ditto.services.utils.persistence.mongo.config;
 
+import static org.mutabilitydetector.unittesting.AllowedReason.assumingFields;
 import static org.mutabilitydetector.unittesting.AllowedReason.provided;
 import static org.mutabilitydetector.unittesting.MutabilityAssert.assertInstancesOf;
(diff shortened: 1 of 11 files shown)


services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/config/DefaultOptionsConfigTest.java
```diff
@@ -49,7 +50,9 @@ public void initMongoDbConfig() {
     public void assertImmutability() {
         assertInstancesOf(DefaultOptionsConfig.class,
                 areImmutable(),
-                provided(ReadPreference.class, ReadConcern.class, WriteConcern.class).areAlsoImmutable());
+                provided(ReadPreference.class, ReadConcern.class, WriteConcern.class).areAlsoImmutable(),
+                assumingFields("extraUriOptions")
+                        .areSafelyCopiedUnmodifiableCollectionsWithImmutableElements());
```
Yannic92, line 55: This assumption seems not to be correct. The map stored in DefaultOptionsConfig seems to be modfiable (and is provided by a getter)

Yannic92 (changes requested):

services/utils/persistence/src/test/java/org/eclipse/ditto/services/utils/persistence/mongo/config/DefaultOptionsConfigTest.java
thjaeckle: You're right 👍

thjaeckle:
@Yannic92 I made the map really immutable now, could you have another look?

Yannic92 (approved):
LGTM

PR #966, opened 2021-02-15
fixed serialization issue of AmqpClientActor.props 

thjaeckle:
which accepted a JmsConnectionFactory as argument (which is not serializable)
* provide a Config instead for applying overrides to the Amqp10Config


yufei-cai (approved):
LGTM

PR #970, opened 2021-02-16
AmqpSpecificConfig: support double-encoding of credentials

yufei-cai:
Add support for double-encoded credentials, which was the only way to use special characters in passwords of AMQP connections previously.


thjaeckle (approved):
LGTM :+1:

PR #973, opened 2021-02-17
Fix duplicate extension creation by QueryCriteriaValidator.

yufei-cai:
It is important that each extension ID stays unique according to object identity because they are used as lookup keys.


thjaeckle (approved):
LGTM 👍

Issue #512, opened 2019-10-07
OIDC integration should support arbitrary claims for authorization subjects


jokraehe:
The current implementation reads the "sub" claim only to create authorization subjects from a JSON web token. We should provide configuration options to support any claims.

Implementation suggestion (see below for discussion how we got there) for the configuration "syntax":
```
google = {
  issuer = "accounts.google.com"
  auth-subjects = [
    "{{ jwt:sub }}",
    "{{ jwt:sub }}/{{ jwt:scope }}",
    "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:client_id }}",
    "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:non_existing }}",
    "{{ jwt:roles/support }}"
  ]
}
```

w4tsn:
Is there any advance in this topic?

thjaeckle:
No, we don't have that on our short term agenda currently.

JulianFeinauer:
Just to get the issue clear. Is the aim to get arbitrary claims as the authorization subject or should the claims be used otherwise, also?

ffendt:
I think the idea behind this was to provide configuration on how (multiple) authorization subjects would be generated from the claims of a JWT. So basically a mini-DSL for generating the subjects.

So just a small example to clarify what I think this means:
For a token 
```
{
  "sub": "ffendt",
  "client_id": "ditto-users",
  "scp": [
    "user",
    "admin"
   ],
   ...
}
```
and a configuration for the issuer  `[sub]+[...scp][-@client_id]` (where  `'...' -> 'each array value'` and `'-' -> 'optional'`), Ditto would generate a list of auth subjects: 
```
[
  "ffendt+user@ditto-users",
  "ffendt+user",
  "ffendt+admin@ditto-users",
  "ffendt+admin"
]
```

w4tsn:
@JulianFeinauer @ffendt 

> Just to get the issue clear. Is the aim to get arbitrary claims as the authorization subject or should the claims be used otherwise, also?

In my understanding the main goal is to support e.g. roles and group claims in subjects such that I may add a OIDC role or group to a policy entry instead of a user id. This would already be much more flexible and allow access management being made from e.g. keycloak instead of ditto policies directly. With this I could write policy entries reflecting roles and add these roles to users to give them more capabilites.

[...]

PR #974, opened 2021-02-18
OIDC: Add support for arbitrary claims for authorization subjects

vhdirk:
See https://github.com/eclipse/ditto/issues/512 for what is being implemented here


services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/DefaultOAuthConfig.java
```diff
@@ -123,32 +125,33 @@ public String toString() {
...
             return HashMap::new;
         }
 
         @Override
-        public BiConsumer<Map<SubjectIssuer, String>, Map.Entry<String, ConfigValue>> accumulator() {
+        public BiConsumer<Map<SubjectIssuer, SubjectIssuerConfig>, Map.Entry<String, ConfigValue>> accumulator() {
             return (map, entry) -> map.put(SubjectIssuer.newInstance(entry.getKey()),
-                    entry.getValue().unwrapped().toString());
+                    // TODO: I don't know how to go back from a ConfigValue to a Config
+                    DefaultSubjectIssuerConfig.of(entry.getValue().atPath(entry.getKey()), entry.getKey()));
```
thjaeckle, line 144:
```suggestion
                    DefaultSubjectIssuerConfig.of(ConfigFactory.empty().withFallback(entry.getValue())));
```
This could/should work :)

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/DefaultSubjectIssuerConfig.java
```diff
@@ -0,0 +1,123 @@
...
+
+import javax.annotation.Nonnull;
+
+import com.typesafe.config.Config;
+
+import org.eclipse.ditto.model.policies.SubjectIssuerConfig;
+import org.eclipse.ditto.services.utils.config.ConfigWithFallback;
+import org.eclipse.ditto.services.utils.config.KnownConfigValue;
+
+public class DefaultSubjectIssuerConfig implements SubjectIssuerConfig {
```
thjaeckle, line 29:
```suggestion
public final class DefaultSubjectIssuerConfig implements SubjectIssuerConfig {
```
Please add `final` and javadoc for that public class.

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/DefaultSubjectIssuerConfig.java
```diff
@@ -0,0 +1,123 @@
...
+        issuer = configWithFallback.getString(SubjectIssuerConfigValue.ISSUER.getConfigPath());
+        authSubjectTemplates = configWithFallback.getStringList(SubjectIssuerConfigValue.AUTH_SUBJECTS.getConfigPath());
+    }
+
+    private DefaultSubjectIssuerConfig(final String issuer, final List<String> authSubjectTemplates) {
+        this.issuer = issuer;
+        this.authSubjectTemplates = authSubjectTemplates;
+    }
+
+    public static DefaultSubjectIssuerConfig of(final Config config, final String issuerName) {
```
thjaeckle, line 45: Please add javadoc for public methods.

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/DefaultSubjectIssuerConfig.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(issuer, authSubjectTemplates);
+    }
+
+    @Override
+    @Nonnull
+    public String toString() {
+        return issuer + authSubjectTemplates;
+    }
```
thjaeckle, line 99:
```suggestion
    @Override
    public String toString() {
        return getClass().getSimpleName() + " [" +
                "issuer=" + issuer +
                ", authSubjectTemplates=" + authSubjectTemplates +
                "]";
    }
```
That toString() does not make too much sense - concatenating a string with a List.
Suggesting our "default" toString() here..

services/gateway/util/src/test/resources/oauth-test.conf
```diff
@@ -1,10 +1,22 @@
...
+          "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:client_id }}",
+          "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:non_existing }}",
+          "{{ jwt:roles/support }}"
+        ]
+      }
   }
   openid-connect-issuers-extension = {
-    additional = "https://additional.google.com"
+    additional = {
+      issuer: "https://additional.google.com"
```
thjaeckle, line 17:
```suggestion
      issuer = "https://additional.google.com"
```

thjaeckle (changes requested):
Finishing first round of review.
Looks already really promising, very cool. Thanks for the contribution 👍 

The parsing for JsonArray values is IMO a strong "must" as JWT claims may also be a JsonArrays of strings, I commented accordingly inline to your "TODO".
When you add this, please also provide another unit test case testing with a JWT containing a JsonArray as claim.

thjaeckle:
@vhdirk you have one commit which did not include the "sign-off-footer" - but maybe you squash the commits again at the end, so we can worry about that later :)

thjaeckle:
@vhdirk the remaining test error is because of test file `devops-test.conf` still having the "old" format of the issuers.

model/policies/src/main/java/org/eclipse/ditto/model/policies/SubjectIssuerConfig.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.ditto.model.policies;
```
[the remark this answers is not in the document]
vhdirk: done

model/policies/src/main/java/org/eclipse/ditto/model/policies/SubjectIssuerConfig.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+
+public interface SubjectIssuerConfig {
```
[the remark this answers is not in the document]
vhdirk: done

vhdirk:
@thjaeckle I systematically forget to sign off my commits, so we'll have to squash this for sure :)

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/SubjectIssuerConfig.java
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.ditto.services.gateway.util.config.security;
+
+import java.util.List;
+
+/**
+ * Represents configuration for a {@link SubjectIssuer }, containing
```
thjaeckle, line 18: The `@link SubjectIssuer` won't work now because we are not in the same package and do not have that imported.
Suggesting:
```suggestion
 * Represents configuration for a {@link org.eclipse.ditto.model.policies.SubjectIssuer}, containing
```

services/gateway/security/src/main/java/org/eclipse/ditto/services/gateway/security/authentication/jwt/DittoJwtAuthorizationSubjectsProvider.java
```diff
@@ -58,11 +60,17 @@ public static DittoJwtAuthorizationSubjectsProvider of(final JwtSubjectIssuersCo
...
-                .map(subject -> SubjectId.newInstance(jwtSubjectIssuerConfig.getSubjectIssuer(), subject))
-                .map(AuthorizationSubject::newInstance)
-                .collect(Collectors.toList());
+
+        final ExpressionResolver expressionResolver = PlaceholderFactory.newExpressionResolver(
+                PlaceholderFactory.newPlaceholderResolver(JwtPlaceholder.getInstance(), jsonWebToken));
+
+        return jwtSubjectIssuerConfig.getAuthorizationSubjectTemplates().stream()
+            .map((subjectTemplate) -> expressionResolver.resolvePartially(subjectTemplate))
+            // TODO: needed?
```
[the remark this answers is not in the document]
vhdirk: done

services/gateway/util/src/main/java/org/eclipse/ditto/services/gateway/util/config/security/DefaultSubjectIssuerConfig.java
vhdirk: done

services/gateway/security/src/test/resources/oauth-test.conf
```diff
@@ -1,8 +1,20 @@
...
+        "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:client_id }}",
+        "{{ jwt:sub }}/{{ jwt:scope }}@{{ jwt:non_existing }}",
+        "{{ jwt:roles/support }}"
+      ]
+    }
   }
   openid-connect-issuers-extension = {
-    additional = "https://additional.google.com"
+    additional = {
+      issuer: "https://additional.google.com"
```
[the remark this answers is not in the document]
thjaeckle: @vhdirk this is still pending - another config as well used "issuer:"

thjaeckle:
Ah, some documentation on that would be nice as well: https://www.eclipse.org/ditto/installation-operating.html#openid-connect
There we also need to adjust the documented format anyways.

We could however also do that as follow-up task in order to not block the 2.0.0-M1 release.

thjaeckle:
@vhdirk I just noticed that the `gateway.conf` still uses the old pattern (for the only configured default "google":
https://github.com/eclipse/ditto/blob/master/services/gateway/config/src/main/resources/gateway.conf#L176

We have to adjust that, otherwise I fear the Ditto gateway won't start.

thjaeckle:
I fixed that in `gateway.conf` and resolved the conflict which was caused by that by merged master into this branch.

vhdirk:
Thanks!

thjaeckle:
@vhdirk I did try out the new feature in my local setup.
It basically works 👍 however .. when a `{{jwt:foo}}` placeholder cannot be resolved, it is added as unresolved placeholder to the list of SubjectIds, e.g. I used `"{{ jwt:sub }}/{{ jwt:scope }}"` instead of `"{{ jwt:sub }}/{{ jwt:scp }}"` which would have been correct and got at the `/whoami` HTTP API:
```json
{
    "defaultSubject": "mocky:bumlux",
    "subjects": [
        "mocky:bumlux",
        "mocky:bumlux/jwt:scope "
    ]
}
```

I'll look into where we can prevent this before merging the PR.

vhdirk:
Wow, that was quick. I was looking at it out of curiosity and was just going to ask how you wanted this handled :)

thjaeckle (approved):

thjaeckle:
Thanks again for this awesome contribution 🎉

PR #977, opened 2021-02-18
fixed that the "ssl" option in ditto-mongo.conf was not applied

thjaeckle:
... for MongoDB connections using Akka Persistence

* the configured "ssl" option is now added to the "extra-uri-options" if not configured explicitly there resulting in an adjustment of the used MongoDB uri string


dguggemos (approved):
LGTM :+1:

PR #979, opened 2021-02-19
Bugfix/cleanup

yufei-cai:
Stop considering policy and connectivity journal insert times for background cleanup so that those services may use a slower write concern.


thjaeckle (approved):
LGTM

PR #987, opened 2021-03-02
Use content-type to describe payload of ditto protocol message

Yannic92:
This PR corrects the handling of the content-type header in ditto protocol messages.
It now should consequently describe the payload of the ditto protocol message.


thjaeckle (approved):
LGTM 👍 
Just fixed a dead link in the docs

PR #988, opened 2021-03-02
Bugfix for accidentally applied default header mappings

Yannic92:
We applied a default target header mapping for all types of connections (including mqtt, which should not support them at all).
This PR does check for the connection type before migrating a target.
Additionally this PR makes headerMapping of sources and targets non-nullable so new connections are not affected from this live migration.


thjaeckle (approved):
LGTM 👍

Issue #991, opened 2021-03-03
Allow multiple policies per Thing


vhdirk:
As discussed with @thjaeckle over gitter, we're looking for a way to assign multiple policies to a Thing. That would make granular permission control a lot easier to implement.

Below a simple demonstration of what we'd like to achieve:
![multiple-policies](https://user-images.githubusercontent.com/1424486/109845953-f59d8800-7c4d-11eb-9ee1-c632de1f3c83.png)
Take for example a situation where Things are installed at 2 physical locations. Operators on site would have permission to view properties of the Thing, but not modify (Teams A and B). Overseeing all operations is Team X, which has permission to modify/delete every Thing. Additionally, there's a maintenance team (Team C) that can travel to both physical locations, but only has permission to change specific features of Things (not necessarily restricted that much), for a specific subset of Things.

To achieve something like that, it would be (for a ditto developer) a lot easier if things could be assigned multiple policies.

Compare that to what we currently would have to configure:
![single-policy](https://user-images.githubusercontent.com/1424486/109845958-f6361e80-7c4d-11eb-9778-ccded1331ced.png)

We understood that this may have a negative performance impact on the thingssearch index, as it would have to be rebuilt if a policy is modified. However, whether using multiple policies or not, that bottleneck is always going to be there.

thjaeckle:
Just to clarify: #298 would also let you do exactly what you described, won't it?
Because I think that this would API-wise be a better enhancement than e.g. adding an optional `"additionalPolicyIds"` to the Thing JSON.

vhdirk:
Yes, indeed. #298 would make this situation possible, but it has some drawbacks, too.

What we're proposing here would mean breaking the current API: a Thing would really just take an array of `PolicyId`s, rather than a single one. This would keep the possibility to filter things based on policyId. With the 'import' proposal made in #298, that would require an indirection: You'd first have to find all policies that import the one you're querying, and then match those with the actual devices.

thjaeckle:
I'm afraid that breaking the JSON API is not a short- to midterm option, at least not until Ditto requires other JSON breakages in the "thing JSON" which would justify an "API schema version 3" (e.g. `/api/3` at HTTP level). 
With Ditto 2.0 we remove "API schema version 1" (ACL based) and maintaining several schema versions at the same time is a huge amount of additional maintenance work.

[...]

Issue #298, opened 2018-11-27
Let policies import other policies to enable re-use when securing things


zubairhamed:
Hi,
I wonder how this can be done/modeled:

Imagine Things owned by a company. Within the company you have 2 access/users. One which has ADMIN rights, the other is a TECHNICAL user which has perhaps READ rights for only specific sensitive features. Such rights would be applied to all things.

However, there would be a different class of users who have perhaps rights to another set of features. This could be domain data such as sensor data or device state but NOT the sensitive keys mentioned above. Additionally each user would have rights to only variable set of things and not all of them.

Using a single policy would mean:
1. Maintaining a one-to-one relation between policies and things
2. make it very tideous to maintain such relationships programmatically.

Example:
![folie1](https://user-images.githubusercontent.com/2336402/49063607-30ffaf80-f218-11e8-8c1f-585ebd60791c.JPG)

Is there a solution to this? Am i missing something?

if not, perhaps allowing multiple policies to be applied per thing might make sense. (i.e. one policy per user or actor instead of per asset).
![folie2](https://user-images.githubusercontent.com/2336402/49063889-31e51100-f219-11e8-829b-84aeaa39d47f.JPG)

WDYT?

Z

thjaeckle:
Hi Zubair.
I recently thought about something which might help you (but I'm afraid this is only a first idea). 
But would be great to get a feedback whether this would also help your use case.

The idea is to enable "importing" other policies in a policy, e.g.:
```json
{
  "policyId": "org.eclipse.ditto:solution-policy",
  "entries": {
    "solution-owner": {
    ...
    }
  }
}
```
The above could be imported:
```json
{
  "policyId": "org.eclipse.ditto:customer-policy",
  "imports": {
    "org.eclipse.ditto:solution-policy": {
      "included": [], // optionally specify includes/excludes 
      "excluded": ["end-user"] // if both are empty or missing, all entries are included
    }
  },
  "entries": {
    "end-user": { // we have to think about what to do with overwrites
    ...
    }
  }
}
```

That would cause that for the second policy, the first policy is imported and merged.
There is no special overwrite semantics - the standard policy way applies: 
* deeper nested hierarchy is more important than less nested 
* on the same hierarchy level `revoke` is stronger than `grant`

Benefits of that:
* policies basically can inherit from other policies (and even multiple ones)
* a single "solution policy" could contain administrative entries - once that is changed, all policies and therefore things importing from that are automatically adjusted

[...]

PR #993, opened 2021-03-04
WIP: Import policies in a policy

vhdirk:
This is an implementation for #298 (referenced by #991)

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/signals/FeatureToggle.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/signals/FeatureToggle.java
@@ -28,9 +28,22 @@ public final class FeatureToggle {
     public static final String MERGE_THINGS_ENABLED = "ditto.devops.feature.merge-things-enabled";
 
     /**
+     * System property name of the property defining whether the policy import feature is
+     * enabled.
+     *
+     * @since 2.x.0 TODO TJ
+     */
+    public static final String POLICY_IMPORTS_ENABLED = "ditto.devops.feature.policy-imports-enabled";
+
+    /**
      * Resolves the system property {@value MERGE_THINGS_ENABLED}.
      */
     private static final boolean IS_MERGE_THINGS_ENABLED = resolveProperty(MERGE_THINGS_ENABLED);
+
+    /**
+     * Resolves the system property {@value MERGE_THINGS_ENABLED}.
+     */
+    private static final boolean IS_POLICY_IMPORTS_ENABLED = resolveProperty(POLICY_IMPORTS_ENABLED);
 
     private static boolean resolveProperty(final String propertyName) {
         final String propertyValue = System.getProperty(propertyName, Boolean.TRUE.toString());
@@ -52,6 +65,49 @@ public final class FeatureToggle {
      */
     public static DittoHeaders checkMergeFeatureEnabled(final String signal, final DittoHeaders dittoHeaders) {
         if (!IS_MERGE_THINGS_ENABLED) {
+            throw UnsupportedSignalException
+                    .newBuilder(signal)
+                    .dittoHeaders(dittoHeaders)
+                    .build();
+        }
+        return dittoHeaders;
+    }
+
+    /**
+     * Checks if the policy import feature is enabled based on the system property {@value POLICY_IMPORTS_ENABLED}.
+     *
+     * @param signal the name of the signal that was supposed to be processed
+     * @param dittoHeaders headers used to build exception
+     * @param message the error message
+     * @return the unmodified headers parameters
+     * @throws UnsupportedSignalException if the system property {@value POLICY_IMPORTS_ENABLED} resolves to
+     * {@code false}
+     * @since 2.x.0 TODO TJ
+     */
+    public static DittoHeaders checkPolicyImportsFeatureEnabled(final String signal, final DittoHeaders dittoHeaders,
+            final String message) {
+        if (!IS_POLICY_IMPORTS_ENABLED) {
+            throw UnsupportedSignalException
+                    .newBuilder(signal)
+                    .dittoHeaders(dittoHeaders)
+                    .message(message)
+                    .build();
+        }
+        return dittoHeaders;
+    }
+
+    /**
+     * Checks if the policy import feature is enabled based on the system property {@value POLICY_IMPORTS_ENABLED}.
+     *
+     * @param signal the name of the signal that was supposed to be processed
+     * @param dittoHeaders headers used to build exception
+     * @return the unmodified headers parameters
+     * @throws UnsupportedSignalException if the system property{@value POLICY_IMPORTS_ENABLED} resolves to
+     * {@code false}
+     * @since 2.x.0 TODO TJ
+     */
+    public static DittoHeaders checkPolicyImportsFeatureEnabled(final String signal, final DittoHeaders dittoHeaders) {
+        if (!IS_POLICY_IMPORTS_ENABLED) {
             throw UnsupportedSignalException
                     .newBuilder(signal)
                     .dittoHeaders(dittoHeaders)
--- a/concierge/service/src/main/java/org/eclipse/ditto/concierge/service/common/CachesConfig.java
+++ b/concierge/service/src/main/java/org/eclipse/ditto/concierge/service/common/CachesConfig.java
@@ -38,6 +38,13 @@ public interface CachesConfig {
     CacheConfig getIdCacheConfig();
 
     /**
+     * Returns the config of the policy cache.
+     *
+     * @return the config.
+     */
+    CacheConfig getPolicyCacheConfig();
+
+    /**
      * Returns the config of the enforcer cache.
      *
      * @return the config.
--- a/concierge/service/src/main/java/org/eclipse/ditto/concierge/service/common/DefaultCachesConfig.java
+++ b/concierge/service/src/main/java/org/eclipse/ditto/concierge/service/common/DefaultCachesConfig.java
@@ -36,11 +36,13 @@ public final class DefaultCachesConfig implements CachesConfig {
 
     private final AskWithRetryConfig askWithRetryConfig;
     private final CacheConfig idCacheConfig;
+    private final CacheConfig policyCacheConfig;
     private final CacheConfig enforcerCacheConfig;
 
     private DefaultCachesConfig(final ScopedConfig config) {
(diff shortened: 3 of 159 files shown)


thjaeckle:
@vhdirk I will try to merge "master" onto my 2 year old experiment of adding policyImports - that should give a really good start for putting additional required work on top of it (e.g. how to handle search index updates then).

vhdirk:
Oh, that'd be very cool! Thanks!

thjaeckle:
I pushed a branch: https://github.com/eclipse/ditto/tree/lab/policy-imports
I invested a little trying to merge master onto that - I think from Policy Model side I got very far - however there still is a lot of stuff not even compiling - I did not even come to the tricky part the "policy enforcement" and the policy cache loading / cache invalidation, etc.

To be honest: I think without deep understanding of the Ditto codebase this will be really hard to get finished.

thjaeckle:
@vhdirk I got the build working - did some first rough code adjustments right on your branch, I hope you don't mind.
Will look into the functionality tomorrow.

What still is missing are unit tests, e.g. for:
* ImmutableEffectedImports
* ImmutableImportedLabels
* ImmutablePolicyImport
* ImmutablePolicyImports
* LabelInvalidException
* PolicyImportHelper
And also for:
* PolicyCacheLoader

thjaeckle:
@vhdirk I added a "feature toggle" for the policy import feature - we recently already added something like this for the "merge things" feature.

I started Ditto on this feature branch, however the imported policies were not yet applied - I will now look into this, but I fear that currently the policy import is not yet working.

thjaeckle:
@vhdirk I invested some time in getting the functionality to actually "work".
* For direct access to a thing/policy (knowing their IDs) I would say that the policy import does work perfectly

We have *a few open topics before being able to merge this PR*.

# search index

However, the search index is only correctly updated for the "happy path":
* when a `ThingUpdater` responsible for updating a thing with ID `x:y-1` is currently "alive" (is running), updating a policy which is imported in other policies will correctly update the things using the imported policy
   * so this is only the case when changes were recently done on a thing and e.g. not on a cluster restart

This is how the search index is updated when a policy is changed:
1. the `PolicyEvent` is received in `PolicyEventForwarder` of the search service
2. `PolicyEventForwarder` finds out all things which (directly) use this policy (via `ThingsSearchUpdaterPersistence.getPolicyReferenceTags(Map)` and creates for each matched thing one `PolicyReferenceTag` message
3. each created `PolicyReferenceTag` message is locally forwarded to `ThingsUpdater`
4. the `ThingsUpdater` invalidates the `policyCache` of that policy (on the local "search" instance)
5. the `ThingsUpdater` sends each `PolicyReferenceTag` via cluster sharding to a `ThingUpdater` in a search instance "responsible" for the shard of the thingId (consistently always the same instance for the same ID)

The problem is now that step `2.` does not yet respect imported policies when calculating the `PolicyReferenceTag` message based on the update of a policy which was imported.
* therefore, we would need to adjust the MongoDB collection `searchThings` (or introduce another persistent MongoDB collection) in which such information is persisted
* basically we must answer the question: "which things are affected (because they link to a policy 'B' importing policy 'A') when policy 'A' is modified?"
* in the implementation `MongoThingsSearchUpdaterPersistence.getPolicyReferenceTags(Map)` those affected policyIds must additionally be returned as `PolicyReferenceTag`s

When enhancing the search index to know that we:
* must ensure to stay backwards compatible (existing indexed things must still be found the same as before)
* must look into MongoDB indexes - so that those queries are fast
* have the freedom of add something like `additionalPolicyIds` - as the search index is "only" used in order to determine thingIds matching a search query .. the structure of it is never showed to the Ditto API user

# importable

And one more thing we found in a peer-review with the team:
* a (potentially to be imported) policy must be able to declare single entries as "not importable" - otherwise when importing a complete policy there could be entries imported which are not desired to "include"
* as when creating existing policies this had not to be thought about we think that it's "safer" to assume that existing policies shall not be imported

We propose to add an optional `"importable": true` (the default value to assume if the field is not present is `false` - following the *principle of least privilege*) JSON field to the Policy Entry, e.g.:
```json
{
  "policyId": "foo:bar",
  "entries": {
    "SUPER_ADMIN": {
      "importable": false,
      "subjects": {
        ...
      },
      "resources": {
        ...
       },
...
```

# summing up
1. Search must be fixed to consistently update things in the index when imported policies are changed
2. Support *principle of least privilege* by adding "importable" enum to PolicyEntry defaulting to `false` - that way importing must be explicitly enabled
3. Documentation (OpenAPI + Ditto website docs) is missing
4. Additional (integration) tests testing applying the policies, e.g. added to `ThingCommandEnforcementTest`

I would say we are 70% through :-)

services/thingsearch/persistence/src/main/java/org/eclipse/ditto/services/thingsearch/persistence/write/streaming/EnforcementFlow.java
```diff
@@ -298,4 +315,76 @@ private static boolean shouldReloadCache(@Nullable final Entry<?> entry, final M
...
+                        return POLICY_NONEXISTENT;
+                    });
+
+            return Source.completionStageSource(policyFuture);
+        });
+
+        return lazySource.viaMat(Flow.create(), Keep.none());
+    }
+
+    private Source<List<Entry<Policy>>, NotUsed> readCachedImportedPolicies(final Metadata metadata,
```
vhdirk, line 367: @thjaeckle I think this is somewhat like you meant on gitter?

thjaeckle: @vhdirk hm, almost :)
I don't get why you load the complete policy of the "imported policies" - for the `List<PolicyTag> policyImports`  passed into the `EnforcedThingMapper.toWriteModel(..)` method you don't need to load the `Policy`, the id of the imported policies + the revision number is sufficient.
I would simplify the return types of the private "read..." methods. Also, I don't think that you have to use `Entry` there as wrapping cache entry type - just use the `PolicyTag` or the list of dertermined `PolicyTag`s.

I would expect a method signature:
```
private Source<List<PolicyTag>, NotUsed> getPolicyImports(final PolicyId policyId);
```
Which can simply be called in addition to `getEnforcer` to get input for the to be called `EnforcedThingMapper.toWriteModel()`.

vhdirk: Ok, I'll refactor it. You weren't kidding when you said that this requires messing about with some deep internals :p

vhdirk: I suppose it also doesn't quite make sense to reload the cached entry? https://github.com/vhdirk/ditto/blob/7d57d4f590300b14d242cdb299d31beaf04c90db/services/thingsearch/persistence/src/main/java/org/eclipse/ditto/services/thingsearch/persistence/write/streaming/EnforcementFlow.java#L362

vhdirk: While refactoring, I think I do need to load at least the thing policy; it may be that a cached version of a policy that is imported by the 'current' policy already exists. We won't know that unless we fetch the latest revision of the 'current' policy, or do we?

thjaeckle: Good point 👍 
The `Metadata` object contains the revision number of the used policy of a Thing update processed in `ThingUpdater`.
However at this point it can't be known whether any of the referenced imported policies have also changed. Even if all cache invalidator callbacks would work as expected, due to the sharding when horizontally scaling the "things-search" service, this information is not consistently known everywhere.

So I think you're right: when a policy uses imported policies, the imported policies (to be precise: their revision number) must be loaded OR the cache entries in the `policyCache` for those must be invalidated which will cause an automatic new loading of them.

This would make it a very expensive (bandwith and CPU consumption wise for deserialization) operation via the "normal" `policyCache` - as this cache always asks for the full Policy JSON.

TL;DR: So I fear we need another cache which holds `PolicyTag`s.

vhdirk: And for such a new cache to work, I guess we're also going to need a new `sudo` command, `SudoRetrievePolicyRevision` or so? Or, perhaps `SudoRetrievePolicyRevisions`, since we know it's going to be a list.

thjaeckle: Yes. You mean like this already existing one `org.eclipse.ditto.services.models.policies.commands.sudo.SudoRetrievePolicyRevision`? ;)

vhdirk: And to know I did actually look for something like that. Time for more coffee :)
Aside from that; this MR is getting pretty big. Do you think we could/need split it up somehow?

thjaeckle: I don't think that splitting up is useful - we wouldn't want to merge a half-working feature to our master branch.
Documentation and integration tests could be provided in another PR afterwards.

And with Ditto 2.0 to be released next week, there will be some (huge, I fear) merge efforts as well on the overall PR - I can support on that.

thjaeckle:
@vhdirk @BobClaerhout I rebased this PR on the current Ditto master branch (took quite some effort because of some breaking Ditto 2.0 module changes) - I referenced @vhdirk as co-author in the commit.

Summarizing there are still some open TODOs:

* My comment https://github.com/eclipse/ditto/pull/993#discussion_r615890894 still is to be addressed - however I think that it would be easier to achieve when simply invalidating imported and affected policies cluster-wide when a policy was changed which was used in other policies as import
   * see my "// TODO TJ" comment in `ThingsUpdater.processPolicyReferenceTag()`
* Documentation (OpenAPI + Ditto website docs) is missing
* Additional (integration) tests testing applying the policies, e.g. added to `ThingCommandEnforcementTest`

thjaeckle:
Superseded by #1555

PR #996, opened 2021-03-05
Bugfix/password decoding

Yannic92:
This PR fixes an issue in AMQP password decoding.
Previously the decoding was made by using `java.net.URLDecoder` which is meant to be used on `x-www-form-urlencoded` payload.
This decodes the `+` character to a space.

In this PR all `+` characters are replaced by their encoded representation `%2B` before calling the `java.net.URLDecoder`

Additionally I've made an improvement of the error description for failure when sending messages to or from a thing.

### DIFF
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpSpecificConfig.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpSpecificConfig.java
@@ -147,7 +147,6 @@ public final class AmqpSpecificConfig {
         final var username = connection.getUsername();
         final var password = connection.getPassword();
         if (username.isPresent() && password.isPresent()) {
-            // URL-decode the username and password to preserve previous behavior (no apparent encoding of parameters)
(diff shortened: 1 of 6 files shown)


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpSpecificConfig.java
```diff
@@ -148,8 +148,8 @@ private static void addCredentials(final LinkedHashMap<String, String> parameter
         final var password = connection.getPassword();
         if (username.isPresent() && password.isPresent()) {
             // URL-decode the username and password to preserve previous behavior (no apparent encoding of parameters)
```
thjaeckle, line 150: This comment above is no longer valid, is it?

Yannic92: oops, no it isn't. Will remove it.

thjaeckle (approved):
LGTM 👍

PR #997, opened 2021-03-05
Add test for policy revision update in search index; fix timer duplication.

yufei-cai:
- Add `EnforcementFlowTest` to test that thing and policy revisions are updated correctly in the search index.
- Remove duplicate timers in  `EnforcedThingMapper`.
- Force cache reload when requested for search index update.


thjaeckle (approved):
LGTM 👍

PR #1003, opened 2021-03-15
Parse the search criteria before validating it.

Yannic92:
This will ensure that a rql.expression.invalid exception is thrown if e.g. a field named 'a' is
used in the filter


thjaeckle (approved):
LGTM 👍

PR #1006, opened 2021-03-17
Introduce ConnectionUnauthorizedException

Yannic92:
This exception is thrown for invalid credentials in a connection that should be created.


thjaeckle (approved):
LGTM 👍

PR #1008, opened 2021-03-18
Validate extra field selection for connection topics

Yannic92:
This PR validates the extra field selection of connection topics.
This should improve the management of connections as it indicates very early, what's wrong with the connection configuration.


thjaeckle (approved):
LGTM 👍 
Just added some javadoc

PR #1011, opened 2021-03-22
Support for building arm64 ditto snapshot images

ghandim:
Dockerfile for support of building arm64 based ditto snapshot images.
Signed-off-by: Michael Gantert <michael.gantert@bosch.io>

### DIFF
--- a/services/dockerfile-snapshot-arm64
+++ b/services/dockerfile-snapshot-arm64
@@ -1,0 +1,44 @@
+# Copyright (c) 2021 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+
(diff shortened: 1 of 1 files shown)


thjaeckle:
Awesome @ghandim 
Do we maybe already want to add a "release" dockerfile as well and also publish releases to Docker Hub?
Any suggestions for Docker tagnames for such released images?

Then we should also advertise this somewhere in the "installation" or "operation" documentation, that there is the option to run on ARM64.

thjaeckle:
Apparently the "right" way to do this is to use docker "buildx":
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Example: https://hub.docker.com/r/ckulka/multi-arch-example

ghandim:
@thjaeckle I thought also to that this could be the next steps. OpenJ9 prefixes their image tags with the architecture, e.g. "aarch64-centos-jdk-11.0.10_9_openj9-0.24.0". e.g. Don't know if we need/have a special build environment for arm based builds. Otherwise we have to go for multi architecture build like described here https://admantium.medium.com/docker-building-images-for-multiple-architectures-4f142f6dda71.

thjaeckle (approved):
LGTM as first step (building snapshot image locally)

Further steps towards multi-arch images have to be discussed.

ghandim:
@thjaeckle Could also be discussed in the context of the packages project, or?

thjaeckle:
@ghandim I don't see it there .. as producing the Docker images is responsibility of the projects, not of the packages project

PR #1014, opened 2021-03-23
Handle search error responses in Ditto protocol adapter

stmaute:
We do not handle search error responses in the Ditto protocol adapter.
This PR should fix this.


thjaeckle (approved):
LGTM 👍

PR #1015, opened 2021-03-23
fixing a potential NPE in ThingPredicatePredicateVisitor

thjaeckle:
* thrown when using a JsonObject or an JsonArray as filter target
* added basic unit test for ThingPredicatePredicateVisitor
* fixed another found bug which treated integers as longs and compared them wrongly


stmaute:
LGTM 👍🏻

PR #1016, opened 2021-03-30
Get connectivity config via DittoHeaders

stmaute:
Extend the ConnectivityConfigProvider interface with an additional method to retrieve the connectivity config via DittoHeaders.

### DIFF
--- a/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
+++ b/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
@@ -58,8 +58,8 @@ final class HostValidator {
      */
     HostValidator(final ConnectivityConfig connectivityConfig, final LoggingAdapter loggingAdapter,
             final AddressResolver resolver) {
-        this.allowedHostnames = connectivityConfig.getConnectionConfig().getAllowedHostnames();
         this.resolver = resolver;
+        this.allowedHostnames = connectivityConfig.getConnectionConfig().getAllowedHostnames();
         final Collection<String> blockedHostnames = connectivityConfig.getConnectionConfig().getBlockedHostnames();
         this.blockedAddresses = calculateBlockedAddresses(blockedHostnames, loggingAdapter);
     }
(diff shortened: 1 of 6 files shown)


services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
```diff
@@ -58,19 +58,19 @@
...
         this.blockedAddresses = calculateBlockedAddresses(blockedHostnames, loggingAdapter);
     }
 
     /**
      * Validate if connections to a host are allowed by checking (in this order):
      * <ul>
-     *     <li>blocklist is empty? this completely disables validation, every host is allowed</li>
-     *     <li>host is contained in allowlist? host is allowed</li>
+     *     <li>blocklist is empty this completely disables validation, every host is allowed</li>
+     *     <li>host is contained in allowlist host is allowed</li>
```
thjaeckle, line 71:
```suggestion
     *     <li>if the blocklist is empty, this completely disables validation, every host is allowed</li>
     *     <li>if the host is contained in the allowlist, the host is allowed</li>
```

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
```diff
@@ -58,19 +58,19 @@
...
      *     <li>host is resolved to a blocked ip (loopback, site-local, multicast, wildcard ip)? host is blocked</li>
```
thjaeckle, line 72:
```suggestion
     *     <li>if the host is resolved to a blocked ip (loopback, site-local, multicast, wildcard ip), the host is blocked</li>
```

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
```diff
@@ -58,19 +58,19 @@
...
-     *     <li>host is contained in the blocklist? host is blocked</li>
+     *     <li>host is contained in the blocklist host is blocked</li>
```
thjaeckle, line 73:
```suggestion
     *     <li>if the host is contained in the blocklist, the host is blocked</li>
```

thjaeckle (changes requested):
I think the javadoc list should be rephrased - good that you removed the `?` in the comment, that was already confusing - I suggested to make complete sentences out of it.

services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/validation/HostValidator.java
stmaute: It makes sense to extend the javadoc in this way 👍🏻

thjaeckle (approved):

PR #1018, opened 2021-03-31
Priorization of connections for reconnecting on startup

Yannic92:
Prioritize very active connections over inactive connections for reconnecting

* The higher the priority, the earlier it will be reconnected on startup
* The priority is updated automatically in an interval that can be configured
  by ditto.connectivity.connection.priority-update-interval.
* The priority is the sum of consumed and dispatched messages in the last
  24 hours (based on the collected connection metrics)


thjaeckle (approved):
LGTM 👍 
Pushed one commit with review findings.

PR #1022, opened 2021-04-07
Improves the priority ordering in MongoReadJournal

Yannic92:
* Defines a prefix 'priority-'
* removes irrelevant sorting step
* adds projection step to remove all other tags than 'priority-.*' to avoid that other tags can have impact on the sorting of events


thjaeckle (approved):
LGTM 

I adjusted the query to be still MongoDB 3.6 compatible.
And I renamed the method and javadoc in order to reflect that sorting is always done by the priority tag.

PR #1024, opened 2021-04-09
Validate JWT signature also in default validator

ffendt:
We're validating for an existing signature in JWTs in our default implementation `ImmutableJsonWebToken`, which will then be validated in `DefaultJwtValidator#validateWithPublicKey:85`.
However, the used [`io.jsonwebtoken.JwtParser#parse(String)`](https://github.com/jwtk/jjwt/blob/a4130dd1ec5d4bef4b1835927a2a1d966a519109/api/src/main/java/io/jsonwebtoken/JwtParser.java#L428) basically also would work for unsigned tokens. Therefore changed to [`io.jsonwebtoken.JwtParse#parseClaimsJws(String)`](https://github.com/jwtk/jjwt/blob/a4130dd1ec5d4bef4b1835927a2a1d966a519109/api/src/main/java/io/jsonwebtoken/JwtParser.java#L594) which will throw an exception if the JWT isn't signed.


thjaeckle (approved):
LGTM 👍

PR #1025, opened 2021-04-09
Update JS payload mapping scripts

ffendt:
Use sensible defaults, which also work per default for Ditto protocol messages.


thjaeckle (approved):
LGTM 👍 
* simplified the ditto-scope.js helper functions a little
* added one missing replacement of "id" with "name"

PR #1028, opened 2021-04-14
Fix ddata sharding message duplication

thjaeckle:
When sharding was introduced into Ditto pub/sub in milestone 2.0.0-M2, a bug was introduced that when using "groups" the messages/events were duplicated if > 1 shard count was configured.

This PR fixes this bug and configures the default sharding count to 5 (instead of defaulting to 1).


stmaute:
LGTM 👍🏻

PR #1030, opened 2021-04-19
Don't return IDs of deleted Connections in RetrieveAllConnectionIdsResponse

Yannic92:
Stream journal entries instead of just the pid string

* This allows to distinguish between deleted or other events in
  ConnectionIdsRetrievalActor
* Skip PIDs of connections of which the last event ist a deleted event
  in ConnectionIdsRetrievalActor


thjaeckle (approved):
LGTM 👍

Issue #1000, opened 2021-03-09
Ditto 2.0


stmaute:
As we already started the discussion in #978 the work on Ditto 2.0 is in progress.

We keep #978 for the discussion on the topics Ditto 2.0 aims to "break" and use this issue to track progress on the actual doing.

Issue #978, opened 2021-02-19
[Discussion] The road to Ditto 2.0 with some breaking changes on the way


Yannic92:
As we plan to release Ditto 2.0 in the near future (March/April 2021), it's time to discuss some breaking changes helping the Ditto core team in the future development (reducing maintenance of e.g. already `@Deprecated` marked APIs) but also keep an eye on the effort that would be required by our users to migrate from Ditto 1.x to Ditto 2.x.

thjaeckle:
Here is what we currently foresee as "breaking changes" regarding our "API compatibility"/semantic versioning definition in https://github.com/eclipse/ditto/blob/master/documentation/src/main/resources/architecture/DADR-0005-semantic-versioning.md#api-compatibility

We mainly focus on binary incompatible changes to Ditto's model classes - the HTTP APIs (in API version `2`) are not affected and are still compatible. API version `1` will be dropped as we already deprecated API version `1` some time ago.

---

## 1. remove `setDittoHeaders(DittoHeaders)` from `WithDittoHeaders`
Not every class implementing `WithDittoHeaders` should also be capable of setting `DittoHeaders`. Extract a new interface for setting `DittoHeaders`.

## 2. Rework default connection behavior of Ditto Java client
By default a newly created Ditto client shall not connect itself automatically, but only when a `connect()` method is explicitly invoked. As this requires a breaking API change, this is listed here.

## 3. Make `AbstractDittoHeaders` implement `Map<String,String>`
..

## 4. Java Client API handlers should return `CompletionStage` instead of `CompletableFuture`
In Ditto Java Client 1.x all APIs return a `CompletableFuture`. That is bad as the Ditto client user may complete this `CompletableFuture` by himself which could break things in the client.  
Therefore all APIs formerly returned `CompletableFuture` will return a `CompletionStage` instead - on this the Ditto client user may still call `.toCompletableFuture()` in order to get a "new" CompletableFuture.

## 5. Make `DittoRuntimeException` abstract
We don't want to instantiate `DittoRuntimeException` itself, but only concrete implementation with a better context about the exception.

## 6. Remove dependency of placeholders model on all signals
In order to get some dependency 

## 7. Move live signals to Ditto Java client
The `ditto-signals-*-live` modules are only used in the Java client, so move it there.

## 8. Remove API 1 from models, signals, routes
API 1 (with the "ACL" based authorization) will be completely removed, both from the codebase as well as from the HTTP endpoints.

[...]

PR #1031, opened 2021-04-21
First part of Ditto 2.0

Yannic92:
This PR contains the first part of Ditto 2.0 (discussed in #1000 and #978).

Contained parts of #978 are:

* remove setDittoHeaders(DittoHeaders) from WithDittoHeaders
* Rework default connection behavior of Ditto Java client
* Make AbstractDittoHeaders implement Map<String,String>
* Java Client API handlers should return CompletionStage instead of CompletableFuture
* Make DittoRuntimeException abstract
* Remove dependency of placeholders model on all signals
* Move live signals to Ditto Java client
* Remove API 1 from models, signals, routes
* Fix EntityId+WithId and remove from Signal interface
* Enforce Non-nullable in PolicyResponses
* Remove all APIs marked with @Deprecated
* Remove all remaining raw types where generics should be used

The last part: Re-structure Ditto's maven modules + packages will come with a separate PR


thjaeckle (approved):
Reviewed in a peer group of @Yannic92 @jbartelh @stmaute @yufei-cai and myself.

PR #1033, opened 2021-04-22
Allow setting mqtt retain message flag depending on header

### LINKED ISSUE #1029: Allow setting MQTT retain flag depending on message header
The retain flag of MQTT messages published via a managed connection is set according to the message header `mqtt.retain`. The topic and qos level can also modified in the same way.

dguggemos:
Fixes #1029.


VadimGue:
LGTM

PR #1035, opened 2021-04-22
Simplified ditto-model-query dependencies + API

thjaeckle:
Got rid of ditto-model-rql-parser dependency in ditto-model-query
* hid implementations of query expresssions behind interfaces
* made a lot of ditto-model-query classes package private
* created factory methods for creating instances


stmaute (approved):
LGTM 👍🏻

PR #1037, opened 2021-04-23
Cleaned up Ditto error documentation

thjaeckle:
* fixed Ditto Protocol examples "topic" - removed "unknown" from them
* included correlation-id header in examples
* made clear that "error" code cannot be expected to be stable API
* adjusted json schemas
* added some links
* did minor navigation restructuring
* harmonized used thing- and policy-ids


stmaute:
LGTM 👍🏻

PR #1038, opened 2021-04-26
Update docker base image for ditto container images  

stmaute:
Use new version jdk-11.0.11_9_openj9-0.26.0-alpine-slim of adoptopenjdk/openjdk11-openj9 .

Signed-off-by: Stefan Maute <stefan.maute@bosch.io>


thjaeckle (approved):
LGTM

PR #1039, opened 2021-04-27
Enable connection logging when creating or opening a connection

VadimGue:
When creating or opening a connection the connection logging is enabled per default. This allows to log possible errors on connection establishing.


connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/persistence/ConnectionPersistenceActorTest.java
```diff
@@ -1064,9 +1087,13 @@ public void disabledConnectionLogsAreNotEnabledAfterModify() {
 
             // create connection
             underTest.tell(createConnection, getRef());
+            probe.expectMsg(enableConnectionLogs);
             probe.expectMsg(openConnection);
             expectMsg(createConnectionResponse);
 
+            //Close logging which are automatically enabled via create connection
+            underTest.tell(LoggingExpired.of(connectionId, DittoHeaders.empty()), getRef());
```
Yannic92, line 1095: This seems to change the nature of this test. If I understand correctly, this PR should change the behaviour this test expects, so I'd delete this test.

VadimGue: With this we want to simulate a connection that has been in use for a longer time and therefore the logging (which is enabled on connection opening) is disabled

Yannic92: Then I suggest to rename the test accordingly.

dguggemos: I think the added line prepares the actor state (disabled logging) which is required to test if logging is affected by the modify command. No need to rename IMO.

dguggemos (approved):
LGTM

PR #1040, opened 2021-04-27
Fixing remaining known Ditto 2.0 bugs/issues

thjaeckle:
* fixed bug in PublisherIndex which led to missing Ditto pub/sub messages in the clust
* removed authContext deduplication in AbstractDittoHeaders, no longer required with API 1 removed
* made ConnectionId also accept ":" in the ID in order to fix "lower bound" based cleanup
* added logging for enforcements (which headers were calculated)
* added logging about selected subscribers for Ditto pub/sub publishing


stmaute:
LGTM 👍🏻

PR #1041, opened 2021-04-28
Support reading JWT from query parameter for WebSocket requests

### LINKED ISSUE #667: Provide JWT tokens to Websocket without passing via HTTP headers
For the `WebSocket` browser API it is not possible to pass a JWT token as `Authorization` HTTP header as this API does not allow to set custom headers.

So there is no straight-forward way to connect to a Ditto websocket from the browser.

We should enable that by either
* allowing to open an unauthenticated WS session and passing as first message the JWT token via the WS
* or by using the sub-protocol field of the WS (for which there is an API in the browser) in order to transport the JWT token

I would prefer the first one as we already do the JWT token refresh also via a WS message.

dguggemos:
This PR fixes #667.

Due to the limitations described in #667 this PR allows to provide the JWT via the query parameter `access_token`. The parameter is only used for WebSocket requests. The existing means to specify a token (via the `Authorization` header) should be preferred whenever possible because of the security considerations mentioned in the [OAuth 2.0 specification](https://tools.ietf.org/html/rfc6750#section-2.3).


thjaeckle (approved):
LGTM 👍

PR #1042, opened 2021-04-28
Ensure that TimeMeasuringFlow does not break the parallelism of the measured flow

Yannic92:


jufickel-b (changes requested):
TimeMeasuringFlowTest#timeMeasuringInaccuracyIsLessThanFiveMS fails and need to be fixed. The durations average cannot be determined.

Yannic92:
@jufickel-b It should be solved now.

jufickel-b (approved):
LGTM

PR #1045, opened 2021-04-29
Refactor SnaphsotTaken events to be reusable

DerSchwilk:
Added intermediate class `SnapshotTaken` to make events of that kind reusable.
Add missing parts of events to `SnapshotTaken` (revision, metadata and timestamp).
Add method to `DittoHeaderInvalidException` to get the specific header that was invalid.

### DIFF
--- a/services/models/things/src/main/java/org/eclipse/ditto/services/models/things/ThingSnapshotTaken.java
+++ b/services/models/things/src/main/java/org/eclipse/ditto/services/models/things/ThingSnapshotTaken.java
@@ -105,16 +105,16 @@ public final class ThingSnapshotTaken extends SnapshotTaken<ThingSnapshotTaken> 
     }
 
     /**
-     * Deserializes a {@code ThingSnapshotTaken} instance from the specified JSON object.
+     * Deserializes a {@code DittoThingSnapshotTaken} instance from the specified JSON object.
      *
      * @param jsonObject the JSON object that should be deserialized.
      * @param dittoHeaders the headers of the deserialized event.
-     * @return the deserialized ThingSnapshotTaken instance.
+     * @return the deserialized BoschThingSnapshotTaken instance.
      * @throws NullPointerException if any argument is {@code null}.
      * @throws org.eclipse.ditto.json.JsonMissingFieldException – if {@code jsonObject} did not contain all required
      * fields.
      * @throws org.eclipse.ditto.json.JsonParseException if {@code jsonObject} does not represent a valid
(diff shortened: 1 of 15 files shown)


services/models/things/src/main/java/org/eclipse/ditto/services/models/things/ThingSnapshotTaken.java
```diff
@@ -105,26 +105,26 @@ public static Builder newBuilder(final ThingId thingId,
     }
 
     /**
-     * Deserializes a {@code ThingSnapshotTaken} instance from the specified JSON object.
+     * Deserializes a {@code DittoThingSnapshotTaken} instance from the specified JSON object.
      *
      * @param jsonObject the JSON object that should be deserialized.
      * @param dittoHeaders the headers of the deserialized event.
-     * @return the deserialized ThingSnapshotTaken instance.
+     * @return the deserialized BoschThingSnapshotTaken instance.
```
jufickel-b, line 112: Should be `ThingSnapshotTaken` instead of `BoschThingSnapshotTaken`.

signals/events/base/src/main/java/org/eclipse/ditto/signals/events/base/EventsourcedEvent.java
```diff
@@ -36,6 +36,15 @@
...
     long getRevision();
 
+    /**
+     * Sets the event's revision.
+     *
+     * @param revision the revision to be set.
+     * @return the event with the revision set.
+     * @since 2.0.0
+     */
+    T setRevision(long revision);
```
jufickel-b, line 46: Is probably unused.

jufickel-b (changes requested):
LGTM.
However, Javadoc of `ThingSnapshotTaken` needs to be fixed and `setRevision` method of `EventsourcedEvent` is unused and could probably removed (from there and all subclasses and tests etc.).

services/models/things/src/main/java/org/eclipse/ditto/services/models/things/ThingSnapshotTaken.java
DerSchwilk: Changed.

signals/events/base/src/main/java/org/eclipse/ditto/signals/events/base/EventsourcedEvent.java
DerSchwilk: Deleted.

jufickel-b (approved):

PR #1046, opened 2021-05-03
Restructure Maven Modules

Yannic92:
This PR provides part 13 of #978 

For details please check the ditto 2.0 release notes


stmaute (commented):
Maybe we should document since when these modules are marked as public API?

Yannic92:
> Maybe we should document since when these modules are marked as public API?

I thought about it, but I think git is a better tool to solve this. To find out at what time which set of modules are public API I'd suggest to look at this file for the respective version tag.
Otherwise we would have to keep all versions tracked in DADR-0005.
I don't like this. What do you think?

stmaute:
> > Maybe we should document since when these modules are marked as public API?
> 
> I thought about it, but I think git is a better tool to solve this. To find out at what time which set of modules are public API I'd suggest to look at this file for the respective version tag.
> Otherwise we would have to keep all versions tracked in DADR-0005.
> I don't like this. What do you think?

I also thought about your approach to let git handle this. But I would propose to add something like "Updated API Models for Ditto 2.0 on 04.05.2021" in the first line of DADR-0005. I do not think that we will update the Ditto API this often. And I think it makes sense to have a bit of history in architectural decisions.

thjaeckle:
I agree that we should at least provide a "changelog" to DADRs.
One can look in the Git history after finding out that the DADR was changed with Ditto 2.0

thjaeckle:
@stmaute @Yannic92 I enhanced the already existing changelog on top of the DADR with the change

thjaeckle (approved):
LGTM 👍 
Awesome refactoring @Yannic92 🎉

PR #1047, opened 2021-05-04
fixed GitHub actions (dockerfiles were moved)

thjaeckle:
* adjusted JMH -benchmark.jar generation to be in a separate profile or policies-model and connectivity-service (JavaScript benchmark)


stmaute:
LGTM 👍🏻

PR #1048, opened 2021-05-04
Extended TopicPath, removed empty TopicPath and refactored AbstractAdapter

### LINKED ISSUE #1043: Add methods to TopicPath to make its evaluation easier
There are situations where a `TopicPath` of a Ditto Protocol Message should be evaluated.
For example, if one wants to know whether the group is "Things" or the its action is "create", the appropriate getter methods have to be called.
The result of those getters is sometimes an `Optional` which puts another indirection to the evaluation.

To make evaluation of a `TopicPath` easier, methods should be added which return `boolean` like for example
```
public boolean isGroup(final Group expectedGroup) { ... }
```

jufickel-b:
Added methods to `TopicPath` to make checking some its properties easier. Fixes #1043.

Removed the unnecessary "empty TopicPath". Any Ditto Protocol message is required to have a topic. An empty TopicPath implementation has no benefit but defers the revelation of potential problems.

Had a look at `AbstractAdapter`: removed some unused methods and moved one method to `AbstractThingAdapter`.


protocol/src/main/java/org/eclipse/ditto/protocol/ProtocolFactory.java
```diff
@@ -81,117 +77,54 @@ public static Adaptable setExtra(final Adaptable existingAdaptable, final JsonOb
...
-        return ImmutableTopicPathBuilder.empty();
-    }
-
-    /**
-     * Returns a new {@code TopicPath} for the specified {@code path}.
+     * Parses the string argument as a {@link TopicPath}.
      *
-     * @param path the path.
-     * @return the builder.
+     * @param topicPathString a String containing of the {@code TopicPath} representation to be parsed.
```
DerSchwilk: "topicPathString a String containing {of} the {@code TopicPath} representation to be parsed"

DerSchwilk (approved):
LGTM

protocol/src/main/java/org/eclipse/ditto/protocol/ProtocolFactory.java
DerSchwilk: https://github.com/eclipse/ditto/pull/1048/commits/3d9428e1802956307b7133a32241b21301f8b9e9

PR #1049, opened 2021-05-05
last minute Ditto 2.0 fixes

thjaeckle:
* added "akka-discovery-kubernetes-api" and "akka-lease-kubernetes" dependencies to ditto-base-service so that all services depend on it
* made EmptyEvent not require the added "eventType" JSON field in order to not break old events

### DIFF
--- a/base/service/pom.xml
+++ b/base/service/pom.xml
@@ -77,6 +77,15 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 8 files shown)


stmaute:
LGTM 👍🏻

stmaute (approved):

PR #1050, opened 2021-05-05
fix MQTT publisher warnings when no MQTT consumer was present

thjaeckle:
* in combination with when "separatePublisherClient=true":
* create a dummy MQTT subscription handler in such cases which does not actually subscribe


stmaute:
LGTM 👍🏻

PR #1051, opened 2021-05-05
Filter sensitive parameter/header values before logging

dguggemos:
Replace sensitive values (credentials/token) of known query parameters and HTTP headers with `***` before logging them.


thjaeckle (approved):
LGTM 👍 
Just added that `filteredRawRequestUri` is not calculated when no DEBUG logging is active.

PR #1053, opened 2021-05-07
Restructure kubernetes deployment Readmes

stmaute:
Extract common deployment section from k3s and minikube Readmes into separate Readme.


thjaeckle (approved):
LGTM

PR #1054, opened 2021-05-07
Avoids throwing exceptions in ditto metrics 

Yannic92:
Avoids throwing exceptions in ditto metrics and return more or lessreasonable return values in combination with a WARN log instead

* The reason for this is that we don't want productive code to break
  because of errors in metrics

The alternative could be checked exceptions and handle those exceptions all over our code. Could also be a valid option, but for a first fix this is the faster way.

@jufickel-b @thjaeckle What do you think? Should we use checked exceptions for this?


thjaeckle:
I don't think that throwing checked exception at those places would lead to better error handling.
I think it's a good approach here logging WARNings like you did instead 👍

thjaeckle (approved):
LGTM

PR #1062, opened 2021-05-17
Relax MQTT validation

dguggemos:
Fixes MQTT validation which was too strict and did not allow setting source header mappings.


thjaeckle (approved):
LGTM

PR #1064, opened 2021-05-18
optimized MQTT protocol level acknowledgements

thjaeckle:
* when "reconnectForRedelivery=false", sucessful MQTT ACKs are issued even when a redelivery of the message is expected -> kept this behavior as this is the only mechanism MQTT provides in order to provide message re-deliveries
* made "cleanSession" config not to be based on "reconnectForRedelivery" value - apply a default of "false", but make the default configurable in connectivity.conf
* added a new "event-loop-threads" option configuring the used HiveMQTT client to use a fixed amount of netty event-loop threads instead of having them dynamically calculcated by default based on the processor count
* reworked the manual re-connection in order to be super-fast (should be <5ms now for local MQTT broker), don't create the complete MQTT connection again, but reconnect the already instantiated client
* made min and max reconnection delays configurable, apply exponential backoff for manual reconnects
* adjusted MQTT docs


stmaute:
LGTM 👍🏻

PR #1066, opened 2021-05-20
Add file appennder to ditto logback config in order to write logs to files

stmaute:
With this PR it is possible to configure logback to write logs into files via an env variable FILE_APPENDER set to true.


thjaeckle (approved):
LGTM 👍

PR #1069, opened 2021-05-21
Connectivity: Improve debug log for incoming messages.

yufei-cai:
- Log actor ref of response collector for message correlation.

- Add toString() method to MappingOutcome.

- Fix an IndexOutOfBoundException in BaseConsumerActor logging.


stmaute:
LGTM 👍🏻

PR #1076, opened 2021-05-31
Fix mounting of ditto log files on host

stmaute:
Extended docker-compose.yml with a volume to be able to access the ditto log files on the host system. 
Same for ditto-cluster.yaml.

### DIFF
--- a/deployment/docker/docker-compose.yml
+++ b/deployment/docker/docker-compose.yml
@@ -25,7 +25,7 @@ services:
        TZ: Europe/Berlin
 
   policies:
-    image: docker.io/eclipse/ditto-policies:${DITTO_VERSION:-latest}
+    image: thingsdev.azurecr.io/eclipse-ditto/policies:${DITTO_VERSION:-latest}
     mem_limit: 384m
     networks:
       default:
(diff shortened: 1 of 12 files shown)


deployment/docker/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
        TZ: Europe/Berlin
 
   policies:
-    image: docker.io/eclipse/ditto-policies:${DITTO_VERSION:-latest}
+    image: thingsdev.azurecr.io/eclipse-ditto/policies:${DITTO_VERSION:-latest}
```
thjaeckle, line 28: @stmaute this shouldn't be in here .. please change back

gateway/service/src/main/resources/logback.xml
```diff
@@ -49,15 +49,15 @@
                 <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
                     <level>info</level>
                 </filter>
-                <file>connectivity.log</file>
+                <file>/var/log/ditto/gateway.log</file>
```
thjaeckle, line 52: Oops, good catch 👍

thjaeckle (changes requested):
See comment - the image location changed from the official docker image to our internal development URL

deployment/docker/docker-compose.yml
stmaute: No this shouldn't be here. I will fix it. Good catch.

thjaeckle (approved):

deployment/kubernetes/deploymentFiles/ditto/ditto-cluster.yaml
```diff
@@ -435,13 +435,13 @@ spec:
         imagePullPolicy: IfNotPresent
         args: ["java", "-jar", "starter.jar"]
         # Resource can be managed by setting the following resource config.
-        resources:
-          requests:
-            cpu: "1"
-            memory: "1Gi"
```
thjaeckle, line 429: 1Gi is IMO too much for policies as request

deployment/kubernetes/deploymentFiles/ditto/ditto-cluster.yaml
```diff
@@ -233,13 +233,13 @@ spec:
...
```
thjaeckle, line 233: 512Mi should suffice

deployment/kubernetes/deploymentFiles/ditto/ditto-cluster.yaml
```diff
@@ -132,13 +132,13 @@ spec:
...
```
thjaeckle, line 134: 512Mi should suffice

thjaeckle:
LGTM 👍

PR #1079, opened 2021-06-01
Fix signal enrichment caching for ThingMerged events

thjaeckle:
The caching of the signal enrichment contained a bug which lead to outdated enriched fields as they were not correctly calculated in the `CachingSignalEnrichmentFacade`.


yufei-cai (approved):
LGTM

PR #1080, opened 2021-06-01
Enable separate configuration for minimum reconnect delay for MQTT broker disconnects

thjaeckle:
by default, 0ms were applied for those as well which might not be a good default and "DoS" MQTT brokers


stmaute (approved):
LGTM 👍🏻

PR #1084, opened 2021-06-07
Improve ditto log file mounting

stmaute:
To get access to the log files directory on the host system it is necessary to change the owner of this directory.
With this PR the log files directory will be made available on the host system.
Create ditto group with gid 1000 in docker image.

### DIFF
--- a/documentation/src/main/resources/pages/ditto/installation-operating.md
+++ b/documentation/src/main/resources/pages/ditto/installation-operating.md
@@ -177,8 +177,8 @@ Gathering logs for a running Ditto installation can be achieved by:
        * `DITTO_LOGGING_TOTAL_LOG_FILE_SIZE` (default: 1GB)
    * the format in which logging is done is "LogstashEncoder" format - that way the logfiles may easily be imported into
      an ELK stack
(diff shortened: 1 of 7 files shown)


documentation/src/main/resources/pages/ditto/installation-operating.md
```diff
@@ -177,6 +177,8 @@ Gathering logs for a running Ditto installation can be achieved by:
        * `DITTO_LOGGING_TOTAL_LOG_FILE_SIZE` (default: 1GB)
    * the format in which logging is done is "LogstashEncoder" format - that way the logfiles may easily be imported into
      an ELK stack
+   * when running Ditto in Kubernetes uncomment the initContainers part in the `ditto-cluster.yaml` to be able to
```
thjaeckle, line 180: @stmaute is this documentation still valid? I don't see a "initContainer" section in the `ditto-cluster.yaml`

stmaute: @thjaeckle you are right. I fixed the docs.

thjaeckle (approved):
LGTM 👍

PR #1085, opened 2021-06-09
fix rounding up "policyExpiryGranularity" to "0" (disabling it)

thjaeckle:
* this caused an arithmetic error before as "modulo 0" was done


stmaute (approved):
LGTM 👍🏻

PR #1089, opened 2021-06-15
Fix supervision in ConnectionPersistenceActor

### LINKED ISSUE #1088: Connection gets stuck and stops sending messages
I observed a bug, that when a `BaseClientActor` throws an exception during creation (which akka turns into a `ActorInitializationException`), the connection won't be able to send/receive messages (even for client count > 1), nor will it recover.

The reason for this behavior can be found in the supervision strategies of the used `ClusterRouterPool` (which starts the `BaseClientActor`s) and `ConnectionPersistenceActor` (which starts the pool).

The default strategy of the pool is, to escalate failures to the parent.
The default strategy of the `ConnectionPersistenceActor` is, to stop children which fail with a `ActorInitializationException`. As a result, the pool actor will be stopped, still used by the connection and never restored until the connection is either modified or restarted.

ffendt:
Fixes #1088:
Uses a supervisor strategy for the used `ClusterRouterPool`, which retries starting the `BaseClientActor` multiple times, before escalating exceptions (including `ActorInitializationException`) to the `ConnectionPersistenceActor`.
The `ConnectionPersistenceActor` will now always escalate exceptions, which will cause the `ConnectionSupervisorActor` to restart the failing `ConnectionPersistenceActor`.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/config/ConnectionConfig.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/config/ConnectionConfig.java
@@ -39,6 +39,12 @@ public interface ConnectionConfig extends WithSupervisorConfig, WithActivityChec
     Duration getClientActorAskTimeout();
 
     /**
+     * Returns how often the connection actor will retry starting a failing client actor. If exceeded, errors will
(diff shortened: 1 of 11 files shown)


thjaeckle (approved):
LGTM 👍

PR #1094, opened 2021-06-21
Improve+stabilize cluster failover / coordinated and graceful shutdown

thjaeckle:
Several improvements in this area:
* adjust timeout of coordinated shutdown phases to match the performed tasks
* stop akka management during coordinated shutdown
* stop logging as last stept of coordinated shutdown
* don't try to remove declared acks when own cluster instance terminates
* adjusted split brain resolver configuration to look at whole cluster and not only at a single role to device whether a split brain situation is there
* improved shutdown logging
* configure other shard allocation strategy which does improved rebalancing


stmaute (approved):
LGTM 👍🏻

Issue #586, opened 2020-01-02
Support consuming messages from Apache Kafka in Ditto connections


thjaeckle:
Currently Ditto can only [publish](https://www.eclipse.org/ditto/connectivity-protocol-bindings-kafka2.html) messages towards Apache Kafka, but cannot consume from Kafka.

In many architectures Apache Kafka is used at the edge (e.g. as mentioned [here](www.kai-waehner.de/blog/2020/01/01/apache-kafka-edge-computing-industrial-iot-retailing-logistics/)) which would require Device messages also be consumed via Ditto Connection "sources".

kaiwaehner:
Looking forward to this feature.

As soon as available, I will build a demo and use this is a lightweight but scalable solution for a Digital Twin architecture (relates to my Digital Twin architectures discussed here: https://www.kai-waehner.de/blog/2020/03/25/architectures-digital-twin-digital-thread-apache-kafka-iot-platforms-machine-learning/)

kaiwaehner:
Looking forward to this feature.

As soon as available, I will build a demo and use this is a lightweight but scalable solution for a Digital Twin architecture 
(relates to my [Digital Twin architectures with Apache Kafka](https://www.kai-waehner.de/blog/2020/03/25/architectures-digital-twin-digital-thread-apache-kafka-iot-platforms-machine-learning/))

memelet:
Can someone leave a hint as to where in the codebase this would be added?

thjaeckle:
@memelet that would have to be added to the following package:
https://github.com/eclipse/ditto/tree/master/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/kafka

There we need:
* a `KafkaConsumerActor` extending `BaseConsumerActor`
* this consumer must be started in `KafkaClientActor.startConsumerActors(ClientConnected)` (protected class to be overwritten)
* have to think about consumer offset management (where and how to persist offsets)

The existing consumers for other protocols should give a good example on what needs to be implemented in the consumer:
* [AmqpConsumerActor](https://github.com/eclipse/ditto/blob/master/services/connectivity/messaging/src/main/java/org/eclipse/ditto/services/connectivity/messaging/amqp/AmqpConsumerActor.java)
[...]

PR #1098, opened 2021-06-22
Support consuming messages from Apache Kafka 

dguggemos:
This PR implements consuming messages from Apache Kafka (#586) using Ditto managed connections. Please note however that all messages are consumed with "at-most-once" semantics, Ditto's acknowledgements will be implemented for Kafka sources in a follow-up issue. Publishing messages to Kafka was already implemented.


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/kafka/KafkaConsumerActor.java
```diff
@@ -0,0 +1,264 @@
...
+import akka.kafka.Subscriptions;
+import akka.kafka.javadsl.Consumer;
+import akka.stream.Materializer;
+import akka.stream.javadsl.Flow;
+import akka.stream.javadsl.RunnableGraph;
+import akka.stream.javadsl.Sink;
+import scala.util.Either;
+
+/**
+ * TODO
```
thjaeckle, line 54: TODO

PR #1099, opened 2021-06-23
Improved subject deletion announcement granularity to milli seconds

stmaute:
made it possible to round policy announcements to milli seconds; 
min granularity is now 1ms;


thjaeckle (approved):
LGTM 👍

PR #1100, opened 2021-06-24
increase connectivity updating-state-timeout from default 5s to 15s

thjaeckle:
this fixes cluster starting problems in connectivity service


stmaute (approved):
LGTM 👍🏻

PR #1103, opened 2021-06-25
Improve KafkaConsumer setup and add Ditto Header to overcome some internal restrictions

Yannic92:
This PR
* correctly configures the authentication mechanism also for kafka consumers,
* allows to configure a consumer groupId by setting the "groupId" key of the specificConfig of a kafka connection.
* changes the behaviour of the kafka publisher. Messages are no longer sequentially published (waiting for ack for first message before publishing the next). This increases the throughput.
* adds the ditto-sudo header which has boolean values. This header is meant to be used in piggyback commands to overcome some internal restrictions. (Something like "--force".)

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaders.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaders.java
@@ -191,11 +191,12 @@ public interface DittoHeaders extends Jsonifiable<JsonObject>, Map<String, Strin
      * Indicates whether this command is flagged as sudo command which should ignore some preventions.
      *
      * @return True if the command is flagged as sudo command, otherwise false.
+     * @since 2.1.0
      */
     boolean isSudo();
 
     /**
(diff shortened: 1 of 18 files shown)


base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaders.java
```diff
@@ -187,6 +187,13 @@ default DittoHeadersBuilder toBuilder() {
      */
     boolean isDryRun();
 
+    /**
+     * Indicates whether this command is flagged as sudo command which should ignore some preventions.
+     *
+     * @return True if the command is flagged as sudo command, otherwise false.
```
thjaeckle, line 193: `@since 2.1.0` is missing here

dguggemos: fixed it

dguggemos (approved):
LGTM :+1:

PR #1109, opened 2021-07-08
Prevent misconfiguration of config values

stmaute:
With this PR miss configuration of config values should be prevented.
E.g. if a duration in the config file is not in the expected range than a DittoConfigError exception is thrown when starting the Ditto service.


thjaeckle (approved):
LGTM

PR #1115, opened 2021-07-15
Extend host validation for subnets and add regex pattern for host validation

stmaute:
With this PR it is possible to restrict opening of connections to specific subnets.
Another possible way to restrict connections is to provide a regex pattern for host validation.


thjaeckle (approved):
LGTM 👍

PR #1116, opened 2021-07-16
Stabilize resiliency in concierge cache loading during rolling updates

thjaeckle:
Ditto's concierge service uses the Akka "ask" pattern in order to:
* fill its caches asynchronously
* retrieve things/policies and perform filtering based on the policy on them

During rolling updates (when e.g. a "shard region" of things is not available for some seconds), "asks" during this timeframe timed out at a static timeout, were not retried and the causing API call of that cache loading got an error "503" (temporarily not availabe).

This PR improves this situation by enabling configuring a "ask-with-retry" strategy via configuration in conciege + things-search services which make use of the "ask" pattern.

Short unavailabilities of a shard region will automatically be compensated and the caller of the API will just encounter an increased processing time instead of getting an error.


stmaute (approved):
LGTM 👍🏻

PR #1120, opened 2021-07-18
Fix spurious ConnectionNotAccessibleException during TestConnection; stabilize DefaultEvictingQueue

yufei-cai:
Fixed ConnectionNotAccessibleException received by client actors during TEST_CONNECTION due to actor ref broadcasting.

DefaultEvictingQueue had a small chance to increase in size due to polling when the queue is empty.


thjaeckle (approved):
LGTM 👍

PR #1123, opened 2021-07-20
Fix that configured persistence parallelism had no effect on search updater throughput

yufei-cai:


thjaeckle (approved):
LGTM 👍 
Awesome fix

PR #1126, opened 2021-07-21
fix AbstractSSLContextTest for JDK >= 11.0.12

thjaeckle:
OpenJDK changed a behavior don't wrapping `SocketException`s in `SSLException`s for JDK >= 11.0.12: https://bugs.openjdk.java.net/browse/JDK-8259662
Adjusted unit test to check for their common base exception `IOException`.


stmaute (approved):
LGTM 👍🏻

PR #1127, opened 2021-07-22
Improve Kafka Sources

Yannic92:
This PR:
* Improves visibility of connection issues of a kafka connection. Errors are now visible in the connection logs.
* Fixes a metric counting issue for consumed messages. Messages were counted twice.
* Allows to configure the consumer group ID via the `groupId` key in the `specificConfig` section of a Kafka connection
* Supports some special kafka record properties in header mapping:
  * The kafka topic (can be mapped from `{{header:kafka.topic}}`)
  * The timestamp of the kafka record (can be mapped from `{{header:kafka.timestamp}}`)
  * The key of the kafka record (can be mapped from `{{header:kafka.key}}`)
* Refactored the inbound payloadmapping to a streaming component which now supports backpressure to the kafka consumer. The kafka consumer throttling config now respects the number of produced ditto protocol messages after the payload mapping instead of the number of consumed kafka records.


thjaeckle (approved):
LGTM 👍

PR #1130, opened 2021-07-22
TimeMeasuringFlow: Remove backpressure; include buffer size as parameter; use in search updater stream

yufei-cai:
Reason: Not possible to remove back pressure without introducing ~250ms of inaccuracy to the timers.


DerSchwilk (changes requested):
Overall LGTM, minor change requests.
Also please await additional review by @Yannic92 before merging.

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
```diff
@@ -74,53 +77,55 @@ public void tearDown() {
...
-        final Sink<Duration, TestSubscriber.Probe<Duration>> sink = TestSink.<Duration>probe(system);
-        final Pair<TestSubscriber.Probe<Duration>, Sink<Duration, NotUsed>> sinkPair = sink.preMaterialize(system);
-        final TestSubscriber.Probe<Duration> durationProbe = sinkPair.first();
+
+        final Queue<Duration> durations = new ConcurrentLinkedQueue<>();
         new TestKit(system) {{
+            final int numberOfRepetitions = 10;
             Source.repeat("Test")
-                    .via(TimeMeasuringFlow.measureTimeOf(flowThatNeedsSomeTime, timerMock, sinkPair.second()))
+                    .take(10)
```
Yannic92, line 99: I think here we should use `numberOfRepititions`

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
```diff
@@ -74,53 +77,55 @@ public void tearDown() {
...
             for (int i = 0; i < numberOfRepetitions; i++) {
                 expectedResults.add("Test");
             }
-            durationProbe.request(numberOfRepetitions);
             sinkProbe.request(numberOfRepetitions);
             sinkProbe.expectNextN(CollectionConverters.asScala(expectedResults).toSeq());
-            final List<Duration> durations =
-                    CollectionConverters.asJava(durationProbe.expectNextN(numberOfRepetitions));
 
+            if (durations.size() < numberOfRepetitions) {
```
Yannic92, line 112: Can this case ever happen? Can't we be sure that all expected repetitions are finished, after line 110?

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
```diff
@@ -74,53 +77,55 @@ public void tearDown() {
...
+
             final double averageDurationInNanos = durations.stream()
                     .mapToLong(Duration::toNanos)
                     .average()
                     .orElseThrow();
             /*
-             * Needs to be lesser than sleep duration to verify that potential inaccuracies are not mixed up with
-             * what this test you ensure.
+             * Needs to be lesser than sleep duration to verify that potential inaccuracies are not mixed up.
+             * Allow 25% inaccuracy.
```
Yannic92, line 123: Why do we allow such a high inaccuracy? This means that this test allows 250ms of inaccuracy which seems to make this time measuring pointless.

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
```diff
@@ -129,71 +134,112 @@ public void timeMeasuresOnlyTimeOfWrappedFlow() {
...
+
+            if (durations.size() < numberOfRepetitions) {
+                Thread.sleep(sleepDuration.toMillis());
+            }
+
+            final double averageDurationInNanos = durations.stream()
+                    .mapToLong(Duration::toNanos)
+                    .average()
+                    .orElseThrow();
+            // allow 50% inaccuracy due to load
```
Yannic92, line 240: See question above. If the measured time is that inaccurate, I think it is no longer valuable.

Yannic92 (changes requested):
I really like how the time measuring flow is simplified now!
Also great that your changes fixed this "off-by-one" error in the number of executions.
What makes me skeptical is the accuracy of the measured times.
Could you explain what causes this inaccuracy?

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
yufei-cai: Sadly no, because the wiretap has no influence on the base flow. It is possible for the base flow to complete before wiretap handler gets scheduled. A countdown latch is more elegant here. I will amend.

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
yufei-cai: This test applies delay to all elements, which means time taken to process previous elements are added to subsequent timers. I will make the delay per element. It should make the measurement more accurate up to 1s. Sub-1s measurements are inaccurate because scheduling delay is still measured.

internal/utils/akka/src/test/java/org/eclipse/ditto/internal/utils/akka/controlflow/TimeMeasuringFlowTest.java
yufei-cai: Yeah maybe it is better to remove the component. A 250ms -500ms scheduling delay seems unavoidable.

Yannic92 (approved):

PR #1133, opened 2021-07-24
Reduce unnecessary error logs

yufei-cai:
- MQTT: reduced log level of disconnecting a disconnected client to
  DEBUG.

- AMQP: reduced log level of consumer initialization erros to INFO
  because it can be triggered by incorrect credentials.
  Disabled Qpid JMS Failover log.

- Pub/sub: reduced log level of ddata update failure to WARNING
  before it persists for 3 consecutive clock ticks.

- HTTP: log the entire response on DEBUG level in 1 statement.
  Payload and sensitive headers are hidden automatically.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/amqp/AmqpClientActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/amqp/AmqpClientActor.java
@@ -18,7 +18,6 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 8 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/amqp/AmqpClientActor.java
```diff
@@ -397,32 +396,6 @@ protected void cleanupFurtherResourcesOnConnectionTimeout(final BaseClientState
...
-                return Objects.equals(sender, connectConnectionHandler);
-            case DISCONNECTING:
-                return Objects.equals(sender, disconnectConnectionHandler);
-            case TESTING:
-            default:
-                // no need to check testConnectionHandler because test runs only once during this actor's lifetime
-                // ignore random events by default - they could come from a connection handler that is already dead
-                return false;
-        }
-    }
```
thjaeckle, line 424: @yufei-cai why did you remove this? 
If it's not required, it's sure a good thing, but I don't understand how you came to the conclusion to remove it :)

yufei-cai: The reason is that connection failures can now come from consumer actors in connected state. Instead of maintaining the "isEventUpToDate"  table for this change and all future changes I think it better to risk restarting due to stale events than to ignore relevant events.

PR #1134, opened 2021-07-27
speed up validation of message path

Yannic92:
I tested both variants with a jmh benchmark and it showed that this variant is ~3 times faster.

### DIFF
--- a/protocol/src/main/java/org/eclipse/ditto/protocol/mappingstrategies/AbstractMessageMappingStrategies.java
+++ b/protocol/src/main/java/org/eclipse/ditto/protocol/mappingstrategies/AbstractMessageMappingStrategies.java
@@ -131,12 +131,12 @@ abstract class AbstractMessageMappingStrategies<T extends Jsonifiable.WithPredic
     private static void validatePathForLiveMessages(final MessagePath path, final String messageSubject) {
         final String pathAsString = path.toString();
         final boolean valid;
-        if (pathAsString.startsWith("/features")) {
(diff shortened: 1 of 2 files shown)


protocol/src/main/java/org/eclipse/ditto/protocol/mappingstrategies/AbstractMessageMappingStrategies.java
```diff
@@ -131,12 +129,19 @@ protected static MessageHeaders messageHeadersFrom(final Adaptable adaptable) {
...
-        final Matcher matcher = pattern.matcher(path.toString());
-        if (!matcher.matches()) {
+        final String pathAsString = path.toString();
+        final boolean valid;
+        if (pathAsString.startsWith("/features")) {
+            valid = pathAsString.endsWith("/inbox/messages/" + messageSubject) ||
+                    pathAsString.endsWith("/outbox/messages/" + messageSubject);
+        } else {
+            valid = pathAsString.equals("/inbox/messages/" + messageSubject) ||
+                    pathAsString.endsWith("/outbox/messages/" + messageSubject);
```
dguggemos, line 139: Shouldn't this be `pathAsString.equals("/outbox/messages/" + messageSubject)` ?

Yannic92: Fixed.

thjaeckle (approved):
LGTM 👍

PR #1136, opened 2021-07-27
Do not publish SubjectDeletionAnnouncement for modified subjects.

yufei-cai:
Do not categorize subject modification as subject deletion in order to avoid triggering cascading policy actions.


thjaeckle (approved):
LGTM 👍

PR #1148, opened 2021-08-19
Upgrade Akka libraries

yufei-cai:


thjaeckle:
Akka 2.6.16 was released today 😅

yufei-cai:
This upgrade was for Gatling interoperability. Let's upgrade to .16 in another round if it has important changes.

PR #1156, opened 2021-08-25
Remove Java serialization

yufei-cai:
Remove all uses of Java serialization in tests and the `Serializable` interface.

Breaks the binary API, but it should not impact existing Ditto clusters because Java serialization is disabled outside of tests.

### DIFF
--- a/internal/utils/test/pom.xml
+++ b/internal/utils/test/pom.xml
@@ -27,7 +27,7 @@
     <dependencies>
         <dependency>
             <groupId>com.typesafe.akka</groupId>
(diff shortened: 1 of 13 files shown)


internal/utils/test/pom.xml
```diff
@@ -27,7 +27,7 @@
     <dependencies>
         <dependency>
             <groupId>com.typesafe.akka</groupId>
-            <artifactId>akka-actor_2.13</artifactId>
+            <artifactId>akka-actor_${scala.version}</artifactId>
```
yufei-cai, line 30: Thanks 👍

PR #1161, opened 2021-08-31
Minor fixes in connectivity to reduce error logs

Yannic92:
This PR:
* Cancels the periodic update of a connection priority when it gets deleted
* Prevents the MergeHub in front of the inboundMappingSink to log errors (They are already handled by the completion stage, when the stream terminates)
* Sends a static ResourceStatus as response for all sources and targets in case the client actor is in a failed state and not all consumer and publisher actor children could be started


thjaeckle (approved):
LGTM :+1:

PR #1164, opened 2021-09-06
round of HTTP parallelism to next factor of two

thjaeckle:
this is required as Akka >= 2.6.15 requires in "flatMapMerge" a power of two

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/httppush/DefaultHttpPushFactory.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/httppush/DefaultHttpPushFactory.java
@@ -244,10 +244,10 @@ final class DefaultHttpPushFactory implements HttpPushFactory {
     }
 
     static int parseParallelism(final Map<String, String> specificConfig) {
-        final int parsedParallelism = Optional.ofNullable(specificConfig.get(HttpPushFactory.PARALLELISM_JSON_KEY))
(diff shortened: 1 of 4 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/httppush/DefaultHttpPushFactory.java
```diff
@@ -243,10 +243,15 @@ private static ConnectionPoolSettings disambiguateByConnectionId(final ActorSyst
                         .withParserSettings(parserSettings.withHeaderValueCacheLimits(disambiguator)));
     }
 
-    private static int parseParallelism(final Map<String, String> specificConfig) {
-        return Optional.ofNullable(specificConfig.get(HttpPushFactory.PARALLELISM))
+    static int parseParallelism(final Map<String, String> specificConfig) {
+        final int parsedParallelism = Optional.ofNullable(specificConfig.get(HttpPushFactory.PARALLELISM_JSON_KEY))
                 .map(Integer::valueOf)
                 .orElse(1);
+        return determineNextPowerOfTwo(parsedParallelism);
```
jokraehe, line 250: We could write this as a mapping stage `.map(this::determineNextPowerOfTwo)`. What do you think?

thjaeckle: Good idea, sure 👍

PR #1168, opened 2021-09-07
Improve background cleanup

yufei-cai:
- Rewrite background cleanup using MongoDB async client in order to limit the number of deleted documents.
- Decentralized background cleanup to reduce cluster traffic.
- Make snapshots of deleted entities empty.


jokraehe (approved):
LGTM

PR #1171, opened 2021-09-10
Introduce QoS 1 consumption for Kafka

Yannic92:
This branch adds an implementation for a KafkaConsumerStream that consumes messages in an at-least-once semantic.
Like for our other supported protocols, we support acknowledgements to achieve e2e qos-1 from kafka to the consuming application.
Since our kafka consumer is completely driven by akka streams, we also have backpressure in the kafka consumer when the acknowledgements won't arrive as fast as new messages are available in the kafka.

To increase the maximum amount of pending acks the configuration property `ditto.connectivity.connection.kafka.consumer.throttling.max-in-flight-factor` can be set to a higher value. Default is `2.0` which means that the number of pending acks is allowed to grow twice as big as the maximum number of messages consumed per second is set to (`ditto.connectivity.connection.kafka.consumer.throttling.limit`)

When some acknowledgement fails, this will cause the consumer stream to fail and restart with an exponential backoff.
The minimum and maximum duration can be configured by:
* `ditto.connectivity.connection.kafka.consumer.restart.exponential-backoff.min` (default 100ms)
* `ditto.connectivity.connection.kafka.consumer.restart.exponential-backoff.max` (default 60s)

When the consumer is restarted, it will consume all messages since the last committed offset, which means that not only the failed message will be reprocessed, but the sequence of messages since the message that failed, which guarantees a consistent state of the twin.


thjaeckle (approved):
LGTM, Awesome work 👍

Issue #1170, opened 2021-09-09
Provide a 'fn:replace' method


ctron:
There should be an out of the box function for calling `String::replace(String,String)` using `fn:replace("from","to")`.

thjaeckle:
Sounds useful, yes 👍

thjaeckle:
Implementation hint: get inspiration e.g. from 
* [PipelineFunctionSubstringBefore](https://github.com/eclipse/ditto/blob/master/internal/models/placeholders/src/main/java/org/eclipse/ditto/internal/models/placeholders/PipelineFunctionSubstringBefore.java)
* [PipelineFunctionSubstringBeforeTest](https://github.com/eclipse/ditto/blob/master/internal/models/placeholders/src/test/java/org/eclipse/ditto/internal/models/placeholders/PipelineFunctionSubstringBeforeTest.java)

thjaeckle:
Closed via PR #1173

PR #1173, opened 2021-09-13
[#1170]: Implement fn:replace

ctron:
This change implements a function of 'fn:replace', which maps to the Java "String::replace" method.


thjaeckle (changes requested):
@ctron thanks for this contribution, awesome 👍 
Now indeed a test fails which has to do with the added function, could you please have a look?

And could you please also document the added function in `basic-placeholders.md`?

ctron:
Ah, indeed. Fixed that locally, let's see what the CI has to say.

I also added it to the docs as suggested.

thjaeckle (approved):

thjaeckle:
LGTM 👍

PR #1179, opened 2021-09-16
Fix double slashes in location header

DerSchwilk:
Currently requests ending with slash lead to duplicate slashes in location header (i.e. POST /things/ leads to /things//entityId).
Fixed by removing trailing slashes from base URI if present.


yufei-cai (approved):
LGTM. Added a unit test.

PR #1184, opened 2021-09-21
Support signal enrichment for ThingDeleted events

Yannic92:
Enriching ThingDeleted events with the last known state of the Twin.


thjaeckle (approved):
LGTM

Issue #1177, opened 2021-09-15
Additionally allow to place an URL inside a Thing / Feature definition


thjaeckle:
Currently only the following format is allowed: `"<namespace>:<name>:<version>"`

Allow to use a http / https URL in addition to the existing format. 

This is a preparation for putting links to WoT "Thing Models" in Ditto definitions as discussed in #1163

thjaeckle:
We should get this into Ditto 2.1.0 to have it usable in an official release (which we don't do that often) - should not be difficult.

PR #1185, opened 2021-09-21
allow to use an URL as "definition" for Thing + Feature definitions

thjaeckle:
in addition to the namespace:name:version

* by that, support linking to e.g. WoT Thing Models (links to JSON-LD documents) in definitions
* described in the documentation how WoT elements roughly map to Ditto concepts

Fixes: #1177


stmaute (approved):
LGTM 👍🏻

PR #1187, opened 2021-09-22
Apply e2e backpressure by using requested acks

Yannic92:
* This makes it possible to apply backpressure even from the event/message
  consuming application back down to the kafka consumer


thjaeckle (approved):
LGTM 👍

thjaeckle (approved):
added empty lines around code block to fix rendering if code block

PR #1189, opened 2021-09-22
Support for conditional updates via APIs

### LINKED ISSUE #559: Support for conditional updates via APIs
In addition to the "If-Match" and "If-None-Match" headers already [in place](https://www.eclipse.org/ditto/httpapi-concepts.html#update-write-only-if-the-resource-already-exists) add more advanced mechanism to support updates based on a condition.

The idea is to use an [RQL filter](https://www.eclipse.org/ditto/basic-rql.html) in order to specify a "condition" when doing modifying API calls (e.g. a PUT on a thing via HTTP).

For HTTP the condition could be added as query param, e.g.:
```
PUT /api/2/things/org.eclipse.ditto:my-fancy-car-1/features/temperature/properties/value?condition=ne(features/temperature/properties/value,23.9)

body: 23.9
```
In words the above would mean:
> Only update the temperature value when it was changed from the value I provide: 23.9

With a more advanced JSON structure this can also enable conditional updates based on timestamps, e.g. when choosing the following JSON structure:
```json
{
  "thingId": "org.eclipse.ditto:my-fancy-car-1",
  "policyId": "org.eclipse.ditto:my-fancy-car-1",
  "features": {
    "temperature": {
      "properties": {
        "value": 23.9,
        "timestamp": "2019-11-26T09:05:32Z"
      }
    }
  }
}
```

You are able to formulate a condition which takes the last modification "timestamp" into accout, e.g.:
```
PUT /api/2/things/org.eclipse.ditto:my-fancy-car-1/features/temperature/properties/value?condition=lt(features/temperature/properties/timestamp,"2019-11-26T09:00:32Z")

body: 24.2
```
In words the above would mean:
> Only update the temperature value when it the last reported timestamp is older than 5 minutes.

stmaute:
With this PR it is possible to use conditional updates based on the actual thing state.
It is possible to specify a condition by using  [RQL syntax](https://www.eclipse.org/ditto/basic-rql.html)

Fixes #559


documentation/src/main/resources/_posts/2021-09-16-conditional-requests.md
```diff
@@ -0,0 +1,224 @@
+---
+title: "Support conditional requests for things resources"
+published: true
+permalink: 2021-09-16-conditional-requests.html
```
thjaeckle, line 4: Could we re-date the blogpost to 22. or 23.09. ?
Also the filename. :)

stmaute: Yes - this is already on my To-Do list ;)

documentation/src/main/resources/_posts/2021-09-23-conditional-requests.md
```diff
@@ -102,7 +103,8 @@ it was **lastModified** after "2021-08-25T12:38:27".
 
 ### Permissions to execute the example
 For this example, the authorized subject could have READ and WRITE permissions on the complete thing resource.
-However, it is only necessary on the path _thing/features/water-tank/status_, because the **temperature** as well as the conditional part **lastModified** are located there.
+However, it is only necessary on the path _thing/features/water-tank/status_, because the **temperature** as well as
```
thjaeckle, line 106: Is here a colon missing? `_thing:/features/water-tank/status`

stmaute: Good catch. I will correct it.

thjaeckle: Oh, and 'properties' is missing from that resource path.

stmaute: Also fixed the resource path ;)

thjaeckle (approved):
LGTM 👍

PR #1194, opened 2021-09-27
make readConcern and readPreference to use for MongoThingsSearchUpdaterPersistence configurable

thjaeckle:
* in order to have more fine-grained options for those very often executed queries
* defaulting to default mongoDB settings of ditto search


yufei-cai (approved):
LGTM.

PR #1196, opened 2021-09-28
Use classloader of EntityId to find annotated classes

Yannic92:
This should solve issues for situations where multiple class loaders are in place.


thjaeckle (approved):
LGTM

This is a requirement for e.g. OSGi environments

PR #1200, opened 2021-09-30
Fix event publishing policy enforcement

jokraehe:
The internal ditto header "read-subjects" is not calculated correctly according to the thing's policy. A revoke in the hierarchy is not treated properly. This PR fixes that issue.


yufei-cai (approved):
LGTM

PR #1202, opened 2021-10-05
Fix that subject deletion announcement was not published when 2 subjects of the same ID were deleted

yufei-cai:


DerSchwilk (approved):
LGTM

PR #1224, opened 2021-11-03
also use batchSize() when using limit() for MongoDB aggregations

thjaeckle:


thjaeckle:
Does not have an effect on performance, therefore closing ..

PR #1226, opened 2021-11-09
Collect Apache Kafka consumer metrics

DerSchwilk:
Collect Apache Kafka consumer metrics via Alpakka and expose them to Prometheus via Kamon metrics.


connectivity/service/src/main/resources/kamon.conf
```diff
@@ -5,9 +5,4 @@ kamon {
     service = "Connectivity"
     service = ${?CONNECTIVITY_KAMON_IDENTIFIER}
   }
-  instrumentation.kafka {
```
DerSchwilk: Also good catch :+1:

PR #1229, opened 2021-11-11
Reduce defaut fetch.max.wait.ms for kafka consumers to 5 seconds

Yannic92:
* We observed issues with our consumers getting rebalanced with the 10
  seconds timeout


thjaeckle (approved):
LGTM 👍

PR #1230, opened 2021-11-11
Added support for publishing "connection logs" to fluentd/fluent bit endpoints

thjaeckle:
With this PR Ditto adds support of configuring a fluentd / fluent bit endpoint via the configuration of the `connectivity` service which enables forwarding of connection logs to a provided [fluentd](https://www.fluentd.org) or [Fluent Bit](https://fluentbit.io) endpoint.

From there, a forwarding of the connection logs to almost arbitrary log systems can be configured.

Please have a look at the added options to configure the `ditto.connectivity.monitoring.logger.publisher` accordingly to make use of this feature.


stmaute (approved):
LGTM 👍🏻

PR #1249, opened 2021-12-08
Use fully qualified container images

ctron:
Ditto uses non-qualified container images. This may be an issue, because someone might re-configure the default so something different than `docker.io`, which would give you different images.

Some distributions require you too choose in this case, which breaks the build script of ditto:

~~~
STEP 1/12: FROM adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
? Please select an image: 
  ▸ registry.fedoraproject.org/adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
    registry.access.redhat.com/adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
    docker.io/adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
    quay.io/adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
~~~

The script stops with an interactive prompt.

The safest solution is to fully qualify which images you have in mind. This PR implements this.

### DIFF
--- a/dockerfile-release
+++ b/dockerfile-release
@@ -9,7 +9,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 
-FROM adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
+FROM docker.io/adoptopenjdk/openjdk11-openj9:jdk-11.0.12_7_openj9-0.27.0-alpine-slim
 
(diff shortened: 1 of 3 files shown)


thjaeckle (approved):
LGTM 👍

PR #1250, opened 2021-12-08
Allow pushing containers too

ctron:
This change allows one to not not only, but also push images, by adding `-P` as an argument to the script.


thjaeckle (approved):
LGTM, thx 👍

PR #1252, opened 2021-12-09
re-enabled executing Junit4 tests in modules ...

thjaeckle:
which have both Junit4 and -5 tests

* fixed broken tests in modules "ditto-things-model", "ditto-things-service", "ditto-concierge-service"
* performed module refactoring of "ditto-rql-parser", pulled out "things-search" dependency to another new module "ditto-rql-search-option-parser" in order to prevent a transitive dependency from "ditto-things-model" to "ditto-things-search-model"


dguggemos (approved):
LGTM :+1:

PR #1255, opened 2021-12-15
Introduce search update observer

stmaute:
With this PR it is possible to instantiate a search update observer as akka extension which than consumes all search updates.
Added multiple update reasons to categorize search updates.


dguggemos (approved):
LGTM :+1:

Issue #106, opened 2018-01-23
Add HTTP API for "live" commands


thjaeckle:
Ditto's ["live" channel](https://www.eclipse.org/ditto/protocol-twinlive.html) is currently only available via the [WebSocket binding](https://www.eclipse.org/ditto/protocol-bindings-websocket.html).
Which means that via HTTP it is currently not possible to address a device listening on the "live" channel and answering with its live state or changing its live state.

The HTTP API for that could be really simple:
* just add a query parameter `channel=live` to all existing `/things` HTTP routes
* omitting the `channel=live` query parameter would use the default `twin` perspective (the cached value) which should still be the default as the value is the cached (persisted) one

Example: Finding out whether a "lamp" is on:

```
# retrieve the last reported value:
GET /api/2/things/org.eclipse.ditto:fancy-lamp-1/features/lamp/properties/on
returns: true

# retrieve the live value from the device itself:
GET /api/2/things/org.eclipse.ditto:fancy-lamp-1/features/lamp/properties/on?channel=live
returns: false
# someone must have manually switched it off in the meantime
```

w4tsn:
Since live commands are implemented on devices the ditto SDK (Java AND JavaScript, capital 'and' since I fear the latter is a bit under-developed feature wise :sweat_smile:) needs to provide as much guidance / abstraction as possible to ease the agent implementation. Some initial thoughts on this:

* if a command is not yet implemented for a feature on the live channel a 400 with an "unimplemented" notice should be returned. 400, because I don't know if there's any better suited status. 406 sound similar, but is semantically different.
* implementing the commands needs an API to 'plug-in' implementations
* for AMQP and MQTT connected clients I suppose (at least for mqtt) that the client will need an implementation to interact with an mqtt / amqp client ?

This might be worth another issue.

I'll additionally require a python SDK for some devices which I plan to open-source / contribute if we'll be the firsts to implement this ;)

thjaeckle:
> This might be worth another issue.

I agree 👍 
This issue is about providing the already existing "live command" concept via the HTTP API (only triggering "live commands", not answering them).

w4tsn:
I've "discovered" an interesting use-case: I'm wanting to use the ditto commands API in "live" channel, but I'd also restrict users via policy from using the "twin" channel.

[...]

PR #1259, opened 2021-12-20
Add HTTP API for "live" commands

jufickel-b:
With this PR Ditto's ["live" channel](https://www.eclipse.org/ditto/protocol-twinlive.html) is also available for commands via HTTP API.

To qualify a command as a "live" command, the header or query parameter
`channel=live` has to be applied.

Live commands bypass the twin and go directly to the devices. However, a twin is still required for policy enforcement.

Fixes: #106


thjaeckle (approved):
LGTM 👍

PR #1260, opened 2021-12-21
Make connection metric alerting connection type dependent

DerSchwilk:
In effort to activate throttling metric alerting for Kafka after mapping the metric alerting now matches the connectionType of the corresponding connection. This enables the differentiation i.e. between AMQP and Kafka throttle alerting.


jokraehe (approved):
LGTM

PR #1264, opened 2021-12-29
Fix concurrent modification of global variables in a javascript mapper

yufei-cai:
The javascript payload mapper of a connection can set the global variables of one another during parallel message mapping. It introduces nondeterminism in mapper execution and makes it harder to debug.

This patch made it so that instances of a javascript mapper running in parallel cannot modify the global variables of each other. Global variables remain persistent for performance, but their initial state in a mapper instance can come from any previous mapper instance.


DerSchwilk (approved):
LGTM

PR #1267, opened 2022-01-07
Fix reconnect backoff for Kafka connections

jokraehe:
Connections to Apache Kafka did not respect the configured backoff for authentication failures because the client actor did not wait properly for the underlying consumer and producer actors to initialise the connection.


DerSchwilk (approved):
LGTM!

PR #1271, opened 2022-01-12
Fixed/simplified using multiple MappingProcessors in inbound/outbound connectivity mapping

thjaeckle:
via construct "concatLazy(Source.lazySource())" with a simpler and more efficient operation


DerSchwilk (approved):
Nice!

PR #1272, opened 2022-01-12
Fix dead client actor refs of other client actors

Yannic92:
* Former the refreshing was started by all clients. The way it happened
  resulted in the problem that dead client actors are never removed from
  the client actor refs
* Now the refreshing is triggered by the ConnectionPersistenceActor which
  provides a SourceRef of all client actor refs (including themselfs) to
  every client actor


jokraehe (approved):
LGTM

PR #1275, opened 2022-01-17
Count connectivity messages awaiting acknowledgements

DerSchwilk:
Count the connectivity messages awaiting acknowledgements.
In effort to evaluate possible back-pressure.


jokraehe (approved):
LGTM

PR #1276, opened 2022-01-17
Allow to disable automatic passivation for specific shard regions

Yannic92:


thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/updater/actors/ShardRegionFactory.java
```diff
@@ -115,11 +115,31 @@ public ActorRef getSearchUpdaterShardRegion(final int numberOfShards,
      * @return the shard region.
      */
     public ActorRef createShardRegion(final int shards, final Props props, final String name, final String role) {
+        return createShardRegion(shards, true, props, name, role);
```
thjaeckle, line 118: Is that a good default defaulting to `true` when we know that after 2 minutes the entities will be passivated?

Yannic92: I'm not sure if its "good". It's the currenty default used for all shard regions within things search. I just wanted to allow to change it for the ones we think could be problematic.

DerSchwilk (approved):

PR #1277, opened 2022-01-17
Allow * as a placeholder for feature id in selected fields

dguggemos:
The introduced syntax can be used in http requests (`fields` parameter) and signal enrichment (`extra` fields).


stmaute (approved):
LGTM 👍🏻

Issue #1273, opened 2022-01-14
Target-only Kafka connection remains "open" even if Kafka broker is not available.


jufickel-b:
A Kafka connection without sources does no regular poll and thus cannot detect whether the broker is available.

This means a Kafka connection with only targets remains open even if Kafka broker is not available.
Such a connection should be marked as "misconfigured".
As it is not possible to detect this state pre-emptively it should be detected at least with the first sent message that times out.

PR #1278, opened 2022-01-18
Deal with unavailable Kafka broker in publisher

jufickel-b:
Until now a Ditto Kafka connection with targets only did not cope properly with the Kafka broker becoming unavailable.
In such cases publishing timed out and produced appropriate entries in connection log.
Connection live status however, remained always 'open'.
With this commit, a `TimeoutException` while publishing is handled in a way that a `ConnectionFailure` is propagated to parent `BaseClientActor`.
This sets connection live status to 'misconfigured' and triggers reconnect with back-off semantic.
Thus, a potentially unavailable Kafka broker gets appropriately reflected by connection status.

Fixes: #1273


stmaute (approved):
LGTM 👍🏻

stmaute:
This PR fixes #1273.

PR #1279, opened 2022-01-18
Fix things-search inconsistencies in search index on shard relocation

thjaeckle:
E.g. seen during rolling updates for very often updated things.


yufei-cai (approved):
LGTM.

PR #1280, opened 2022-01-19
Sync Ditto pubsub distributed data against the cluster state at regular intervals

yufei-cai:
Add a configurable interval for Ditto pubsub to sync its distributed data against the cluster state. This prevents errors due to cluster instability from staying in the distributed data for a long time.


thjaeckle (approved):
LGTM 👍

PR #1282, opened 2022-01-19
Make ClientActorRefs serializable ..

Yannic92:
so we can send this directly instead of sending a SourceRef

Signed-off-by: Yannic Klem <Yannic.Klem@bosch.io>

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/ClientActorRefs.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/ClientActorRefs.java
@@ -15,6 +15,7 @@ package org.eclipse.ditto.connectivity.service.messaging;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/ClientActorRefs.java
```diff
@@ -29,7 +30,7 @@
  * Collection of all client actor refs of a connection actor.
  */
 @NotThreadSafe
-public final class ClientActorRefs {
+public final class ClientActorRefs implements AkkaJacksonCborSerializable {
```
jufickel-b, line 33: We should add a unit test for serialization.
You could have a look at `JsonValueSourceRefTest` for an example. (PS: nowadays I would use `ActorSystemResource` as ClassRule.)

jufickel-b (changes requested):
Please test if serialization actually works.

jufickel-b (approved):
LGTM!

PR #1288, opened 2022-01-21
Test and fix false positives of ClusterStateSyncBehavior due to unserialized local actor refs

yufei-cai:
Before serialization, local actor refs do not include the IP of
its cluster member as a part of its address. ClusterStateSyncBehavior
should treat such addresses like the address of its own cluster
member.

Changed the unit tests to detect mishandling of unserialized addresses.


thjaeckle (approved):
LGTM 👍

PR #1292, opened 2022-01-24
fix that placeholder `time:now` did not resolve to the current time

thjaeckle:
when used in a connection's headerMapping


stmaute (approved):
LGTM 👍🏻

PR #1293, opened 2022-01-26
Enable use of entity placeholders for MQTT and HTTP connections

stmaute:
Until now it wasn't possible to use the entity placeholder in MQTT and HTTP connections. This PR fixes this.


thjaeckle (approved):
Cool, LGTM 👍

PR #1294, opened 2022-01-26
Remove reporting of inconsistent things in background sync actor

dguggemos:
The report contained false positives for things that were updated at high frequency.


stmaute (approved):
LGTM 👍🏻

PR #1295, opened 2022-01-28
fixed that JSON `null` in "correlation-id" of Ditto Protocol headers were parsed as JSON String `"null"`

thjaeckle:
* also fixed correlation-id logging in WebSocketRoute which did not respect the signal specific correlation-id but always used the "channel-correlation-id" for logging


yufei-cai (approved):
LGTM.

PR #1296, opened 2022-01-31
Reduce the likelihood of search index inconsistency due to reordering of patch updates

yufei-cai:
Reduce the likelihood of search index inconsistency due to reordering of patch updates.


thjaeckle (approved):
LGTM 👍

PR #1299, opened 2022-02-01
Ignore DittoMessageMapper for hono delivery failed notifications

DerSchwilk:
hono delivery failed notifications lead to mapping failures in connection logs, because the notifications are not in ditto-protocol format and thus the DittoMessageMapper can't map them.


stmaute:
LGTM 👍🏻

stmaute (approved):

PR #1300, opened 2022-02-01
Fix search consistency for failed patch updates

Yannic92:
isPatchUpdate would always be false

* this results in things never getting consistent in the search

Signed-off-by: Yannic Klem <Yannic.Klem@bosch.io>


thjaeckle (approved):
LGTM

PR #1301, opened 2022-02-01
Fix search sync infinite loop

Yannic92:


thjaeckle (approved):
Approved via pair programming 👍

PR #1305, opened 2022-02-03
Fixed potential race condition in LiveSignalEnforcement.handleAskTimeoutForCommand()

thjaeckle:
which threw a 503 (ThingNotAvailable) exception and caused a race condition against a "graceful timeout"
* made the handleAskTimeoutForCommand() return an Optional and proceed with a null element in the CompletionStage if the optional was empty


stmaute (approved):
LGTM 👍🏻

PR #1306, opened 2022-02-04
Document 'is-group-topic' and 'aggregate' header for piggyback commands

stmaute:
Add 'is-group-topic' header to connectivity piggyback commands.


thjaeckle (approved):
LGTM 👍

PR #1312, opened 2022-02-16
Adapt error propagation of tunnel errors 

stmaute:
Propagate all errors to the parent actor in case the SSH tunnel failed.
Adapt connecting-min-timeout to 50s because this timeout needs to be smaller than ditto.connectivity.connection.client-actor-ask-timeout in connectivity.conf which is currently 55s.


dguggemos (approved):
LGTM :+1:

PR #1316, opened 2022-02-18
Bugfix/release build

Yannic92:
There seems to be a bug in the maven-sure-fire plugin which causes that during the build neither the jar nor the "classes" directory is part of the classpath. This causes a problem when loading annotated classes by ClassIndex.
The other commits where necessary to find the place where things went wrong.

### DIFF
--- a/connectivity/service/pom.xml
+++ b/connectivity/service/pom.xml
@@ -331,22 +331,6 @@ jmh-generator-annprocess). jmh-generator-annprocess overwrites the whole META-IN
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <executions>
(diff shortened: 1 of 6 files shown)


connectivity/service/pom.xml
```diff
@@ -417,6 +401,14 @@ jmh-generator-annprocess). jmh-generator-annprocess overwrites the whole META-IN
                     </additionalOptions>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>3.0.0-M5</version>
```
thjaeckle, line 407: maven-plugin versions should be defined in the top-level pom.xml
or does this module explicitly need another maven-failsafe-plugin version than all the others?

PR #1317, opened 2022-02-18
fixed that ThingUpdater was already stopped when processing "ThingDeleted" events 

thjaeckle:
with a consistency-lag > 5 seconds
* also added IntelliJ run-configurations in .run/ folder

### DIFF
--- a/.run/ConciergeService.run.xml
+++ b/.run/ConciergeService.run.xml
@@ -12,7 +12,6 @@
   -->
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="ConciergeService" type="Application" factoryName="Application" folderName="Ditto" nameIsGenerated="true">
-    <option name="ALTERNATIVE_JRE_PATH" value="11" />
     <envs>
       <env name="ENFORCEMENT_SPECIAL_LOGGING_INSPECTED_NAMESPACES.0" value="org.eclipse.ditto" />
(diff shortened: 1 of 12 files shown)


.run/ConciergeService.run.xml
```diff
@@ -0,0 +1,33 @@
...
+  ~
+  ~ This program and the accompanying materials are made available under the
+  ~ terms of the Eclipse Public License 2.0 which is available at
+  ~ http://www.eclipse.org/legal/epl-2.0
+  ~
+  ~ SPDX-License-Identifier: EPL-2.0
+  -->
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="ConciergeService" type="Application" factoryName="Application" folderName="Ditto" nameIsGenerated="true">
+    <option name="ALTERNATIVE_JRE_PATH" value="11" />
```
Yannic92, line 15: is 11 here correct? We recently upgraded to Java 17. Was this done before the upgrade? In general: is this option required? I just checked my other run configurations and it seems it's not.

Yannic92 (changes requested):

.run/ConciergeService.run.xml
thjaeckle: You're right - that was not added intentionally.

Yannic92: I removed this option and it still works.

PR #1318, opened 2022-02-21
Adapt documentation for building ditto

Yannic92:
* Removed building with docker as this is not working correctly because of
  'docker inside docker' for our MongoDB integration tests
* Adapt versions to recently upgraded versions


thjaeckle (approved):
LGTM

PR #1319, opened 2022-02-22
Don't publish messages with failed enrichments and issue failed ack

DerSchwilk:
Adds the possibility to activate/deactivate whether messages with failed enrichments should be published.
The prior default was, that the messages were published. The new default is that these messages will be dropped and a failed acknowledgement will be issued for the target acknowledgement.
To reactivate the prior behaviour, set `CONNECTIVITY_MESSAGE_MAPPING_PUBLISH_FAILED_ENRICHMENTS` to true.


stmaute (approved):
LGTM 👍🏻

PR #1320, opened 2022-02-22
improve logging of ThingUpdater etc. 

thjaeckle:
in order to be better able to find out which changne caused a "failed patch update"
* also use "trace" logging for logging potentially huge complete "writeModels"


DerSchwilk (approved):
LGTM

PR #1323, opened 2022-02-28
Invalidate ResponseReceiver cache after a response is received.

DerSchwilk:
In order to restore behaviour, that a message with the same correlation-id as a previous message gets handled wit the same correlation-id (not a suffixed correlation-id, which we only want if 2 messages with the same correlation-id are processed at the same time). Previous the default expiry of 2 minutes leads to changed correlation-ids in the second message, even if the first message was already answered.

Additionally this sets the internal correlation-id to the outgoing adaptables. This allows the receiver of a message to answer with the correct correlation-id in the ditto headers. Prior the original correlation-id was kept, even if it was suffixed because of a conflict.


thjaeckle (approved):
LGTM 👍

PR #1325, opened 2022-03-01
Filter for incorrect element types in jsonArray of feature definitions

stmaute:
This PR fixes a serialization error and validates that only strings can be used in the feature definition array.


thjaeckle (approved):
LGTM 👍

PR #1327, opened 2022-03-02
Fix creation of tmp log files

stmaute:
Because of a bug in logback core library there are created many tmp files when the log file appender feature is used.
These files aren't deleted by logback after compressing the log files. 
This PR fixes the issue by removing the <file> property in the logback.xml of each Ditto service.

### DIFF
--- a/concierge/service/src/main/resources/logback.xml
+++ b/concierge/service/src/main/resources/logback.xml
@@ -51,10 +51,12 @@
                     <level>info</level>
                 </filter>
                 <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-                    <!-- daily rollover as default-->
-                    <fileNamePattern>${DITTO_LOGGING_FILE_NAME_PATTERN:-/var/log/ditto/concierge.log.%d{yyyy-MM-dd}.gz}</fileNamePattern>
+                    <!-- The rollover period is inferred from the fileNamePattern -->
+                    <!-- daily rollover as default -->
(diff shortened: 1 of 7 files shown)


concierge/service/src/main/resources/logback.xml
```diff
@@ -50,13 +50,15 @@
                 <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
                     <level>info</level>
                 </filter>
-                <file>/var/log/ditto/concierge.log</file>
                 <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-                    <!-- daily rollover as default-->
-                    <fileNamePattern>${DITTO_LOGGING_FILE_NAME_PATTERN:-/var/log/ditto/concierge.log.%d{yyyy-MM-dd}.gz}</fileNamePattern>
+                    <!-- The rollover period is inferred from the fileNamePattern -->
+                    <!-- daily rollover as default -->
+                    <fileNamePattern>${DITTO_LOGGING_FILE_NAME_PATTERN:-/var/log/ditto/concierge.log.%d{yyyy-MM}.gz}</fileNamePattern>
```
thjaeckle, line 56: but this is configured as "monthly rollover" - probably a typo :)

things/service/src/main/resources/logback.xml
```diff
@@ -50,13 +50,16 @@
                 <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
                     <level>info</level>
                 </filter>
-                <file>/var/log/ditto/things.log</file>
+<!--                <file>/var/log/ditto/things.log</file>-->
```
thjaeckle, line 53: this line can be deleted

thjaeckle (approved):
LGTM 👍

PR #1329, opened 2022-03-04
Respect in-flight messages for throttling alerts

DerSchwilk:
Add throttling alerts based on in-flight message gauge.
This also gives the opportunity to add other gauge based alertings.


jokraehe (approved):
LGTM

PR #1331, opened 2022-03-08
Enhance Placeholders to resolve to multiple values

### LINKED ISSUE #710: Enrichment: Support for a wildcard/placeholder identifying the changed feature in order to enrich e.g. its `definition`
I've stumbled upon the following use-case with the enrichment feature introduced in 1.1.0. I'm opening a WS connection and receive change events of many different things. The events are from different features which I don't know at the time of opening the WS connection.

Together with the state change events I also need to know the feature definition field. Right now I take the feature of an incoming message and query it's definition for each event. What I'd like to do is to request the corresponding definition of the feature the updated property belongs to.

Yannic92:
This PR enhances our placeholder implementation to resolve to multiple values instead of just one.
Furthermore it allows to make use of placeholders in the extra field declaration for both: websockets and connections.

This should fix #710


documentation/src/main/resources/pages/ditto/basic-enrichment.md
```diff
@@ -102,6 +102,16 @@ It is possible to use the wildcard operator '*' as feature ID and add a property
 extraFields=features/*/properties/unit
 ```
 
+If you however want to see a property only for the features changed within this event you could make use of placeholders.
+The following example would enrich the definition of all features that have changed within this event.
+```
+extraFields=features/{{feature:id}}/definition
```
thjaeckle, line 108: This will not work when generating the documentation.

Please use: 
```
{%raw%}extraFields=features/{{feature:id}}/definition{%endraw%}
```

documentation/src/main/resources/pages/ditto/basic-placeholders.md
```diff
@@ -11,16 +11,78 @@ Placeholders may be used at several places in Ditto where something should be re
...
+### Header Placeholder
+| Placeholder    | Description                                                                                                                                                                              |
+|----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `{%raw%}{{ header:<header-name> }}{%endraw%}`  | Value of the header with the given name. For incoming signals the headers of the used protocol are used. For outgoing values the Ditto protocol headers are used.(both case-insensitive) |
+### Topic Placeholder
+| Placeholder    | Description                                                                           |
+|----------------|---------------------------------------------------------------------------------------|
+| `topic:full`            | full [Ditto Protocol topic path](protocol-specification-topic.html)<br/>in the form `{namespace}/{entityId}/{group}/`<br/>`{channel}/{criterion}/{action-subject}`                              |
```
thjaeckle, line 51:
```suggestion
| `topic:full`            | full [Ditto Protocol topic path](protocol-specification-topic.html)<br/>in the form `{namespace}/{entityName}/{group}/`<br/>`{channel}/{criterion}/{action-subject}`                              |
```

documentation/src/main/resources/pages/ditto/basic-placeholders.md
```diff
@@ -11,16 +11,78 @@ Placeholders may be used at several places in Ditto where something should be re
...
+| Placeholder    | Description                                                                           |
+|----------------|---------------------------------------------------------------------------------------|
+| `{%raw%}{{ jwt:<claim-pointer> }}{%endraw%}` | Any standard or custom claims in the body of the authenticated JWT in JsonPointer notation - e.g., `jwt:sub` for the JWT "subject" or `jwt:extra/roles` for (potentially multiple) "roles" extracted from a JsonObject in the JWT named `extra`.<br/>JWT claims being JSON arrays are expanded (each value of the array is added) and provided [functions](#function-expressions) are performed on each array element.  |
+
+
+### Scope: Entity creation / modification
+
+Whenever creating or modifying [things](basic-thing.html) or [policies](basic-policy.html), the following placeholders
+may be used:
+* [request placeholder](basic-placeholders.html#request-placeholder)
```
thjaeckle, line 84: when adding a link within the same file, you can use:
```
[request placeholder](#request-placeholder)
```

This affects a lot of places in this file ;)

documentation/src/main/resources/pages/ditto/basic-placeholders.md
```diff
@@ -11,16 +11,78 @@ Placeholders may be used at several places in Ditto where something should be re
...
+### Connection Placeholder
+| Placeholder                            | Description                                                                                                                   |
+|----------------------------------------|-------------------------------------------------------------------------------------------------------------------------------|
+| `{%raw%}{{ connection:id }}{%endraw%}` | the ID of the connection which receives/publishes a signal |
+### Header Placeholder
+| Placeholder    | Description                                                                                                                                                                              |
+|----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `{%raw%}{{ header:<header-name> }}{%endraw%}`  | Value of the header with the given name. For incoming signals the headers of the used protocol are used. For outgoing values the Ditto protocol headers are used.(both case-insensitive) |
+### Topic Placeholder
```
thjaeckle, line 48: The "Topic Placeholders" and "Resource Placeholders" are not listed with curly braces.
I guess that is because the sections were moved from "Scope: RQL expressions ..." section - where there must not be curly braces around the placeholders.
For other scopes however there must be the curly braces.

I would suggest to add them everywhere in the tables - and e.g. mention for the new "Scope: RQL expressions ..." section to not include the curly braces.

PR #1333, opened 2022-03-09
Bump swagger-ui  and nginx versions

stmaute:
bump swagger-ui version to 4.6.1 and nginx version to 1.21-alpine


DerSchwilk (approved):
:+1:

PR #1335, opened 2022-03-11
Fixing that search ThingUpdater could be blocked when waiting for bulk write completions

thjaeckle:
The behavior change added in PR #1320 caused problems when the "bulk write complete" was not received by the ThingUpdater.
In that case it was blocked for 2 minutes, not processing further updates.

The fix removes this "behavior change" and blocking of new WriteModels in ThingUpdater again.

Instead, it applies sharding (based on the thingId hash) creating substreams in the `EnforcementFlow.create()` with a configured "bulk-shard-count" ensuring that the same things are updated by the same bulkWriter, or at least reducing probability of concurrent modifications in async bulk writes.


yufei-cai:
Closing and reopening due to actions not updating.

PR #1336, opened 2022-03-14
Add a "recoveryStatus" to the connection status

thjaeckle:
This contains one of the values:
* ongoing
* succeeded
* backOffLimitReached
* unknown

And can be used to find out whether an automatic failover is still ongoing or if the max amount of configured reconnects applying backoff was reached and that recovery is no longer happening.


stmaute (approved):
LGTM 👍🏻

PR #1340, opened 2022-03-17
Fixing that search ThingUpdater could be blocked when waiting for bulk write completions

yufei-cai:
@thjaeckle wrote:

The behavior change added in PR https://github.com/eclipse/ditto/pull/1320 caused problems when the "bulk write complete" was not received by the ThingUpdater.
In that case it was blocked for 2 minutes, not processing further updates.

The fix removes this "behavior change" and blocking of new WriteModels in ThingUpdater again.

Instead, it applies sharding (based on the thingId hash) creating substreams in the EnforcementFlow.create() with a configured "bulk-shard-count" ensuring that the same things are updated by the same bulkWriter, or at least reducing probability of concurrent modifications in async bulk writes.

Reopened #1335


.github/workflows/maven.yml
```diff
@@ -29,9 +29,9 @@ jobs:
       - name: Set up JDK 17
         uses: actions/setup-java@v2
         with:
-          distribution: 'zulu'
+          distribution: 'adopt-openj9'
```
thjaeckle, line 32: why OpenJ9 and not `temurin` which we nowadays use for our docker images?

yufei-cai: openj9 is not working. will switch to temurin.

thjaeckle (approved):
LGTM 👍

PR #1342, opened 2022-03-21
Add class to calculate a JSON merge patch between to JSON values

Yannic92:


json/src/main/java/org/eclipse/ditto/json/JsonMergePatch.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
+/**
+ * Can be used to get the diff in form of a JSON merge Patch according to
+ * <a href="https://datatracker.ietf.org/doc/html/rfc7386">RFC 7386</a> between two {@link JsonValue json values}.
+ */
```
thjaeckle, line 24: `@since 2.4.0` is missing from class' javadoc

json/src/main/java/org/eclipse/ditto/json/JsonMergePatch.java
```diff
@@ -0,0 +1,90 @@
...
+public final class JsonMergePatch {
+
```
thjaeckle, line 26: could you please add a private constructor throwing an AssertionError when invoked?

thjaeckle (changes requested):

json/src/main/java/org/eclipse/ditto/json/JsonMergePatch.java
```diff
@@ -0,0 +1,219 @@
...
+                result = value1;
+            }
+        }
+
+        return result;
+    }
+
+    private static JsonObject mergeJsonObjects(final JsonObject jsonObject1, final JsonObject jsonObject2) {
+
+        if (jsonObject1.isNull() || (jsonObject1.isNull() && jsonObject2.isNull())) {
```
thjaeckle, line 143: this could be simplified to `jsonObject1.isNull()`, or not?

thjaeckle (approved):
LGTM 👍

PR #1346, opened 2022-03-25
Make double decoding of connection usernames/passwords configurable

DerSchwilk:
Adds the possibility to configure whether connections usernames and passwords should be double decoded when connecting.
Leaves the default as enabled.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/AbstractHiveMqttClientFactory.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/AbstractHiveMqttClientFactory.java
@@ -12,7 +12,6 @@
  */
 package org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq;
 
-import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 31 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/AbstractHiveMqttClientFactory.java
```diff
@@ -56,25 +57,38 @@ protected AbstractHiveMqttClientFactory(final Supplier<SshTunnelState> tunnelSta
...
+        final Optional<String> password =
+                connection.getPassword().map(p -> doubleDecodingEnabled ? tryDecodeUriComponent(p) : p);
+        if (username.isPresent() && password.isPresent()) {
+            simpleAuth.username(username.get())
+                    .password(password.get().getBytes(StandardCharsets.UTF_8))
                     .applySimpleAuth();
         }
     }
 
+    private static String tryDecodeUriComponent(final String string) {
```
Yannic92, line 72: why was the decoding logic moved out of the model? Wouldn't it be possible to keep it in there and instead pass the boolean to the "getPassword" method? Now we have this logic duplicated at three places.

Yannic92 (changes requested):

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/AbstractHiveMqttClientFactory.java
DerSchwilk: You're right!
https://github.com/eclipse/ditto/pull/1346/commits/0417fd4b5e6a18d04fb81f7cedd0ad8a2a9f152b

Yannic92 (approved):

PR #1358, opened 2022-04-08
Add field selector for RetrievePolicy command

DerSchwilk:
Makes it possible to specify fields of a policy which should be returned in the RetrievePolicyResponse.
This also gives the opportunity to retrieve hidden fields. (i.e. _revision)


jokraehe (approved):
LGTM

PR #1365, opened 2022-04-12
adjust `DITTO_LOGGING_FILE_APPENDER` env key to be "true" so that file appender is used

thjaeckle:
* added new `DITTO_LOGGING_FILE_APPENDER_THRESHOLD` variable configuring the "level" threshold for the logging file appender


stmaute (approved):
LGTM 👍🏻

PR #1366, opened 2022-04-13
Add workaround for ssh tunnel connections to Rebex SSH server 

stmaute:
Add  possibility to activate the workaround by setting an entry in the specificConfig;

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/SshTunnelActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/SshTunnelActor.java
@@ -177,7 +177,8 @@ public final class SshTunnelActor extends AbstractActorWithTimers implements Cre
             logger.debug("SSH session connected successfully.");
             sshSession = connectFuture.getClientSession();
             sshSession.addSessionListener(new TunnelSessionListener(getSelf(), logger));
-            sshSession.addChannelListener(new TunnelChannelListener(getSelf(), logger));
+            sshSession.addChannelListener(new TunnelChannelListener(getSelf(),
+                    this.connection.getSpecificConfig().getOrDefault("initialWindowSize","0"), logger));
             sshSession.setServerKeyVerifier(serverKeyVerifier);
             sshSession.setUserAuthFactoriesNames(sshUserAuthMethod);
             connection.getSshTunnel()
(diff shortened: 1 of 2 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/SshTunnelActor.java
```diff
@@ -177,7 +177,8 @@ private void handleConnectResult(final ConnectFuture connectFuture) throws IOExc
             logger.debug("SSH session connected successfully.");
             sshSession = connectFuture.getClientSession();
             sshSession.addSessionListener(new TunnelSessionListener(getSelf(), logger));
-            sshSession.addChannelListener(new TunnelChannelListener(getSelf(), logger));
+            sshSession.addChannelListener(new TunnelChannelListener(getSelf(),
+                    this.connection.getSpecificConfig().getOrDefault("initialWindowSize","0"), logger));
```
dguggemos, line 181: Maybe we should call it e.g. "initialSshChannelWindowSize" to be more precise what is affected by the config property.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/TunnelChannelListener.java
```diff
@@ -46,14 +57,22 @@ public void channelInitialized(final Channel channel) {
 
     @Override
     public void channelOpenSuccess(final Channel channel) {
+        if (initialWindowSize > 0) {
+            // workaround for Rebex SSH server
```
dguggemos, line 61: I'm not sure if we should "blame" the specific server in the comment, as it seems so be a problem with the client not waiting for the adjust message and an ending in a deadlock. The behavior sending a zero window size seems to be common across different SSH server implementations.

dguggemos (approved):

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/SshTunnelActor.java
stmaute: Good idea. I will change the name.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/tunnel/TunnelChannelListener.java
stmaute: I have adjusted the comment for the workaround.

PR #1367, opened 2022-04-14
Fixed illegal argument exception in outbound mapping processor actor

kalinkostashki:
Added a separate check for the mapped mappedSignals variable since filterFailedEnrichments() may return an empty list


DerSchwilk (approved):
LGTM :+1:

PR #1368, opened 2022-04-14
In case of ThingDeleteModel always consider model as not outdated

Yannic92:
See java comment:

```
             * This branch is required because missed ThingDeletes will always have revision -1.
             * This also comes with a downside that it can happen that a thing delete model will come after a create
             * model in wrong order, so the thing will be removed from the index incorrectly.
             * However, in this case the thing will become eventually consistent either by the next background sync or
             * the next update.
```


yufei-cai (approved):
LGTM.

thjaeckle:
-> cherry-picked 3 commits to [release-2.4](https://github.com/eclipse/ditto/tree/release-2.4) branch

PR #1375, opened 2022-05-03
Allow non-standard http status codes in the http api

jokraehe:


stmaute (approved):
LGTM 👍🏻

PR #1377, opened 2022-05-05
Log buffer overflow on warn instead error level

stmaute:
replace deprecated method startPeriodicTimer() with startTimerAtFixedRate();


jokraehe (approved):
LGTM

PR #1381, opened 2022-05-10
Implement graceful shutdown for http publisher actor

jokraehe:


stmaute (approved):
LGTM 👍🏻

PR #1398, opened 2022-06-21
Search: Make simple-field-mappings configurable; make parsed query available to query criteria validators.

yufei-cai:


dguggemos (approved):
LGTM 👍

PR #1416, opened 2022-07-21
Eliminate Concierge service

### LINKED ISSUE #1339: Reduce amount of services
I think it would be a good idea to eliminate the two services:
* Gateway and
* Concierge

The plan that came to my mind is to move the HTTP API and also command authorization to the respective service the command is related to. So each of the remaining services will have an HTTP API and handles the authorization inside their service instead of concierge.
Remaining services would be:
* Things
* Policies (Not sure if this really needs to be a separate service. Maybe something that should be discussed)
* Things-Search
* Connectivity

Of course this should not be reimplemented for all services but instead we could introduce a library for common HTTP / Authorization  topics and only implement the specific parts in the respective service.

Benefits:
* Reduce complexity of understanding the interaction between the services
* More self contained services. Gateway for example depends right now on Things, Policies, Things-Search and Concierge. This service is basically useless without those other services. With the new apporach. We have a completely independent policy service and a completely independent connectivity service. Only Things-Search (Things) and Things (Policies) do still have dependencies on other services.
* More stable deployments: akka http does already provide a way to finish open http requests during coordinated shutdown. Right now this is a problem because a http request crosses multiple services to be finished (e.g. gateway -> concierge -> things). So when the concierge service was restarted this caused a lot of failing requests.
* Reduced overhead because of an additional container

Downsides:
* Containers probably will be "bigger", meaning they will need more memory and most likely also more CPU. (I expect lesser resource consumption overall). We could also horizontally scale the remaining services in order to reduce the amout of memory per service. I think we should analyze what fits best here.

Please feel free to add arguments for or against this change. (Of course this would require a Ditto 3.0 release)

Yannic92:
This Pull request targets what is discussed in #1339.

Resolves #1339

The concierge service will be eliminated and authorization of signals are moved to the entity services.
Furthermore we introduced a common extension mechanism based on Akka's Extensions.
All points that could make sense to be extended are converted to a DittoExtensionPoint.


thjaeckle (approved):
Reviewed in peers while doing the commits over the last months.
Approved :+1:

PR #1418, opened 2022-07-24
Explorer UI - review improvements for connections

thfries:
Hi again,

I made most of the proposed improvements from last review:
* Enable logs now more clear
* Confirm before deleting connection or thing
* Allow to edit connection before storing
* Retrieve logs, status and metrics on several events
Last issue pending is the creation of javascript mappings by templates.


thjaeckle (approved):
Looks good to me, thanks for the contribtution 👍

PR #1426, opened 2022-08-03
Add subscriber pooling to Ditto PubSub

yufei-cai:
A throughput bottleneck is the deserialization of signals published over Ditto PubSub. The reason is that all published signals to 1 cluster member were sent to a single instance-level subscriber, which prevents parallel deserialization. It puts a hard limit on throughput obtainable via vertical scaling.

This patch adds subscriber pools of a configurable size. The size must be identical in every cluster member in order not to change the content of the distributed data used in Ditto PubSub. Each instance-level subscriber starts a fixed number of "sub-subscribers" who share the duty of distributing published signals to local actors. They form the "subscriber pool" together with the parent subscriber. Each publisher selects a subscriber from the pool based on the entity IDs of the published signals so that the ordering of signals related to the same entity is preserved. The default subscriber pool size is equal to the deserialization parallelism configured for Akka.

Since the content of the distributed data stays the same, this patch does not require a clean cluster restart.


internal/utils/pubsub/src/main/java/org/eclipse/ditto/internal/utils/pubsub/actors/Subscriber.java
```diff
@@ -99,6 +76,26 @@ public static <T> Props props(final Class<T> messageClass, final PubSubTopicExtr
...
+     * @return An actor selection containing the subscriber in the pool responsible for the signal.
+     */
+    static ActorSelection chooseSubscriber(final ActorRef parentSubscriber, final PublishSignal signal,
+            final int poolSize) {
+
+        if (poolSize > 1) {
+            final int index = PubSubFactory.hashForPubSub(signal.getSignal().getEntityId()) % poolSize;
+            if (index > 0) {
+                return ActorSelection.apply(parentSubscriber, String.valueOf(index));
+            }
```
thjaeckle, line 94: @yufei-cai are you sure this is correct?
If you have `poolSize=4` - then the modulo `%` operation will also return `0`.
However, the "sub" actors are started starting with an index number of `1` - `4`.

So how I see it, when the modulo is `0`, the `parentSusbcriber` will be selected - and the "sub" subscriber with index number `4` will never be selected.

We could solve whit by adding `+1` to the `final int index` after calculating the modulo.

thjaeckle (approved):
Cool, I like the simplicity behind this 👍 
Reviewed

PR #1429, opened 2022-08-04
Also consider an event as semantic change when previous entity was deleted

Yannic92:


yufei-cai (approved):
LGTM.

PR #1430, opened 2022-08-05
Explorer UI -  add support for policies

thfries:
Hi Ditto team,

please find a new PR that adds support for policies.
Along with this, several changes were made:
* Moved the add and delete buttons away from the tables and moved Create/Update/Delete consistently on the details view. This worked out better for CRUD operations including templates and allowing to edit the key of an entity.
* Changed the templates for the connection and removed the modal. Templates are handled similar to definitions in the UI now everywhere
* Added definitions for things and allow now to create sample things instead of empty things only
* Added borders to tables and ace editors
* Added more validations - also for connections - (which is quite painful and causes a lot of code)
* Worked on de-selection on many places
* Improved initialization on changing environments

Looking forward to any feedback or comments.

Thomas


thjaeckle (changes requested):
Thanks @thfries for the policies part of the UI, great 👍 

I tested the UI a little and had some findings:
* [bug] deselecting a "Subject" or a "Resource" in the "Policies" page results in a Console error
* [typo] the UI has the labels "Ressource" and "Ressources" - however, "Resource" is written with only one "s"
* What is the "Subject Clipboard" about? I could not find out :D
* [layout] I think the "Who am I" area and the "Subject Clipboard" should not be below the "Policy" section - and next to the PolicyID input field
** as this is nothing related to this Policy ID, but something "global"
** the "Who am I" could IMO even be located somewhere next to the "Authorize" button at the top - or at the top right of the content section of the "policies" page

The layout adjustments you did are nice, e.g. the common approach for templates and the header in the things table 👍 the adjusted CRUD is also cool

thfries:
Many thanks for trying out and the feedback, @thjaeckle 

I'll take care of your findings.
Subject clipboard is meant to be used for authorization subjects from connections and also for subjects of other users if you want to grand access to others.
When creating subjects for a policy entry, you can select them either from the Whoami list or the subject clipboard. When you select a subject of a policy entry, the row in whoami and clipboard list is highlighted.
So this should enable pick and chose for subjects and validate subjects of policies

You are right, these are "global" and not part of the policy - so the layout has to be changed.
Subject clipboard does not work yet, because there is need to save global settings and it makes more value if there is a way to extract the authorization subjects from the connections. Most probably I will simply remove it for now.

thjaeckle (approved):
All mentioned issues have been resolved, thanks for the great contribution 👍

PR #1432, opened 2022-08-08
Add support for JsonWebTokens with an EC keyType

DerSchwilk:
Prior the deserialization of an Eliptic Curve JsonWebToken failed, because Ditto assumed it to be an RSA token and missed the modulus and exponent information.


stmaute (approved):
LGTM 👍🏻

PR #1437, opened 2022-08-09
Inbound cloud events

Pranshu-G:
WIP Pull request
Validates and Maps incoming Structured CloudEvents to Ditto Protocol for JSON as well as ```base64``` encoded String.
Missing Tests(I would like to know what kind of tests to add) and support for Binary CloudEvents format


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+    } else {
+      return true;
+    }
+  }
+
+  private JsonifiableAdaptable extractData(final String message)
+      throws UnsupportedEncodingException {
+    final JsonObject payloadJson = JsonFactory.newObject(message);
+    if (payloadJson.getValue("data_base64").isPresent()) {
+      String base64Data = payloadJson.getValue("data_base64").orElse(null).asString();
```
thjaeckle, line 103: please replace `.orElse(null)` with just `.get()` of the Optional API

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+      String decodedData = base64decoding(base64Data);
+
+      final JsonifiableAdaptable decodedJsonifiableAdaptable =
+          DittoJsonException.wrapJsonRuntimeException(
+              () ->
+                  ProtocolFactory.jsonifiableAdaptableFromJson(JsonFactory.newObject(decodedData)));
+      return decodedJsonifiableAdaptable;
+
+    } else {
+      String data = payloadJson.getValue("data").get().toString();
```
thjaeckle, line 113: when `"data"` is not in the JSON, this will throw an exception.
Please handle that more explicitly (e.g. as another "if" cause).

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+  @Override
+  public List<Adaptable> map(final ExternalMessage message) {
+    // extract message as String
+    final String payload = extractPayloadAsString(message);
+    if (validatePayload(payload)) {
+      try {
+        JsonifiableAdaptable adaptable = extractData(payload);
+        return singletonList(ProtocolFactory.newAdaptableBuilder(adaptable).build());
+      } catch (Throwable e) {
+         e.printStackTrace();
```
thjaeckle, line 128: Please handle the exception instead of just printing it to `STDERR`.
Build and throw a `MessageMappingFailedException` which is defined to be throwed in the interface the mapper implements.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  @Override
+  public List<Adaptable> map(final ExternalMessage message) {
+    // extract message as String
+    final String payload = extractPayloadAsString(message);
+    if (validatePayload(payload)) {
+      try {
+        JsonifiableAdaptable adaptable = extractData(payload);
```
thjaeckle, line 125: this variable can and should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+      return true;
+    }
+  }
+
+  private JsonifiableAdaptable extractData(final String message)
+      throws UnsupportedEncodingException {
+    final JsonObject payloadJson = JsonFactory.newObject(message);
+    if (payloadJson.getValue("data_base64").isPresent()) {
+      String base64Data = payloadJson.getValue("data_base64").orElse(null).asString();
+      String decodedData = base64decoding(base64Data);
```
thjaeckle, line 104: this variable can and should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
thjaeckle, line 103: this variable can and should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
thjaeckle, line 113: this variable can and should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+    final String payload = extractPayloadAsString(message);
+    if (validatePayload(payload)) {
+      try {
+        JsonifiableAdaptable adaptable = extractData(payload);
+        return singletonList(ProtocolFactory.newAdaptableBuilder(adaptable).build());
+      } catch (Throwable e) {
+         e.printStackTrace();
+      }
+    }
+    return null;
```
thjaeckle, line 131: You can't return `null` here - all our APIs are marked to return non-null values.

Return an empty list, e.g. with `return Collections.emptyList();`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @Override
+  public List<ExternalMessage> map(final Adaptable adaptable) {
+    return List.of(
+        ExternalMessageFactory.newExternalMessageBuilder(getExternalDittoHeaders(adaptable))
+            .withTopicPath(adaptable.getTopicPath())
+            .withText(getJsonString(adaptable))
+            .asResponse(isResponse(adaptable))
+            .asError(isError(adaptable))
+            .build());
```
thjaeckle, line 147: So the outbound CloudEvents mapping is not part of this PR?
I don't think that it is a good idea to split that up.

Because if we merge this PR, this functionality will be in the code - while only half of the "CloudEvents" mapping functionality is present.

Please also provide the outbound CloudEvent mapping as part of this PR.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
```
thjaeckle, line 13: Please put the copyright header above the package declaration.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+import static java.util.Collections.singletonList;
+import static org.eclipse.ditto.connectivity.service.mapping.AbstractMessageMapper.extractPayloadAsString;
+import static org.junit.Assert.assertEquals;
+
+public class CloudEventsMapperTest {
+
+  private static final ThingId THING_ID = ThingId.of("thing:id");
+  private static final ProtocolAdapter ADAPTER = DittoProtocolAdapter.newInstance();
+
+  String payload =
```
thjaeckle, line 45: Please use the Java multiline strings for defining the test JSON payloads, e.g.:
```java
String testPayload = """
          {
            "foo":"bar"
          }
          """;
```

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+      "{\"specversion\": \"1.0\", \"id\":\"3212e\", \"source\":\"http:somesite.com\",\"type\":\"com.site.com\"}";
+
+  String incompletePayload =
```
thjaeckle, line 48: Please use the Java multiline strings for defining the test JSON payloads

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+          "{ \"id\":\"3212e\", \"source\":\"http:somesite.com\",\"type\":\"com.site.com\"}";
+  String testPayload =
```
thjaeckle, line 50: Please use the Java multiline strings for defining the test JSON payloads

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+          + "              \"manufacturer\": \"Well known sensors producer\",\n"
+          + "                \"serial number\": \"100\", \n"
+          + "                \"location\": \"Ground floor\" },\n"
+          + "                \"features\": {\n"
+          + "                  \"measurements\": \n"
+          + "                   {\"properties\": \n"
+          + "                   {\"temperature\": 100,\n"
+          + "                    \"humidity\": 0}}}}}}";
+
+  String data =
```
thjaeckle, line 69: Please use the Java multiline strings for defining the test JSON payloads

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+          + "          \"manufacturer\": \"Well known sensors producer\",\n"
+          + "            \"serial number\": \"100\", \n"
+          + "            \"location\": \"Ground floor\" },\n"
+          + "            \"features\": {\n"
+          + "              \"measurements\": \n"
+          + "               {\"properties\": \n"
+          + "               {\"temperature\": 100,\n"
+          + "                \"humidity\": 0}}}}}";
+
+  String base64payload =
```
thjaeckle, line 86: Please use the Java multiline strings for defining the test JSON payloads

thjaeckle (changes requested):
Hi @Pranshu-G 
I was not sure if you are still working on the draft PR or if you are waiting for feedback.
I did a first rough review and commented my findings.

The direction of this PR is currently to only provide the "Structured Content Mode" (e.g. defined in https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/kafka-protocol-binding.md#33-structured-content-mode) of CloudEvents (where the complete CE is sent as JSON), do I understand correctly?
If that is the case, the "content-type" must be `"application/cloudevents+json"` - so that check is currently missing. 
IMO only messages with such a content-type can be handled by the mapper.

I am looking forward to see the support for the "Binary Content Mode".
When consuming via Kafka I would expect the CE headers to be extracted from the Kafka protocol headers, when consuming from AMQP 1.0 I would expect them to be extracted from the AMQP 1.0 properties

*edit: I saw that you mentioned that work on "Binary Content Mode" will come at a later step

Pranshu-G:
Thank you for the initial review. I will make the required changes.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+
+
+  static final String CE_ID = "ce-id";
+  static final String CE_TYPE = "ce-type";
+  static final String CE_SOURCE = "ce-source";
+  static final String CE_SPECVERSION = "ce-specversion";
+
+  static final String contentType = "application/cloudevents+json";
+
+  // make all of them static
```
thjaeckle, line 55: please remove that comment

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  static final String SPECVERSION = "specversion";
+  static final String ID = "id";
+  static final String SOURCE = "source";
+  static final String TYPE = "type";
+  // declare uuid in the function
+  // make static variables uppercase
```
thjaeckle, line 61: Please remove those comments

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+
+
+  static final String OUTBOUNDTYPE = "org.eclipse.ditto.outbound";
+
+  static final String OUTBOUNDSPECVERSION = "1.0";
+
+  static final String OUTBOUNDSOURCE = "org.eclipse.ditto";
+
+
+  private String base64decoding(final String base64Message) {
```
thjaeckle, line 71: please move private methods below public ones

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    byte[] messageByte = Base64.getDecoder().decode(base64Message);
+    String decodedString = new String(messageByte);
+    return decodedString;
```
thjaeckle, line 74: those 2 lines can be combined in 
```suggestion
    return new String(messageByte);
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  boolean validatePayload(String payload) {
```
thjaeckle, line 77: please move package private methods below public ones, above private methods

thjaeckle, line 77: the `payload` param should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    final JsonObject jsonObject = newObject(payload);
+    if (jsonObject.getValue(SPECVERSION).isPresent() && jsonObject.getValue(TYPE).isPresent()
+        && jsonObject.getValue(ID).isPresent() && jsonObject.getValue(SOURCE).isPresent()) {
+      return true;
+    } else {
+      return false;
+    }
```
thjaeckle, line 84: simplify to:
```suggestion
    return jsonObject.getValue(SPECVERSION).isPresent() && 
            jsonObject.getValue(TYPE).isPresent() && 
            jsonObject.getValue(ID).isPresent() && 
            jsonObject.getValue(SOURCE).isPresent();
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
thjaeckle, line 77: please choose a better method name ..
what is checked here? if the passed payload is
a) a JsonObject
b) contains CloudEvent fields expected in "structured" format

so maybe: `isStructuredCloudEvent()` ?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  private JsonifiableAdaptable extractData(final String message) throws Exception {
```
thjaeckle, line 87: remove `throws Exception` - this is too unspecific

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    final JsonObject payloadJson = newObject(message);
+    if (payloadJson.getValue("data_base64").isPresent()) {
+      final String base64Data = payloadJson.getValue("data_base64").get().asString();
```
thjaeckle, line 90: reuse a variable for the optional instead of creating it twice:
```suggestion
    final Optional<String> base64Opt = payloadJson.getValue("data_base64");
    if (base64Opt.isPresent()) {
      final String base64Data = base64Opt.get().asString();
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+      final String decodedData = base64decoding(base64Data);
+      final JsonifiableAdaptable decodedJsonifiableAdaptable = DittoJsonException.wrapJsonRuntimeException(
+          () -> ProtocolFactory.jsonifiableAdaptableFromJson(newObject(decodedData)));
+      return decodedJsonifiableAdaptable;
```
thjaeckle, line 94:
```suggestion
      return DittoJsonException.wrapJsonRuntimeException(
          () -> ProtocolFactory.jsonifiableAdaptableFromJson(newObject(decodedData)));
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+    if (payloadJson.getValue("data").isPresent()) {
+      final String data = payloadJson.getValue("data").get().toString();
```
thjaeckle, line 97: Same here - please reuse the same `Optional` variable

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+      return DittoJsonException.wrapJsonRuntimeException(
+          () -> ProtocolFactory.jsonifiableAdaptableFromJson(newObject(data)));
+    } else {
+      throw new Exception("Invalid CloudEvent");
```
thjaeckle, line 101: Don't throw a checked and unspecific `Exception` here.
Instead e.g. throw a `MessageMappingFailedException` here.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  public boolean checkHeaders(final ExternalMessage message) {
```
thjaeckle, line 105: * this method can be private
* place private methods below public ones
* the name is too unspecific - check the headers for what?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    Map<String, String> headers = message.getHeaders();
+    if (headers.get(CE_ID) == null || headers.get(CE_SOURCE) == null
+        || headers.get(CE_TYPE) == null || headers.get(CE_SPECVERSION) == null) {
+      return false;
+    } else {
+      return true;
+    }
```
thjaeckle, line 112: please use `containsKey` instead of checking for `null`.
Can also be simplified a lot to:
```suggestion
    final Map<String, String> headers = message.getHeaders();
    return message.getHeaders().containsKey(CE_ID) && 
            headers.containsKey(CE_SOURCE) && 
            headers.containsKey(CE_TYPE) && 
            headers.containsKey(CE_SPECVERSION);
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+
+  @Override
+  public List<Adaptable> map(final ExternalMessage message) {
+    // build Exception
+    MessageMappingFailedException mappingFailedException = MessageMappingFailedException.newBuilder(
+            message.findContentType().orElse("")).description("This is not a CloudEvent")
+        .dittoHeaders(DittoHeaders.of(message.getHeaders())).build();
+    if (message.findContentType().get().equals(contentType)) {
```
thjaeckle, line 122: `message.findContentType().get()` - you are accessing an `Optional` contentType with `.get()` which will throw an exception if there is no content type available.
Please fix.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        if (validatePayload(payload)) {
+          final JsonifiableAdaptable adaptable = extractData(payload);
+          final DittoHeaders headers = adaptable.getDittoHeaders();
+          return singletonList(
+              ProtocolFactory.newAdaptableBuilder(adaptable).withHeaders(headers).build());
+        } else {
+          throw mappingFailedException;
+        }
+      } catch (Exception e) {
```
thjaeckle, line 141: You should not just catch *any*  exception and replace it with another - that way the root "cause" will be lost.
Build and throw a new MessageMappingFailedException with the catched exception as `cause`.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  public List<ExternalMessage> map(final Adaptable adaptable) {
+    return List.of(ExternalMessageFactory.newExternalMessageBuilder(
+            Map.of(ExternalMessage.CONTENT_TYPE_HEADER, contentType))
+        .withTopicPath(adaptable.getTopicPath())
+        .withText(getExternalCloudEventSpecifications(adaptable))
+        .asResponse(isResponse(adaptable)).asError(isError(adaptable)).build());
+  }
+
+  private static String getExternalCloudEventSpecifications(Adaptable adaptable) {
+    final String OUTBOUNDID = UUID.randomUUID().toString();
```
thjaeckle, line 163: this is no constant, please name the variable with lowercase letters

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  @Override
+  public List<ExternalMessage> map(final Adaptable adaptable) {
+    return List.of(ExternalMessageFactory.newExternalMessageBuilder(
+            Map.of(ExternalMessage.CONTENT_TYPE_HEADER, contentType))
+        .withTopicPath(adaptable.getTopicPath())
+        .withText(getExternalCloudEventSpecifications(adaptable))
+        .asResponse(isResponse(adaptable)).asError(isError(adaptable)).build());
+  }
+
+  private static String getExternalCloudEventSpecifications(Adaptable adaptable) {
```
thjaeckle, line 162: the parameter should be `final`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    final String OUTBOUNDID = UUID.randomUUID().toString();
+    JsonObject dataObject = JsonFactory.newObject(getJsonString(adaptable));
+    Map<JsonKey, JsonValue> outboundJson = new HashMap<>() {{
```
thjaeckle, line 165: please don't initialize Maps like this, this is an anti-pattern

> Because Double Brace Initialization (DBI) creates an anonymous class with a reference to the instance of the owning object, its use can lead to memory leaks if the anonymous inner class is returned and held by other objects. Even when there’s no leak, DBI is so obscure that it’s bound to confuse most maintainers.

Instead, just use a `JsonObjectBuilder` for building a new JsonObject:
```java
final JsonObject externalMessageObject = JsonObject.newBuilder()
            .set("data", dataObject)
            .set(...)
            .build();
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+      put(JsonKey.of("data"), JsonValue.of(dataObject));
```
thjaeckle, line 166: why don't you use constants here? but at another place in the mapper you do for the same identifiers

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,187 @@
...
+      put(JsonKey.of("specversion"), JsonValue.of(OUTBOUNDSPECVERSION));
+      put(JsonKey.of("id"), JsonValue.of(OUTBOUNDID));
+      put(JsonKey.of("source"), JsonValue.of(OUTBOUNDSOURCE));
+      put(JsonKey.of("type"), JsonValue.of(OUTBOUNDTYPE));
+    }};
+    JsonObject externalMessageObject = newObject(outboundJson);
+    return externalMessageObject.toString();
+  }
+
+  private static DittoHeaders getExternalDittoHeaders(final Adaptable adaptable) {
```
thjaeckle, line 176: this method is not used - can be removed

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/InboundDispatchingSink.java
```diff
@@ -192,19 +194,20 @@ private InboundDispatchingSink(final Connection connection,
     /**
      * Creates a Sink that dispatches inbound messages after they are mapped.
      *
-     * @param connection the connection
-     * @param headerTranslator the headerTranslator to use.
-     * @param proxyActor the actor used to send signals into the ditto cluster.
-     * @param connectionActor the connection actor acting as the grandparent of this actor.
+     * @param connection                           the connection
```
thjaeckle, line 197: Please undo any changes made to this file - you only reformatted the code but did not change anything.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
thjaeckle: Still open: copyright header must be placed above the package declaration.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,186 @@
...
+import java.nio.charset.StandardCharsets;
+
+import static java.util.Collections.*;
+import static org.eclipse.ditto.connectivity.service.mapping.AbstractMessageMapper.extractPayloadAsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+public class CloudEventsMapperTest {
+
+  private static final ThingId THING_ID = ThingId.of("thing:id");
```
thjaeckle, line 45: unused constant

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final ProtocolAdapter ADAPTER = DittoProtocolAdapter.newInstance();
```
thjaeckle, line 46: unused constant

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final String cloudEventsContentType = "application/cloudevents+json";
```
thjaeckle, line 47: constants should be in capital letters

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -68,6 +68,7 @@ The following message mappers are included in the Ditto codebase:
 | [RawMessage](#rawmessage-mapper) | For outgoing message commands and responses, this mapper extracts the payload for publishing directly into the channel. For incoming messages, this mapper wraps them in a configured message command or response envelope. | ✓ | ✓ |
 | [ImplicitThingCreation](#implicitthingcreation-mapper) | This mapper handles messages for which a Thing should be created automatically based on a defined template. | ✓ |  |
 | [UpdateTwinWithLiveResponse](#updatetwinwithliveresponse-mapper) | This mapper creates a [merge Thing command](protocol-specification-things-merge.html) when an indiviudal [retrieve command](protocol-specification-things-retrieve.html) for an single Thing was received via the [live channel](protocol-twinlive.html#live) patching exactly the retrieved "live" data into the twin. | ✓ |  |
+| [CloudEvents](#CloudEvents-mapper) | Maps incoming CloudEvents to a Ditto Protocol Message. Supports both binary and structured CloudEvents.                                   | ✓ | ✓ |
```
thjaeckle, line 71: is the "outbound" mapping really ready? I don't think so and I wouldn't check the "Outbound" column yet.

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -351,6 +352,84 @@ Example configuration:
...
+            "features": {
+              "measurements":{
+                "properties":{
+                  "temperature": 100,
+                  "humidity": 0
+                  }
+                }
+              }
+            }
+         }
```
thjaeckle, line 393: Please apply JSON proper formatting

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -351,6 +352,84 @@ Example configuration:
...
+                  "features": {
+                    "measurements":{
+                      "properties":{
+                        "temperature": 100,
+                         "humidity": 0}
+                        }
+                       }
+                     }
+                    }
+                  }
```
thjaeckle, line 431: Please apply JSON proper formatting

thjaeckle (changes requested):
Did another rough review and commented inline ...

By the way: are you aware of the CloudEvents Java SDK and [Event Formats](https://cloudevents.github.io/sdk-java/core.html#using-event-formats)?
I was somehow expecting that you would use this API in order to parse/create CloudEvents and not do everything "by foot".

The Java SDK also contains a lot of constants which you now duplicated. This should at least be consolidated with the Java SDK (which is already a dependency in the "gateway" service of Ditto).

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
Pranshu-G: To handle this, would adding a catch block for ```NoSuchElementException``` resolve it?

thjaeckle: No, that would be a really bad code style.

thjaeckle: That would throw a `NullPointerException` when then calling `.equals(contentType)` on it.
A simple solution would just be:
```
if (message.findContentType().orElse("").equals(contentType))
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+  private String base64decoding(final String base64Message) {
+    byte[] messageByte = Base64.getDecoder().decode(base64Message);
+    return new String(messageByte);
+  }
+
+  private static String getExternalCloudEventSpecifications(final Adaptable adaptable) {
+    final String outboundID = UUID.randomUUID().toString();
+    JsonObject dataObject = JsonFactory.newObject(getJsonString(adaptable));
+    final JsonObject externalMessageObject = JsonObject.newBuilder()
+        .set("data", dataObject)
```
thjaeckle, line 152: `"data"` should become a constant of this class

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+  private JsonifiableAdaptable extractData(final String message) {
+    final JsonObject payloadJson = newObject(message);
+    final Optional<JsonValue> base64Opt = payloadJson.getValue("data_base64");
+    if (base64Opt.isPresent()) {
+      final String base64Data = base64Opt.get().asString();
+      final String decodedData = base64decoding(base64Data);
+      return DittoJsonException.wrapJsonRuntimeException(
+          () -> ProtocolFactory.jsonifiableAdaptableFromJson(newObject(decodedData)));
+    }
+    final Optional<JsonValue> dataOpt = payloadJson.getValue("data");
```
thjaeckle, line 132: `"data"` should become a constant of this class

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+    final Map<String, String> headers = message.getHeaders();
+    return message.getHeaders().containsKey(CE_ID) &&
+        headers.containsKey(CE_SOURCE) &&
+        headers.containsKey(CE_TYPE) &&
+        headers.containsKey(CE_SPECVERSION);
+  }
+
+  private JsonifiableAdaptable extractData(final String message) {
+    final JsonObject payloadJson = newObject(message);
+    final Optional<JsonValue> base64Opt = payloadJson.getValue("data_base64");
```
thjaeckle, line 125: `"data_base64"` should become a constant of this class

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.*;
+import java.nio.charset.StandardCharsets;
+
+import static java.util.Collections.*;
+import static org.eclipse.ditto.connectivity.service.mapping.AbstractMessageMapper.extractPayloadAsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
```
thjaeckle, line 42: Please let your IDE optimize the imports - a lot of them are not used at the moment.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapperTest.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.ditto.connectivity.service.mapping;
+import org.apache.kafka.common.protocol.types.Field;
```
thjaeckle, line 15: add blank line between `package` and the first `import`

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -350,6 +351,87 @@ Example configuration:
...
+Incoming messages need to have the mandatory CloudEvents fields.
+For example, a Binary CloudEvent for Ditto would look like this:
+
+```
+  headers:
+      ce-specversion:1.0
+      ce-id:some-id
+      ce-type:some-type
+      ce-source:generic-producer
+      content-type:application/cloudevents+json
```
thjaeckle, line 371: I am unsure if this is the correct content-type for "binary CloudEvents" ..
Because the payload is cleary DittoProtocol JSON, so content-type should be `"application/vnd.eclipse.ditto+json"`.

But then the CloudEventsMapper would not process the message, or would it.
What do you think @ctron ?

Shouldn't `content-type:application/vnd.eclipse.ditto+json` together with all required "binary CloudEvent" headers also be handled by the CloudEventsMapper?
Currently it isn't ..

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -350,6 +351,87 @@ Example configuration:
...
+          "temperature": 100,
+          "humidity": 0
+        }
+      }
+    }
+  }
+}
+```
+
+A Structured CloudEvent for ditto would look like this:
```
thjaeckle, line 398:
```suggestion
A Structured CloudEvent for Ditto would look like this:
```

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -350,6 +351,87 @@ Example configuration:
   }
 }
 ```
+### CloudEvents Mapper
+
+This mapper maps incoming [CloudEvent](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md) to Ditto Protocol. It provides support for both Binary CloudEvents as well as Structured CloudEvents.
+
+**Note**: The mapper only supports incoming messages with `content-type:application/cloudevents+json`. Messages that do not conform to this are ignored.
+
+#### Specifications
```
thjaeckle, line 360: That's a bad heading - you provide examples below this heading and not "Specifications".
```suggestion
#### CloudEvents examples
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+  static final String OUTBOUNDSPECVERSION = "1.0";
+  static final String OUTBOUNDSOURCE = "org.eclipse.ditto";
+
+
+  @Override
+  public List<Adaptable> map(final ExternalMessage message) {
+    MessageMappingFailedException mappingFailedException = MessageMappingFailedException.newBuilder(
+            message.findContentType().orElse("")).description("This is not a CloudEvent")
+        .dittoHeaders(DittoHeaders.of(message.getHeaders())).build();
+    if (message.findContentType().orElse("").equals(CONTENT_TYPE)) {
```
thjaeckle, line 67: Must a "binary CloudEvent" really contain the CloudEvent contentType?
As I understood the "binary CloudEvent" the CloudEvent attributes are read from headers whereas the payload does not have anything to do with a CloudEvent.
Or am I wrong?

thjaeckle (changes requested):
Next round of review.

One thing I think still is wrong from a conceptual point of view is that the added `CloudEvents` mapper only handles messages containing the Content-Type `"application/cloudevents+json"`.

But for binary structured CloudEvents the payload is in Ditto Protocol.

So FMPOV the CloudEventsMapper should:
* support either `"application/cloudevents+json"` content-type and structured CloudEvent (CE attributes being in the JSON payload)
* additionally support `"application/vnd.eclipse.ditto+json"` content-type **if** in addition the mandatory binary CloudEvent headers are also available for the processed message

Another thing which comes to my mind and is not yet addressed:
* how does the CloudEventsMapper choose for "outbound" CloudEvents whether to publish them as binary or as structured CloudEvent?
    * as far as I can see, always structured CloudEvents are published
    * FMPOV this would be a configuration of the mapper, being either:
        * structured
        * binary

@ctron what do you think about those remaining issues?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+import org.eclipse.ditto.connectivity.model.ConnectivityModelFactory;
+import org.eclipse.ditto.connectivity.model.MappingContext;
+import org.eclipse.ditto.connectivity.model.MessageMappingFailedException;
+import org.eclipse.ditto.internal.utils.akka.logging.DittoLogger;
+import org.eclipse.ditto.internal.utils.akka.logging.DittoLoggerFactory;
+import org.eclipse.ditto.json.*;
+import org.eclipse.ditto.messages.model.MessageHeaders;
+import org.eclipse.ditto.protocol.Adaptable;
+import org.eclipse.ditto.protocol.JsonifiableAdaptable;
+import org.eclipse.ditto.protocol.ProtocolFactory;
```
thjaeckle, line 37: Please organize imports - many of them are unused.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
ctron: Yes, a cloud event always has a content-type:
* https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#321-content-type
* https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#331-kafka-content-type

In the case of a binary cloud event, the content type is mapped to the "data content type", in the structural cloud event is must have a value of `application/cloudevents+json; charset=UTF-8` (or `application/cloudevents+json`).

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
ctron: Right, in this case the content type is the content type of the data which is equal to the payload in the case of a binary encoded event. So it should be `application/vnd.eclipse.ditto+json`.

From what I see, the takes the "ditto headers", which to my understanding are the incoming headers, and passes them along for further processing.

ctron:
@thjaeckle you are right that the outbound mapper made a decision on how to send out event. however, it also is up to the sender to decide. e.g. see: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/bindings/kafka-protocol-binding.md#3-kafka-message-mapping

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @Override
+  public List<Adaptable> map(final ExternalMessage message) {
+    final String payload = extractPayloadAsString(message);
+    final String contentType = message.findContentType().orElse("");
+    if (contentType.equals(BINARY_CONTENT_TYPE)) {
+      if (isBinaryCloudEvent(message)) {
+        final JsonifiableAdaptable binaryAdaptable = DittoJsonException.wrapJsonRuntimeException(
+            () -> ProtocolFactory.jsonifiableAdaptableFromJson(newObject(payload)));
+        DittoHeaders headers = binaryAdaptable.getDittoHeaders();
```
thjaeckle, line 69: I think we should extract the `"id"` of the incoming CE and use it as `correlation-id` of the DittoHeaders, e.g.:
```suggestion
        final DittoHeaders headers = binaryAdaptable.getDittoHeaders()
                .toBuilder()
                .correlationId(message.getHeaders().get(CE_ID))
                .build();
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+            ProtocolFactory.newAdaptableBuilder(binaryAdaptable).withHeaders(headers).build());
+      } else {
+        throw MessageMappingFailedException.newBuilder(
+                message.findContentType().orElse("")).description("This is not a CloudEvent")
+            .dittoHeaders(DittoHeaders.of(message.getHeaders())).build();
+      }
+    } else if (contentType.equals(STRUCTURED_CONTENT_TYPE)) {
+      if (isStructuredCloudEvent(payload)) {
+        final JsonifiableAdaptable adaptable = extractData(payload);
+        final DittoHeaders headers = adaptable.getDittoHeaders();
```
thjaeckle, line 80: Same as above - also for "structured CEs" the "id" of the CE should be added as the DittoHeaders' `correlation-id` - must be extracted from the CE JSON accordingly.

documentation/src/main/resources/pages/ditto/connectivity-mapping.md
```diff
@@ -68,6 +68,7 @@ The following message mappers are included in the Ditto codebase:
 | [RawMessage](#rawmessage-mapper) | For outgoing message commands and responses, this mapper extracts the payload for publishing directly into the channel. For incoming messages, this mapper wraps them in a configured message command or response envelope. | ✓ | ✓ |
 | [ImplicitThingCreation](#implicitthingcreation-mapper) | This mapper handles messages for which a Thing should be created automatically based on a defined template. | ✓ |  |
 | [UpdateTwinWithLiveResponse](#updatetwinwithliveresponse-mapper) | This mapper creates a [merge Thing command](protocol-specification-things-merge.html) when an indiviudal [retrieve command](protocol-specification-things-retrieve.html) for an single Thing was received via the [live channel](protocol-twinlive.html#live) patching exactly the retrieved "live" data into the twin. | ✓ |  |
+| [CloudEvents Mapper](#cloudevents-mapper) | The mapper maps incoming CloudEvent to Ditto Protocol. Supports both Binary and Structured CloudEvent. | ✓ | |
```
thjaeckle, line 71: I think now we can check also the "outbound" column again - everything should be in place.

thjaeckle (changes requested):
Hi @Pranshu-G 
We are almost done, I promise :)

Everything looks in good shape, I however had a look again at the CloudEvents spec and the defined attributes and suggest to add some additional attributes for the "outbound" event mapping.
And to also extract the "id" of inbound CloudEvents and pass it along as "correlation-id" in the `DittoHeaders`.
That way you could correlate an inbound CE which is processed by Ditto to an outbound Event sent as CE - both identified with the same preserved "id".

You could also rename the PR - as you now include the "full" support of CE mapping and not only inbound.

Afterwards we should have green light to merge 👍

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/CloudEventsMapper.java
```diff
@@ -0,0 +1,166 @@
...
+  static final String CE_SPECVERSION = "ce-specversion";
+  static final String STRUCTURED_CONTENT_TYPE = "application/cloudevents+json";
+  static final String BINARY_CONTENT_TYPE = "application/vnd.eclipse.ditto+json";
+  static final String SPECVERSION = "specversion";
+  static final String ID = "id";
+  static final String SOURCE = "source";
+  static final String TYPE = "type";
+  static final String OUTBOUNDTYPE = "org.eclipse.ditto.outbound";
+  static final String OUTBOUNDSPECVERSION = "1.0";
+  static final String OUTBOUNDSOURCE = "org.eclipse.ditto";
```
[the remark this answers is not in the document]
ctron: Technically, it doesn't need to be a URI, but a URI-reference, which `org.eclipse.ditto` would be.

But also the recommendation is to use an absolute URI. So this change would make sense indeed.

Pranshu-G:
Hi @thjaeckle 
Thank you for the review. I have pushed the necessary changes.
Let me know what you think.

thjaeckle (approved):
Thank you @Pranshu-G for staying on the topic during all the PR reviews.

FMPOV this now looks good to be merged and definitely adds a great additional value to Eclipse Ditto.
Thank you and thanks to @ctron for this addition to our project 👍

PR #1440, opened 2022-08-10
re-enable enrichment of ThingDeleted events

thjaeckle:
* to do that, also retrieve "policyId" in ThingEnforcerActor for already "marked as  deleted" things using the existing DittoHeader "DITTO_RETRIEVE_DELETED"


stmaute (approved):
LGTM 👍🏻

PR #1447, opened 2022-08-18
Fix edge command forwarder ordering

thjaeckle:
After #1416 we found a bug that the order of commands might be mixed up in `EdgeCommandForwarderActor` when applying the `SignalTransformer` in order to modify passed through signals.

As this is a regression (Ditto always ensures that for a single device/thing) all commands are processed in the same order they "arrive", this PR fixes the bug.

### DIFF
--- a/edge/service/src/main/java/org/eclipse/ditto/edge/service/dispatching/EdgeCommandForwarderActor.java
+++ b/edge/service/src/main/java/org/eclipse/ditto/edge/service/dispatching/EdgeCommandForwarderActor.java
@@ -17,6 +17,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 
+import org.eclipse.ditto.base.model.entity.id.EntityId;
 import org.eclipse.ditto.base.model.entity.id.WithEntityId;
 import org.eclipse.ditto.base.model.exceptions.DittoInternalErrorException;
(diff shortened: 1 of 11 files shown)


edge/service/src/main/java/org/eclipse/ditto/edge/service/dispatching/EdgeCommandForwarderActor.java
```diff
@@ -151,25 +163,53 @@ public Receive createReceive() {
...
+                askWithRetryCommandForwarder.forwardCommand(command,
+                        shardRegions.things(),
+                        sender);
+            } else {
+                shardRegions.things().tell(transformed, sender);
+            }
+        }));
+    }
+
+    private void scheduleTask(final Signal<?> signal,
```
kalinkostashki, line 188: maybe rename signal var to thingSignal so readability is improved.

thjaeckle: Not only thing signals are passed into that method, but also policy signals, connectivity commands, etc.

kalinkostashki:
Looks good to me :)

kalinkostashki (approved):
LGTM :)

Yannic92:
This caused some bugs revealed in internal testing library. Reverting this with: https://github.com/eclipse/ditto/pull/1453

PR #1450, opened 2022-08-19
renamed search cluster role to just "search"

thjaeckle:
* adjusted Ditto sandbox to not start connectivity
* start ditto-ui in Ditto sandbox compose


yufei-cai (approved):
LGTM.

PR #1451, opened 2022-08-19
don't log an ERROR log for automatically issued ackList

thjaeckle:
* when the original signal did not request any acknowledements
* only log a INFO log instead


kalinkostashki (approved):
LGTM

PR #1452, opened 2022-08-19
Fix Retrieval of Connection IDS via tag

Yannic92:
This PR fixes the retrieval of Connection IDs via tag. Previously the mongo query "unfold" step aborted, when a complete batch of journal entries was returned, that did not match the tag.

Restructrue acknowledgement aggregation:
* Split Policy Announcement ack aggregation and Edge command ack aggregation
* Move logic for edge command ack aggregation to edge service module and eliminate ditto-internal-model-acks


thjaeckle (approved):
LGTM 👍

PR #1454, opened 2022-08-19
Fix edge command forwarder ordering

Yannic92:
This contains again the changes from https://github.com/eclipse/ditto/pull/1447 plus fixes for issues that were noticed after merging


thjaeckle (approved):
LGTM 👍

PR #1455, opened 2022-08-19
Explorer UI - Improve message to feature and some WoT support

thfries:
Hello again,
please find this PR with some improvements for messages and initial support for WoT 😀
* Ace editors instead of textareas for message payload and response
* Extracted message parts to separate files
* Allow to save message parameters and choose them from a dropdown
* Fixed layout for environment json ace editor
* Fixed deselection in things table
* Support to load WoT TD for things and features.

Still have some things in my mind but they are not finalized:
* Would like to load the actions from TD and pre-fill the message parameters in some way (e.g. fill the description of "input" in the payload editor or set timeout = 0 if there is no "output" specified?)
* May be there should be at least some pre-check if the definition is a valid URL before trying to read the TD?
* Create a thing with a TM url works but some way to pre-fill the {'definition': 'myTMURL'} into the Thing JSON would be nice.

Remark to the code:
There was for the first time the need to have two instances of the same module, because the descriptions of a thing and a feature are very similar. This was solved by the module pattern so that the structure of the new module now deviates from all the other modules. ...still no use of classes after reading some discussions on that topic.


ui/modules/things/wotDescription.js
```diff
@@ -0,0 +1,84 @@
...
+  let aceWoTDescription;
+  let viewDirty = false;
+  const _forFeature = forFeature;
+  const domTheFeatureId = document.getElementById('theFeatureId');
+
+  const ready = async () => {
+    const tabId = Utils.addTab(
+        document.getElementById(targetTab.itemsId),
+        document.getElementById(targetTab.contentId),
+        'Description',
```
thjaeckle, line 30: I would rather name the tab either:
* WoT TD
* WoT Thing Description
* WoT Description

because it is no "Ditto term".
We could also provide more text as tooltip on the tab I suppose.

thjaeckle (changes requested):
Hi @thfries 

I had a look at the suggested improvements on message templates.
I must admit that I did not get it working - I tried adding a new template, but the template "dropdown" did not show it afterwards.
However I saw it saved in the environment as:
```json
"messageTemplates": {
    "": {
      "fafa": {
        "subject": "foobar",
        "timeout": "10",
        "payload": {
          "foo": "bar"
        }
      },
      "Saved message templates": {
        "subject": "foobar",
        "timeout": "10",
        "payload": {
          "foo": "bar"
        }
      }
    }
  }
```

Maybe the "empty string" as key is a problem here?

I would also like to suggest to not use the "star" icon for saving a new template - this is inconsistent to the other "save operations" ("add new" being the plus icon and the "save" icon for saving an existing one) in the UI.

For the WoT "Description" tab I would suggest to more clearly name it - as "Description" is no term used in Ditto.
Maybe also add tooltips to both tabs and buttons (for buttons that would improve what the "creating" vs. "saving" is).

ui/modules/things/wotDescription.js
thfries: Decided for "WoT TD" because it takes less space and is similar to "CRUD".
Feel free to change the tooltip.
(would be good to later rework all tooltips with bootstrap tooltips that would allow links and formatting)

thfries:
Hi @thjaeckle,
thanks for the feedback and sorry for the late response.
I resolved the issues you discovered.
I am open to change the fav icon but please let me give my thoughts:
* fav icon is also used to save rql queries for search
* the usage should be a bit more leightweight (just mark your entries after they worked for later re-use). Otherwise we need 3 buttons instead of one.
* no ditto objects are manipulated (I know the field definitions are also no ditto object so you re right ist is not consistent)
Would be good to have more opinions - I am not sure here.

thjaeckle (approved):
@thfries ah, now I understand how the message templates intend to be used.

It is an easy and good start to ease sending the same messages again and again to the same kind of things.
So as a first step it's ok :+1:

It is however a little bit strange that the message template is stored based on the "featureId".
E.g. when you have different features accepting the same kind of messages but with different IDs you would have to create a message template for each different feature ID.

Regarding the icons I think that adding tooltips might already help a lot. Then re-using an icon for "similar" actions would be ok.

I would approve the PR, as it is an improvement.
The icons or tooltips can be done separately.

thjaeckle:
@thfries is this PR ready to be merged from your side?

thfries:
yes, no further changes. I just intended to use the new feature during daily work.
One open question was if the message templates should be based on the feature ID. It should be either feature definition based or just globally.
But I can be merged from my point of view. Thanks!

PR #1456, opened 2022-08-22
Feature/enable self signed certificates for kafka

kalinkostashki:
Enables support for self-signed certificates.

 - motivation behind this change is to enable eclipse/packages cloud2edge helm chart to set up a hono and ditto to kafka using self signed certs.


thjaeckle (approved):
LGTM 👍

PR #1457, opened 2022-08-22
bugfixes regarding merge patch with empty path and entity ID validation for acks

Yannic92:


thjaeckle (approved):
reviewed during pair programming

PR #1462, opened 2022-08-26
Eliminate race condition with CommandTimeoutException and preserve headers of CommandTimeoutException

Yannic92:


thjaeckle (approved):
LGTM 👍

PR #1464, opened 2022-08-28
Explorer UI - Fix: Avoid storing credentials

thfries:
Hello @thjaeckle,

as discussed, please find a fix to avoid that credentials are stored.

Thomas


thjaeckle (approved):
Thanks @thfries - that should work just fine that way 👍 
I reviewed and ensured that no credentials other than the "default" ones from the environment are added to the local browser storage.

PR #1465, opened 2022-08-30
Make it possible to provide multiple OIDC issuer urls for a single configured openid-connect "prefix"

thjaeckle:
Up to now exactly one "issuer" URL could be configured.
Now, while maintaining backwards compatibility, also a list of "issuers" may be configured. If the list of "issuers" is configured, it takes priority over a configured single "issuer" URL.


dguggemos (approved):
LGTM 👍

PR #1469, opened 2022-09-01
fix failing patch updates of search index

dguggemos:
ThingUpdater actor increased its Thing revision after a skipped update which caused subsequent updates to have no effect (revisions of actor and database did not match).


yufei-cai (approved):
LGTM.

PR #1470, opened 2022-09-02
Don't take snapshot for entities in blocked namespaces

Yannic92:
This PR prevents snapshots from beeing taken in the persistence actor, when the namespace is blocked.
This is done to match the semantics of a blocked namespace not causing any actions anymore.

Also I removed the ditto-internal-utils-aggregator module and moved classes to edge-service and things service instead.


things/api/src/main/java/org/eclipse/ditto/things/api/ThingsMessagingConstants.java
```diff
@@ -26,22 +26,22 @@ public final class ThingsMessagingConstants {
     /**
      * Path of the root actor.
      */
-    public static final String ROOT_ACTOR_PATH = USER_PATH + "/thingsRoot";
+    public static final String ROOT_ACTOR_PATH = USER_PATH + "/thingsRoot/";
```
thjaeckle, line 29: this constant could also be private, correct?
I think we should adjust that ..

things/api/src/main/java/org/eclipse/ditto/things/api/ThingsMessagingConstants.java
```diff
@@ -26,22 +26,22 @@ public final class ThingsMessagingConstants {
...
 
-    /**
-     * Path of the actor that streams from the event journal.
-     */
-    public static final String THINGS_STREAM_PROVIDER_ACTOR_PATH = ROOT_ACTOR_PATH + "/persistenceStreamingActor";
+    public static final String THINGS_PERSISTENCE_STREAMING_ACTOR_NAME = "snapshotStreamingActor";
```
thjaeckle, line 31: A public constant without a javadoc?

things/api/src/main/java/org/eclipse/ditto/things/api/ThingsMessagingConstants.java
```diff
@@ -26,22 +26,22 @@ public final class ThingsMessagingConstants {
...
 
     /**
      * Path of the actor that streams from the snapshot store.
      */
-    public static final String THINGS_SNAPSHOT_STREAMING_ACTOR_PATH = ROOT_ACTOR_PATH + "/snapshotStreamingActor";
+    public static final String THINGS_SNAPSHOT_STREAMING_ACTOR_PATH =
+            ROOT_ACTOR_PATH + THINGS_PERSISTENCE_STREAMING_ACTOR_NAME;
+
+    public static final String THINGS_AGGREGATOR_ACTOR_NAME = "aggregator";
```
thjaeckle, line 39: A public constant without a javadoc?

thjaeckle (changes requested):
Only some minor remarks regarding constants and javadoc :)

thjaeckle (approved):
LGTM 👍

PR #1473, opened 2022-09-06
Fix normalized payload for merge events with extra fields

### LINKED ISSUE #1446: When merging a feature, the normalized payload does not contain full feature
When performing a change and the actual change is also part of what's in the extraFields, the normalized payload contains the actual change. When performing the change with a merge, it's possible that the normalized payload misses information (since you might have more data in the feature compared to what you actually update). In our case, we actually need the data which is not updated (and thus already present in the feature even before the update).
See example below:
Current object before update in ditto:
```
{
    "thingId": "the.namespace:the-thing-id",
    "policyId": "the.namespace:the-policy-id",
    "definition": "digitaltwin:DigitaltwinExample:1.0.0",
    "attributes": {
      "location": "Kitchen"
    },
    "features": {
      "transmission": {
         "properties": {
           "cur_speed": 90,
           "max_speed": 100
         }
       }
    }
}
```
Update:
path: `/features/transmission`
Type: merge
content:
```
{
    "properties": {
        "cur_speed": 80,
    }
}
```
Topic on connection:
`/_/things/twin/events?extraFields=features`

Result, non-normalized. Contains all data, but not normalized:
```
{
  "topic": "the.namespace/the-thing-id/things/twin/events/merged",
  "headers": {
    ...
  },
  "path": "/features/transmission",
  "value": {
    "definition": [
      "idigitaltwin:DigitaltwinExample:1.0.0"
    ],
    "properties": {
        "properties": {
            "cur_speed": 80,
        }
    }
  },
  "extra": {
    "features": {
        "transmission": {
            "properties": {
                "cur_speed": 80, <-- updated
                "max_speed": 100
            }
        }
    }
  }
}
```
Result, normalized. Contains updated feature but misses already existing parts of the feature
```
{
    "thingId": "the.namespace:the-thing-id",
    "policyId": "the.namespace:the-policy-id",
    "definition": "digitaltwin:DigitaltwinExample:1.0.0",
    "attributes": {
      "location": "Kitchen"
    },
    "features": {
      "transmission": {
         "properties": {
           "cur_speed": 80
           <!-- max_speed missing -->
         }
       }
    }
}
```
I would expect max_speed to be present in the normalized object.

jokraehe:
Fixes #1446

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/NormalizedMessageMapper.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/NormalizedMessageMapper.java
@@ -102,9 +102,6 @@ public final class NormalizedMessageMapper extends AbstractMessageMapper {
         final JsonObjectBuilder builder = JsonObject.newBuilder();
         builder.set(THING_ID, ThingId.of(topicPath.getNamespace(), topicPath.getEntityName()).toString());
 
-        // enrich with data selected by "extraFields", do this first - the actual changed data applied on top of that:
-        extraData.ifPresent(builder::setAll);
-
         if (path.isEmpty() && payloadValue.isPresent()) {
             final JsonValue value = payloadValue.get();
             if (value.isObject()) {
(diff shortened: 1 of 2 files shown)


dguggemos (approved):
LGTM (y)

PR #1474, opened 2022-09-06
enhanced ImplicitThingCreationMessageMapper to resolve an optional ExpressionResolver

thjaeckle:
* and fall back to the default one if no custom ExpressionResolver could be loaded


stmaute (approved):
LGTM 👍🏻

PR #1475, opened 2022-09-06
skip hostname verfication check with self signed in kafka

kalinkostashki:
Added the SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG in order to skip hostname verification checks as in general we don't expect self signed certs to necessarily match the container host name.

Signed-off-by: Kalin Kostashki <kalin.kostashki@bosch.io>


thjaeckle (approved):
LGTM 👍

PR #1479, opened 2022-09-13
Adjust the command response enforcement exception handling

DerSchwilk:
Currently, failed policy enforcer loading leads to error logging and 500 exceptions. This case happens when the corresponding policy is deleted before command response enforcement. Thus adjusted the error to 404 and removed error logging.

Additionally, the command response exceptions always get wrapped in an internal exception. Adjusted that exceptions which are already DRE are thrown as is.


dguggemos (approved):
LGTM

PR #1480, opened 2022-09-13
Filter MQTT messages where the message topic does not match the source topic

DerSchwilk:
Filter MQTT messages where the message topic does not match the source topic

Since now we consumed all messages that a client is subscribed to on all sources (not respecting the sources adresses)
The HiveMq client however makes it pretty hard to only consume messages which match a single subscribes topic (subscribePublished)
Since this would require major refactorings of the MQTT client connectivity, the messages not matching the sources adresses will now just be filtered out in the consumer actor.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/subscribing/MqttSubscriber.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/subscribing/MqttSubscriber.java
@@ -20,7 +20,10 @@ import java.util.Optional;
 
 import org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq.client.GenericMqttSubscribingClient;
 import org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq.client.MqttSubscribeException;
+import org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq.message.publish.GenericMqttPublish;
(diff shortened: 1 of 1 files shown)


dguggemos (approved):
LGTM 👍

thjaeckle:
Workaround for https://github.com/hivemq/hivemq-mqtt-client/issues/544

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/subscribing/MqttSubscriber.java
```diff
@@ -120,10 +123,25 @@ private Source<SubscribeResult, NotUsed> subscribe(final GenericMqttSubscribe ge
     private SubscribeResult consumeIncomingPublishesForSubscribedTopics(
             final org.eclipse.ditto.connectivity.model.Source connectionSource
     ) {
+        final List<MqttTopicFilter> topicFilters =
+                connectionSource.getAddresses().stream().map(MqttTopicFilter::of).toList();
```
jufickel-b, line 127: Why making a list out of the stream when it gets right afterwards converted to a stream again?

dguggemos: Because a stream can be consumed only once.

jufickel-b: :+1:

Issue #1441, opened 2022-08-16
Adjust Ditto Explorer UI to use Ditto "/connections" HTTP endpoint for managing connections


thjaeckle:
With #1406 Ditto will add a separate HTTP endpoint for managing connections (which had to be done via DevOps / "piggyback commands" before).

The current Ditto Explorer UI does use the "old piggyback approach" - and does not work with the current "main" branch of Ditto (which is about the be released as Ditto 3.0) because the format of the "piggyback commands" is simplified with Ditto 3.0).

The UI should:
* provide an option (e.g. per environment configuration) to run in either "Ditto 2.x mode" or in "Ditto 3.x mode"
* switch the "connections" functionality for the "Ditto 3.x mode" to use the connections HTTP endpoints

Issue #1406, opened 2022-07-04
Create HTTP API for managing connections


thjaeckle:
Currently, connection management in Ditto can only be done via "piggyback devops commands":
* https://www.eclipse.org/ditto/installation-operating.html#managing-connections
* https://www.eclipse.org/ditto/connectivity-manage-connections.html

This is quite cumbersome as the response is "wrapped" into a JSON envelope and e.g. the actual "statusCode" is a field in that JSON payload instead of being reflected as real HTTP status code.

This issue proposes to add a new HTTP API in order to manage connections, e.g.:
```
POST      /api/2/connections
GET       /api/2/connections/<connectionId>
PUT       /api/2/connections/<connectionId>
DELETE    /api/2/connections/<connectionId>
POST      /api/2/connections/<connectionId>/command
GET       /api/2/connections/<connectionId>/status
GET       /api/2/connections/<connectionId>/metrics
GET       /api/2/connections/<connectionId>/logs
```

This should reflect all available Connection commands with a HTTP endpoint.
The authentication of this API should be done with the Ditto "devops" user - as connections currently do not forsee having a "policyId" and this would lead to much effort adding the "policyId" to connections.

thjaeckle:
@thfries I know you are planning on also working on a "connections" functionality in the Ditto UI.
Please be aware that we aim to soon work on this issue, providing an HTTP API for the connections management instead of being forced to use "devops commands".

thfries:
Thanks for the heads-up, @thjaeckle. I'll try to keep this integration locally so it can easily replaced. But I would like to push the connections UI forward in parallel, because it is quite important.

PR #1484, opened 2022-09-16
Explorer UI - adjusted UI to use new HTTP routes for /connections

thjaeckle:
* also added an example template using a JS payload mapper

To be merged after #1406 is done.

Fixes: #1441 

@thfries would be cool if you could have a look on the changes

### DIFF
--- a/ui/modules/connections/connections.js
+++ b/ui/modules/connections/connections.js
@@ -121,11 +121,11 @@ export function ready() {
   });
 
   dom.buttonCreateConnection.onclick = () => {
-    Utils.assert(theConnection, 'Please enter a connection configuration (select   a template as a basis)', dom.editorValidationConnection);
-    if (API.env() === 'things') {
-      delete theConnection.id;
-    } else {
+    Utils.assert(theConnection, 'Please enter a connection configuration (select a template as a basis)', dom.editorValidationConnection);
+    if (API.env() === 'ditto_2') {
       selectedConnectionId = theConnection.id;
+    } else {
+      delete theConnection.id;
     }
     API.callConnectionsAPI('createConnection', loadConnections, null, theConnection);
   };
@@ -238,7 +238,7 @@ function setConnection(connection, isNewConnection) {
     if (theConnection.mappingDefinitions && theConnection.mappingDefinitions.javascript) {
       incomingEditor.setValue(theConnection.mappingDefinitions.javascript.options.incomingScript, -1);
       outgoingEditor.setValue(theConnection.mappingDefinitions.javascript.options.outgoingScript, -1);
-    };
+    }
   } else {
     dom.inputConnectionId.value = null;
     connectionEditor.setValue('');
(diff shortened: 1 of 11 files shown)


ui/modules/connections/connections.js
```diff
@@ -257,16 +257,16 @@ function loadConnections() {
...
-        row.insertCell(0).innerHTML = connection.name;
-      } else {
+      if (API.env() === 'ditto_2') {
         API.callConnectionsAPI('retrieveConnection', (dittoConnection) => {
-          row.insertCell(0).innerHTML = dittoConnection.name;
-        },
-        id);
+           row.insertCell(0).innerHTML = dittoConnection.name;
+         },
+         id);
```
thfries, line 267: Indentation was proclaimed by eslint
```suggestion
         row.insertCell(0).innerHTML = dittoConnection.name;
       },
       id);
```

thfries (changes requested):
Thanks for doing this, @thjaeckle.
LGTM. The new ditto 3 connection API is not yet on master, so I can not test it?
I created an input field for the new ditto version and tried to provide it as a patch here. Does that work?
[versionFieldPatch.txt](https://github.com/eclipse/ditto/files/9591676/versionFieldPatch.txt)

thjaeckle:
> Thanks for doing this, @thjaeckle. LGTM. The new ditto 3 connection API is not yet on master, so I can not test it? I created an input field for the new ditto version and tried to provide it as a patch here. Does that work? [versionFieldPatch.txt](https://github.com/eclipse/ditto/files/9591676/versionFieldPatch.txt)

The HTTP API is not yet merged - we think it however will be this week.
I just wanted to provide the adjustments so that we quickly can move forward releasing Ditto 3.0.

I included your patch with the input field for the Ditto version, thx 👍

ui/modules/environments/environments.html
```diff
@@ -51,6 +51,10 @@ <h5>Environments</h5>
                         title="URI including protocol, host and port (e.g. http://localhost:8080)">API URI</label>
                     <input type="text" class="form-control form-control-sm" id="inputApiUri"></input>
                 </div>
+                <div class="input-group input-group-sm mt-1">
+                    <label class="input-group-text">Ditto Version</label>
+                    <input type="integer" class="form-control form-control-sm" id="inputDittoVersion"></input>
```
thjaeckle, line 56:
```suggestion
                    <input type="number" class="form-control form-control-sm" id="inputDittoVersion"></input>
```

thfries:
Nice! I like all the changes. Thanks for all the improvements!

ui/modules/things/things.html
```diff
@@ -88,7 +88,7 @@ <h5>Things</h5>
                                 <div class="invalid-feedback"></div>
                             </div>
                             <div class="input-group input-group-sm mb-1">
-                                <label class="input-group-text">Definition</label>
+                                <label class="input-group-text">Definition template</label>
```
thfries, line 91: I would stick to just "Definition", as it represents the definition of the thing (if you select a thing, the definition is filled into that field).

thjaeckle: Ok, agreed, I will undo this. The "template" mechanism with the dropdown at this place however does not have much to do with the "definition" in my opinion.

thjaeckle, line 91:
```suggestion
                                <label class="input-group-text">Definition</label>
```

PR #1487, opened 2022-09-20
Make MQTT 5 SessionExpiryInterval configurable

DerSchwilk:
This makes the MQTT 5 expiry interval configurable. The default applies by HiveMQ was 0s, which then lead to the redelivery by a broker failing, since the session expired directly after the reconnect necessary for the redelivery.

Also this PR uses a copied byte payload of the MQTT message for payload mapping. Before mapping the byte payload failed, due to a write access on a read only ByteBuffer. This prevented i.e. the raw message mapper from working with MQTT3/5 messages.


thjaeckle (approved):
LGTM 👍

PR #1490, opened 2022-09-23
adjust log levels of BaseClientActor and AmqpConsumerActor

kalinkostashki:
changed loggers to WARN where appropriate to avoid logging errors for problems on client side backends

Signed-off-by: Kalin Kostashki <kalin.kostashki@bosch.io>


thjaeckle (approved):
LGTM, thx @kalinkostashki 👍

Issue #1463, opened 2022-08-26
Fix that adding custom Java MessageMappers to Ditto via classpath is no longer possible


thjaeckle:
Ditto's [documentation of custom Java based implementation for mappers](https://www.eclipse.org/ditto/connectivity-mapping.html#custom-java-based-implementation) states that a custom Java implementation can be added for adding a performant, Java based mapping.

That however no longer works as the MessageMappers are annotated with ClassIndex annotations which are scanned during compile-time and added to a registry of all known mappers.

This cannot be altered with classpath extension.

In order to fix that, we could use our new `DittoExtensionPoint` mechanism to support adding custom mappers via classpath.

PR #1492, opened 2022-09-26
Fix that custom Java message mappers could not longer be added via classpath

alstanchev:
As the fix, we changed the mechanism of how message mappers are loaded to loading them via the new DittoExtensionPoint mechanism introduced for Ditto 3.0.

New mappers can easily be added by adding another .jar file to the classpath of Ditto's "connectivity" service and adding the new mapper via configuration to the connectivity service.

An example of how to provide a Protobuf based mapper is provided via https://github.com/eclipse/ditto-examples/pull/62

Fixes: https://github.com/eclipse/ditto/issues/1463


thjaeckle (approved):
Reviewed as part of peering 👍

PR #1495, opened 2022-09-29
Make it possible to retrieve disabled loggers via devops API

jokraehe:
The current implementation filters all loggers set to "off". Add query parameter to control that behaviour.

### DIFF
--- a/base/service/src/main/java/org/eclipse/ditto/base/service/devops/LogbackLoggingFacade.java
+++ b/base/service/src/main/java/org/eclipse/ditto/base/service/devops/LogbackLoggingFacade.java
@@ -59,11 +59,11 @@ public final class LogbackLoggingFacade implements LoggingFacade {
 
     @Override
     @Nonnull
-    public List<LoggerConfig> getLoggerConfig() {
+    public List<LoggerConfig> getLoggerConfig(final boolean includeDisabledLoggers) {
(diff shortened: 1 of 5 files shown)


base/service/src/main/java/org/eclipse/ditto/base/service/devops/LogbackLoggingFacade.java
```diff
@@ -73,17 +73,24 @@ public List<LoggerConfig> getLoggerConfig(@Nonnull final Iterable<String> logger
                 .map(logger -> (Logger) LoggerFactory.getLogger(logger))
                 .toList();
 
-        return loggerConfigsFor(loggerList);
+        return loggerConfigsFor(loggerList, true);
```
DerSchwilk, line 76: Is there a reason this is statically true?
Shouldn't this be either controlled via a parameter or set to false to resemble the prior behaviour?

DerSchwilk (commented):
One question, other than that LGTM :)

base/service/src/main/java/org/eclipse/ditto/base/service/devops/LogbackLoggingFacade.java
DerSchwilk: Changed

DerSchwilk (approved):

Issue #1449, opened 2022-08-19
Select Ditto Explorer UI "environment" via query parameter


thjaeckle:
For the Ditto Sandbox (for example) it would be really helpful if one could provide a query parameter when opening the Ditto Explorer UI specifying which "environment" to use.

A similar feature is possible for Swagger-UI where you can define multiple "definitions" (OpenAPI specs the Swagger-UI should show in its dropdown) and specify the initial one to load via a query parameter.
Swagger-UI describes its options [here](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) (`urls` and `urls.primaryName` are the ones I thought about).

Maybe it's worth thinking about whether the Ditto UI "templates" should also be available (and referencable) as separate files on a webserver similar to Swagger-UI `urls`.

@thfries what do you think about this?

thfries:
First of all: yes, I am also annoyed to need to select the right environment all time 😉

The solution in my mind was to store the last environment of a user in the browser cache - like with all other user specific settings. The Swagger-UI solution looks powerful. So you are thinking to allow to configure all environments and the default environment from outside? Interestingly enough Swagger does not save any user or browser specific data.

I think a user should be able to change his personal settings directly in the UI. A way to share the settings of one user with others is also very useful, but I don't see spontaneously a way to combine both approaches (one idea was to store the environment configuration in a thing for each environment but that is another completely different approach).

You see, I am a bit confused now. Need to think about it and some further discussion would be helpful, too. Thanks.

thfries:
Some more thoughts (hope I am not overcomplicating things):
* The environment templates should be moved to the templates folder (they are currently in the environments.js)
* The environment templates are loaded only the very first time on one browser instance. As soon as a user does any change on any environment (e.g. just authenticates), all environments are taken from the browser store. You will never get an update on the templates again until you clear your browser cache. That is not good. At least each of the environments should be stored independently

Next to that an environment stores different types of data:

1. environment configuration itself (url, solutionId, namespace, ditto-version, etc.): not user specific, should be shared
2. Authorization header (to switch the headers when you switch the environment): should NOT be shared.
3. Fieldlist: the field definitions should be shared with others, while the selection and order of columns is user specific
4. Filterlist: user specific but may also be shared
5. pinned things: very user specific
6. Message Templates: similar to field list, these are not user specific

So as a first solution idea:
* Separate 1 from 2 - 6
* 1 can be configured via url parameter:
[...]

Issue #1496, opened 2022-09-29
Mention Ditto Explorer UI in Ditto documentation


thjaeckle:
Provide a new page for that:
* where is the latest (development branch) UI deployed? -> https://eclipse.github.io/ditto/
* that is is available as Docker image
* Short intro on the "Environments" and how they are used/configured

thjaeckle:
Resolved via #1509

PR #1501, opened 2022-10-04
Explorer UI: URL parameters for environments

thfries:
Hi ditto team,

please find this PR that addresses issue: #1449 

I allows to
* load environments via `environmentsURL` parameter
* select an environment via `primaryEnvironmentName`

Additional improvements:
* Stacking for (multiple) error toasts (because invalid URL parameters can cause additional errors that were not visible)
* Avoid errors to be thrown twice in ditto rest call
* Optimized layout for long thingIds in tables. Multi-line cells are avoided in general by ellipsis and tooltip. Hidden experimental: you can set `shortenUUID` to true and define a `defaultNamespace` in the environment to get better readable thingIds
* fixed some eslint issues in things.js
* Small fix to de-select favorite filter in case of pinned things

Looking forward to your feedback, Thomas

### DIFF
--- a/ui/index.css
+++ b/ui/index.css
@@ -24,6 +24,13 @@
     --bs-table-hover-color: white;
 }
 
+td {
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+    max-width: 300px;
+}
+
 thead {
     position: sticky;
     top: 0;
--- a/ui/index.css
+++ b/ui/index.css
@@ -28,7 +28,7 @@ td {
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
-    max-width: 300px;
+    max-width: 15em;
 }
 
 thead {
--- a/ui/index.html
+++ b/ui/index.html
@@ -101,16 +101,7 @@
         </div>
     </div>
     <!-- Notifications to user ----------------------------------------------------------------->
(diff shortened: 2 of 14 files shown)


thjaeckle:
@thfries again, thanks for the great contribution 👍 
I tried the feature out a little and what I immediately thought:
* it would be great that if you select an environment with the dropdown at the top right to set the query paramter `primaryEnvironmentName=<selected-env>` - that way refreshing the page directly has the correct environment again - and the page incl. selected environment is bookmark-able

Did you already try out the `environmentsURL` feature with a custom hosted environment JSON?

I think once this PR is merged we should really tackle #1496 and add some documentation for the Ditto-UI (especially the "environments" feature) there as well..

thfries:
Hi @thjaeckle,

smart idea, thank you! I added a commit for that 👍

yes, I tested with some sample environments here: https://gist.githubusercontent.com/thfries/5e722197bd74be80fc1c02473753a1e6/raw/4ec788886090daa6c62d90dd5d75e725d80935ab/environments.json. I am going to test next week with some real environments.

Thomas

thjaeckle (approved):
Hi @thfries 
The feature looks great - thanks for adding the suggested parameter change on environment switch, works 👍 

FMPOV I would approve the PR - please let me know when it would be ready to merge from your side :)

thfries:
OK from my side, too. Removing draft status. Thanks 👍

Issue #1503, opened 2022-10-06
Cannot invoke "String.getBytes()" because "payload" is null


tobias-zeptio:
Reading section 3.2 in https://www.amqp.org/sites/amqp.org/files/amqp.pdf the body of an AMQP 1.0 message is defined as follows:

- The body consists of either: one or more data sections, one or more amqp-sequence sections,
or a single amqp-value section.

A valid amqp value is null, which the ditto-connectivity AmqpConsumerActor doesn't seem to handle in the `extractPayloadFromMessage` method.

I think adding a null check and returning a ExternalMessageBuilder with null values should be valid.

The use-case here is handling retrieve commands in the payload-mapper, in our case the header contains which property to read.

tobias-zeptio:
Looks like UnmodifiableExternalMessageBuilder accepts null as payload so I think this would suffice, so no change needed there.

I'm looking into how to get the body from the `JmsMessage`, but it seems to always return `null` which is strange.

```
    @Override
    public boolean isBodyAssignableTo(@SuppressWarnings("rawtypes") Class target) throws JMSException {
        return true;
    }

    @Override
    public final <T> T getBody(Class<T> asType) throws JMSException {
        if (isBodyAssignableTo(asType)) {
            return doGetBody(asType);
        }

        throw new MessageFormatException("Message body cannot be read as type: " + asType);
    }

    protected <T> T doGetBody(Class<T> asType) throws JMSException {
        return null;
    }
```

We use qpid-jms-client in another project, and there is straight up returns the body but that uses `JmsMessageFacade` instead of `JmsMessage`.

tobias-zeptio:
Resolved by #1505

PR #1505, opened 2022-10-06
Accept and test TextMessage with null payload

tobias-zeptio:
Adds the ability to handle TextMessage with null payload, from issue #1503 

Signed-off-by: Tobias Månsson <tobias@zept.io>

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/amqp/AmqpConsumerActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/amqp/AmqpConsumerActor.java
@@ -461,7 +461,11 @@ final class AmqpConsumerActor extends LegacyBaseConsumerActor
             final ExternalMessageBuilder builder, @Nullable final String correlationId) throws JMSException {
         if (message instanceof TextMessage textMessage) {
             final String payload = textMessage.getText();
-            builder.withTextAndBytes(payload, payload.getBytes());
+            if (payload == null) {
+                builder.withText(null);
+            } else {
+                builder.withTextAndBytes(payload, payload.getBytes());
+            }
         } else if (message instanceof BytesMessage bytesMessage) {
             final long bodyLength = bytesMessage.getBodyLength();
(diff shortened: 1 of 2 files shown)


tobias-zeptio:
I tried testing with a BytesMessage, but couldn't figure it out. 

Reading through the code in `AmqpJmsBytesMessageFacade` `BytesMessage` seems to handle a `null` body already.

thjaeckle (changes requested):
Looks almost good to me - just 2 smaller things to adjust in the test.
The byte `null` payload I could also not understand in the JMS / Qpid facade API .. 
I think it is sufficient to support `null` textMessage.

tobias-zeptio:
I tried to add a BytesMessage test as well, but can't get it. Creating a empty BytesMessage makes it write-only for some reason, and setting up the body creation using byteStream is beyond me. This is my attempt:

```
    @SafeVarargs // varargs array is not modified or passed around
    private JmsMessage getEmptyJmsByteMessage(final String correlationId, final Map.Entry<String, ?>... headers) {
        try {
            final AmqpJmsBytesMessageFacade messageFacade = new AmqpJmsBytesMessageFacade();
            // give it a connection returning null for all methods to set any AMQP
            // properties at all
            messageFacade.initialize(mock(AmqpConnection.class));
            messageFacade.setCorrelationId(correlationId);

            final JmsMessage message = messageFacade.asJmsMessage();
            JMSPropertyMapper.setPropertiesAndApplicationProperties(messageFacade.asJmsMessage(),
                    Arrays.stream(headers)
                            .collect(Collectors.toMap(Map.Entry::getKey,
                                    e -> e.getValue().toString())),
                    mock(ThreadSafeDittoLoggingAdapter.class));
            message.setAcknowledgeCallback(mockJmsAcknowledgeCallback());
            return message;
        } catch (final JMSException e) {
            throw new JMSRuntimeException(e.getMessage(), e.getErrorCode(), e.getCause());
        }
    }
```

thjaeckle (approved):

thjaeckle:
@tobias-zeptio I also tried to add a test with a `null` ByteMessage and failed to create it - the byte message does not seem to support this in the used AMQP client.

If you find out that you still require this also for byte messages, I suggest to create a follow-up PR.

PR #1508, opened 2022-10-07
Fixes automatic start of known connections

alstanchev:
Fixes automatic start of known connections.
Changes store type from ddata to event sourcing


thjaeckle (approved):
Reviewed as part of pair programming 👍

PR #1519, opened 2022-10-23
Allow to use namespaces in search in Explorer UI

thfries:
Hi @thjaeckle,

I have a small PR. This allows search to work with namespaces.

Next to that
- Allow to rename an environment
- Small fix: search query fields contained trailing comma

Thank you for review, Thomas


thjaeckle (approved):
LGTM, thanks again for the contribution @thfries

PR #1522, opened 2022-10-27
Adjust blocked subnet validation for IPv6

DerSchwilk:
When configuring blocked subnets and creating a new connection to an IPv6 address, the connection fails, because SubNetUtils can't parse IPv6 addresses.

Thus the change to IPAddressMatcher was done.


stmaute (approved):
LGTM

PR #1530, opened 2022-11-04
Tracing tweaks and Java Result type

jufickel-b:
With this PR the existing tracing implementation of Ditto was reworked:
* Refactored API to be more lean and have more concise names.
* `DittoTracing` now is more stern in a sense that using it before having initialised it would lead to an exception.
* Introduced the ability to configure filters for operation names to include or exclude from tracing.

Furthermore, Ditto now ships its own `Result` type! It is comparable to Scala's `Try` but is written completely in Java and is more flexible. The idea and parts of its API are highly inspired by Rust's Result type but also by Java's `Optional`.
`Result` is meant as return type for methods that might fail. This makes it impossible to neglect runtime exceptions and is more fun to work with than having to deal with checked exceptions. However, for its full potential to become obvious, Java's pattern matching for switch is required to be usable in Ditto first.


dguggemos (approved):
LGTM :+1:

PR #1531, opened 2022-11-05
Improved search filter in explorer ui

thfries:
Two small extensions to improve usage of things searches

### Create search filter for fields
- There is often the need to do a search for known fields in some attribute or feature (like serial number, device type, ...). If you have defined a field (that is already used for columns), you can easily create a rql with the correct path of a field (only eqal search by now)
- Future ideas:
  - have a friendly name for each search filter (like "serialnumber == 42" for "eq(attributes/serialnumber,42")
  - generate all variants for eq, like, not eq, etc.
  - replace the drop down selection with a typeahead search field

### Simple session based history of searches
- If you forget to save a filter as a favourite, you need to type it again.
- Now you have a transient history to recall you last search. This history is cleaned up automatically on the next session.

### DIFF
--- a/ui/index.css
+++ b/ui/index.css
@@ -151,3 +151,8 @@ h5, h6 {
 hr {
     margin: 0.25rem;
 }
+
+.dropdown-menu {
+    max-height: 80vh;
+    overflow-y: auto;
+}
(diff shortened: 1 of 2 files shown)


thjaeckle (approved):
LGTM, thx 👍

PR #1534, opened 2022-11-13
Explorer UI: fixed bug caused by filter dropdown

thfries:
Hi @thjaeckle,
I found a bug and provided a PR to fix it.


thjaeckle (approved):
LGTM, thx

PR #1538, opened 2022-11-29
cancelReceiveTimeout moved after exception creating

abalarev:
Canceling ReceiveTimeout handling in AbstractHttpRequestActor causes IllegalArgumentException.
This happens because of a missing (set to Undefined) timeout Duration.


thjaeckle (approved):
LGTM 👍

PR #1545, opened 2022-12-06
dependency updates

thjaeckle:
* update sshd-core to 2.9.2
* remove commons-net dependency which is no longer needed


kalinkostashki:
LGTM :)

PR #1546, opened 2022-12-06
fix: move visitor accept implementation to the undermost class

piu130:
When calling `rootNode.accept(v)`, `v.visit(Node node)` is called instead of `v.visit(RootNode node)`.

### DIFF
--- a/rql/model/src/main/java/org/eclipse/ditto/rql/model/predicates/ast/RootNode.java
+++ b/rql/model/src/main/java/org/eclipse/ditto/rql/model/predicates/ast/RootNode.java
@@ -17,4 +17,10 @@ package org.eclipse.ditto.rql.model.predicates.ast;
  */
 public final class RootNode extends SuperNode {
 
+    /**
+     * {@inheritDoc}
+     */
+    public void accept(final PredicateVisitor predicateVisitor) {
+        predicateVisitor.visit(this);
+    }
 }
--- a/rql/model/src/main/java/org/eclipse/ditto/rql/model/predicates/ast/SuperNode.java
+++ b/rql/model/src/main/java/org/eclipse/ditto/rql/model/predicates/ast/SuperNode.java
@@ -38,13 +38,6 @@ public abstract class SuperNode implements Node {
         return children;
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public void accept(final PredicateVisitor predicateVisitor) {
-        predicateVisitor.visit(this);
-    }
-
     @Override
     public String toString() {
         return "SuperNode [children=" + children + "]";


piu130:
Hi @thjaeckle , the visitor pattern implementation is always the same:
```java
public void accept(final PredicateVisitor predicateVisitor) {
    predicateVisitor.visit(this);
}
```
If we could implement it the way it is currently (implementation in super class), we could also create a default implementation in the interface (Node) itself. This does not work.
The problem with the current implementation arises, when we call `rootNode.accept(visitor)`. Since rootNode has no accept implementation, it takes the one from SuperNode. SuperNode itself calls `visitor.visit(this)`, where `this` is of type SuperNode. So instead of `visitor.visit(RootNode node)` it calls `visitor.visit(Node node)`, which is wrong.

You're right, there is a breaking change. Maybe we should implement `accept` in RootNode and leave SuperNode as is?

piu130:
This sample (kotlin):
```kotlin
class PrintVisitor : PredicateVisitor {
    override fun visit(node: RootNode) {
        println("RootNode")
    }

    override fun visit(node: LogicalNode) {
        println("LogicalNode")
    }

    override fun visit(node: SingleComparisonNode) {
        println("SingleComparisonNode")
    }

    override fun visit(node: MultiComparisonNode) {
        println("MultiComparisonNode")
    }

    override fun visit(node: ExistsNode) {
        println("ExistsNode")
    }

    override fun visit(node: Node) {
        println("Node")
    }
}

fun main() {
    val parser: RqlPredicateParser = RqlPredicateParser.getInstance()
    val query: String = "eq(a,1)"

    val rootNode: RootNode = parser.parse(query)

    val visitor: PrintVisitor = PrintVisitor()

    print("1: ")
    rootNode.accept(visitor)

    print("2: ")
    visitor.visit(rootNode)
}
```
Gives the following output:
```
1: Node
2: RootNode
```
But it should print `RootNode` twice, right?

What I'm doing currently:
```kotlin
override fun visit(node: Node) {
    if (node is RootNode) {
        visit(node)
    } else {
        println("Node")
    }
}
```
Which prints `RootNode` twice.
So, there is this method `PredicateVisitor.visit(RootNode node)` which is never called when using the visitor pattern correctly.

thjaeckle:
I understood the code issue on a technical level.
But I would like to understand its severity and whether it influences the RQL parsing in a way which would require a bugfix release.

What we discussed - copying the method and overwriting it - we would happily accept as contribution.
Please make sure to fulfil the documented [legal requirements](https://github.com/eclipse-ditto/ditto/blob/master/CONTRIBUTING.md#legal-requirements) and sign your commit with the email address registered in your Eclipse account, so that we are able to merge.

thjaeckle:
@piu130 ping :)
Ditto 3.1.0 is probably coming in the next weeks - so if this should make it into the release we would need the mentioned change and the Eclipse legal stuff done.

piu130:
Hi @thjaeckle , sorry for the delay. I've just read some more articles about the visitor pattern. [This here for example](https://stackoverflow.com/questions/13979160/how-does-the-visitor-pattern-not-violate-the-open-closed-principle).
The pattern there is to create another method `visitOther(Node node)`. I think, this would make it more clear but would definitely introduce a breaking change. I'll leave it for now (I don't have an eclipse account anyway, I missed that when reading the contribution guide, sorry!).
Thank you very much for taking the time!

PR #1549, opened 2022-12-11
Bugfix/client rebalancing

yufei-cai:
- Ensure that client supervisors do not get stuck if restart message is lost.
- Increase rebalancing relative limit to coincide with absolute limit when all shards are active.


thjaeckle (approved):
LGTM, thx @yufei-cai 👍

PR #1554, opened 2022-12-15
Wrong comment on config of json pointers to values to be encrypted

alstanchev:
This fixes a misleading comment in the connectivity.config regarding the fields to be encrypted of a connection

Signed-off-by: Stanchev Aleksandar <aleksandar.stanchev@bosch.io>


thjaeckle (approved):
LGTM

PR #1555, opened 2022-12-15
Policy imports

kalinkostashki:
Enables the imports of policies.
One policy would now be able to import other policies(this isn't transitive).
The maximum number of imported policies is also configurable. Currently set to 10.

Resolves: #298


thjaeckle (approved):
Code reviews were done as part of development process by the Ditto team.

Approving on behalf of the Ditto team 👍

PR #1560, opened 2022-12-28
Revert client actors from cluster sharding to router pool

yufei-cai:
Sharded client actors are not guaranteed to be distributed evenly. It caused uneven load on connectivity instances and made metric reporting inaccurate where the assumption of one client actor per cluster member was made.


thjaeckle (approved):
LGTM 👍

Issue #1498, opened 2022-09-29
Provide API to stream/replay persisted events from the event journal


thjaeckle:
Ditto applies the "Event Sourcing" paradigm for its persistence (of things, policies and connections):
* https://blog.softwaremill.com/3-reasons-to-adopt-event-sourcing-89cb855453f6
* https://eventuate.io/whyeventsourcing.html

Ditto however does not yet make much use of that persistence paradigm.
I would like to propose to provide APIs to be able to retrieve a stream of persisted events from the "event journal".
That can be useful e.g. in order to:
* replay events which were "missed" (similar to what Apache Kafka provides)
* query the changes made to a single property value of a thing
* get an "audit log" of changes done to a thing

The API would only be available in a streaming way, so for:
* SSE 
* WebSocket via DittoProcol
* Connections via DittoProtocol

The DittoProtocol API should be very similar to the [search protocol](https://www.eclipse.org/ditto/protocol-specification-things-search.html).

As the Ditto background cleanup would delete events quite fast, we would also need a configuration of a "event-retention duration" - so it would be possible to configure how long to keep events and snapshots in the DB before finally deleting them via the cleanup.
Example: 7 days

An idea how the SSE API could look like:
```
/api/2/things/org.eclipse.ditto:thing-2?fromRevision=0&fields=thingId,attributes,_revision,_modified
```
The stream would automatically end when the most recent event was returned.

This API could be provided for things, policies and connections (for connections however probably only available via piggyback commands).
The idea would be to use either:
* the *_revision* number of the entity as "from/to" selector
* or the *_modified* timestamp of the event journal entry (which is included in the the MongoDB `_id` field)

With the timestamp, a query could look like:
An idea how the SSE API could look like (to get all changes within one hour):
```
/api/2/things/org.eclipse.ditto:thing-2?fromModified=2022-09-29T12:00:00.000Z&toModified=2022-09-29T11:00:00.000Z&fields=thingId,attributes,features,_revision,_modified
```

thjaeckle:
Input on whether that would be useful by our community would be much appreciated
@BobClaerhout @w4tsn @tobias-zeptio @thlandgraf

hjccjugfcc:
Hello, we find that the historical data in ditto is queried in the things-journal table of the mongodb database in docker, but we also find that these historical data will be deleted every fixed time, and only the latest updated data is retained in the database. We want to know, can this fixed time be modified ? If it can be modified, where should it be modified ?
[...]

PR #1564, opened 2023-01-13
[#1498] Provide historical API access functionality for things, policies, connections

thjaeckle:
This PR enables "History API" support:
* streaming of journal entries of a given entity (thing/policy) as DittoProtocol events (via Websocket/Connections/SSE)
* accessing an entity (thing/policy/connection) "at" a specific revision or "at" a specific timestamp
* added a configurable "history-retention-duration" config for things/policies/connections in order to delay the cleanup for at least that amount of time -> that way we can e.g. keep 30d of snapshots/events
* made history API work via connectivity
* added documentation for history capabilities

Resolves: #1498

There is a new FeatureToggle in `ditto-devops.conf` with key `ditto.devops.feature.historical-apis-enabled` in order to disable the historical API access.
Also, in order to be useful, the "cleanup" of entities has to be configured to have a `history-retention-duration` of greater than 0.
By default this is configured to `0d` which means that events and snapshots may be deleted at any time of the cleanup strategy of Ditto decides so.

If for example "things" should keep a history of up to one week, the configuration inside `things.conf` can be adjusted to `history-retention-duration = 7d`.

The added documentation page `basic-history.md` describes the feature and its configuration in detail.


alstanchev (approved):
LGTM :)

harinip2701:
I am working on a project that involves fetching Ditto historical data from a Delta Lake database and displaying it in an Angular UI **via Eclipse Ditto**. The workflow I aim to implement is as follows:

Angular sends a request to Ditto for historical data of a specific device 
Ditto checks permissions for the request to ensure user authorization.
Ditto forwards the request to a DB service(spark)  via its Connectivity API.
The DB service queries Delta Lake for the required historical data and sends the data back to Ditto.
Ditto sends the processed data to the Angular UI for visualization.

- Is it possible to implement this workflow using Eclipse Ditto’s Connectivity API??

- How can I configure Ditto’s Connectivity API to connect to a DB service?

- if this workflows is possible means could you explain??

Issue #1563, opened 2023-01-11
outgoing mqtt connections fail to publish messages when tracing is enabled


JeffreyThijs:
If we enable open tracing on the ditto services, active outgoing mqtt connections (other types might have this problem as well) fail to publish their messages due to failing to transform `ExternalMessage` to `GenericMqttPublish` due to a blank `tracestate` header in the `ExternalMessage` headers.

The following error is thrown:

```
Ran into a failure when publishing signal: Failed to send message: 
Failed to transform ExternalMessage to GenericMqttPublish: The value 
must not be blank. Sending the message to an external system failed, 
please check if your connection is configured properly and the target 
system is available and consuming messages. - Message headers: 
[content-type=application/vnd.eclipse.ditto+json, tracestate=, 
context-tags=upstream.name=ditto-connectivity;, 
traceparent=00-3f0f66b556dc61e2250e5355bb33f83a-142e83f65527125d-01, 
correlation-id=782a32b1-4660-462d-8ac7-51d865bc3d0c] 
```
Is this expected behavior? If so, why is it required that the headers are blank? It is a bit silly that this requirement completely disrupts the publication of messages.

thjaeckle:
This is surely not the expected behavior. Looks like a bug to me.

The error message seems to come from:
https://github.com/eclipse-ditto/ditto/blob/21f1b0d512857f37b8072d960dbe11de761f9887/internal/utils/metrics/src/main/java/org/eclipse/ditto/internal/utils/metrics/instruments/tag/Tag.java#L50-L60

There were recently some changes to the tracing done in #1530 - probably a bug was introduced there.
@jufickel-b could you have a look?

thjaeckle:
@JeffreyThijs if this is urgent to you, I suggest you propose a PR fixing the issue.
As tracing is currently not in use by the Bosch team, this probably won't get high priority.

JeffreyThijs:
@thjaeckle, ok thanks for the clarification. I'll try to provide a PR to fix this problem. However, we encountered this problem with mqtt3 sessions though.

thjaeckle:
Fixed via #1565

PR #1565, opened 2023-01-13
external message to mqtt publish transform should not fail on blank header values

JeffreyThijs:
potential fix for #1563


thjaeckle (approved):
The fix looks good to me, so approval from my side 👍 
If you mark the PR as "ready" it is good to be merged FMPOV

JeffreyThijs:
Tested the fix on our cluster and it fixed the problem so ready to merge!

PR #1567, opened 2023-01-20
fixes unwanted Connections url userInfo decoding

alstanchev:
Fixes unwanted password url decoding before encryption which later leads to problems when initialising Connection URI.


thjaeckle (approved):
LGTM 👍

PR #1568, opened 2023-01-24
Credentials of URI in Hono-connection encoding added

abalarev:
When credentials of Hono-connection contain special chars (like slash), the generated URI is not valid.
- credentials encoding is added
- tests are updated to cover this case

Signed-off-by: Andrey Balarev <andrey.balarev@bosch.io>


thjaeckle (approved):
LGTM, thx @abalarev

Issue #1087, opened 2021-06-11
Things-Search Case Sensitivity Option


dtmb91:
Currently, in the Ditto API for Things-Search there is no option for case sensitivity.

It is explicitly mentioned here:
"The search is case sensitive. In case you don't know how exactly the spelling of value of the namespace, name, attribute, feature etc. is, use the like notation instead of eq for filtering."
https://www.eclipse.org/ditto/http-api-doc.html#/Things-Search/get_search_things

My proposal is that the "like" filter accepts a third parameter dictating if case will be ignored or not.

thjaeckle:
Hi @dtmb91 
That's a good idea supporting case insensitive searches.
I suggest using a new "function" name for that, e.g. `likeIgnoringCase(<key>,"valUE*")`

A PullRequest for that would be very welcome, as we currently don't have a priority on adding that.
Please let us know if you need guidance where to make additions to the code for supporting that.

thjaeckle:
Here some hints regarding how to implement:
* https://github.com/eclipse/ditto/blob/master/rql/query/src/main/java/org/eclipse/ditto/rql/query/criteria/LikePredicateImpl.java is the implementation of `like` - I would copy from that
* look at where the constructor of `LikePredicateImpl` is used - that is `CriteriaFactoryImpl` in the `CriteriaFactory.like(Object)` implementation
   * add another interface method to `CriteriaFactory` for the new like with case insensitivity
* look at where `CriteriaFactory.like(Object)` is used, that is in `ParameterPredicateVisitor`
   * add something new to ParameterPredicateVisitor similar to the existing one for "like":
   ```java
   SINGLE_COMPARISON_NODE_MAPPING.put(SingleComparisonNode.Type.LIKE, CriteriaFactory::like);
   ```
* look where `SingleComparisonNode.Type.LIKE` enum value is used
   * that is the Scala based `RqlPredicateParser.like`: https://github.com/eclipse/ditto/blob/master/rql/parser/src/main/scala/org/eclipse/ditto/rql/parser/internal/RqlPredicateParser.scala#L105
   * add something new for that as well
   * as well as for its usage, etc.
* enhance `PredicateVisitor` and its implementations with the added "like ignoring case sensitivty": https://github.com/eclipse/ditto/blob/master/rql/query/src/main/java/org/eclipse/ditto/rql/query/criteria/visitors/PredicateVisitor.java
[...]

PR #1577, opened 2023-02-13
Adding case insensitive search for things-search

Abhijeetmishr:
Resolves: #1087


base/model/src/test/java/org/eclipse/ditto/base/model/common/LikeHelperTest.java
```diff
@@ -29,6 +30,15 @@ public void testWildcards() {
         assertExpression("foo..bar", "*bar", true);
         assertExpression("foo.bar.baz", "bar", false);
         assertExpression("foo.bar.baz", "*bar*", true);
+        //case insensitive test cases appending this  (?i) before text
```
thjaeckle, line 33: I would move the case insensitive tests to another test method.
Also, it seems the test is failing.

rql/query/src/test/java/org/eclipse/ditto/rql/query/things/ThingPredicatePredicateVisitorTest.java
```diff
@@ -137,6 +137,13 @@ public void matchingStringLike() {
                 .isTrue();
     }
 
+    @Test
+    public void matchingStringILike() {
+        // the sut already works on regex Pattern - the translation from "*" to ".*" followed by case insensitivity is done in LikePredicateImpl
+        doTest(sut.visitLike("this-is.*"), JsonValue.of("THIS-IS-THE-CONTENT"))
```
thjaeckle, line 143: `visitLike`?
You probably want to test visitILike, or not?
The comment is copy&paste from the like test and wrong.

thjaeckle (changes requested):

base/model/src/test/java/org/eclipse/ditto/base/model/common/LikeHelperTest.java
Abhijeetmishr: Yes I right some tests are failing because of this (?i). I will use flag Pattern.CASE_INSENSITIVE with Pattern.compile

thjaeckle:
Thanks @Abhijeetmishr for the contribution - a small, but important contribution 👍

Abhijeetmishr:
> Thanks @Abhijeetmishr for the contribution - a small, but important contribution 👍

@thjaeckle It would not have been possible without your support looking forward to many more.
Regards

PR #1580, opened 2023-02-20
Explorer UI - autocomplete for search

thfries:
please find this PR that enables a typeahead search for things search.
Let me know, what you think about it.
There is also some work prepared to filter the connection logs but this is not ready yet and commented out.

Best regards, Thomas

### DIFF
--- a/ui/modules/things/searchFilter.js
+++ b/ui/modules/things/searchFilter.js
@@ -40,7 +40,7 @@ const dom = {
   favIcon: null,
   searchFilterEdit: null,
   searchThings: null,
-  searchFavourite: null,
+  searchFavorite: null,
   tabThings: null,
   pinnedThings: null,
 };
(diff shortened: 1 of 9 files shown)


ui/modules/things/searchFilter.js
```diff
@@ -111,22 +109,98 @@ function onEnvironmentChanged() {
...
+      label: `thingId ~ ${FILTER_PLACEHOLDER}`,
+      rql: `like(thingId,"${FILTER_PLACEHOLDER}*")`,
+      group: 'ThingId',
+    },
+    {
+      label: `exists ${FILTER_PLACEHOLDER}`,
+      rql: `exists(${FILTER_PLACEHOLDER})`,
+      group: 'Other',
+    },
+    ...(Environments.current().filterList ?? []).map((f) => ({label: f, rql: f, group: 'Favourite'})),
```
thjaeckle, line 189: I think we mainly use American English in docs and UI .. so please change to "Favorite"

thjaeckle (approved):
Hi @thfries 
I just had a look at the added autocomplete functionality for search and I think it's great 👍 
Also very cool to have the "recent" searches available as suggestion ..

The functionality with the "fieldList" and the "filterList" is kind of hidden .. I wonder if we need to provide documentation for such very hidden features..
What do you think?

thfries:
Hi @thjaeckle,
thanks for the feedback.
I changed the spelling to AE like proposed and added a new commit.

The former menu of the dropdown button is inconsistent to the auto complete dropdown. I was thinking about making them consistent or to remove the dropdown button. The dropdown button is still nice, because otherwise you don't know what you can search for by typing. But the autocomplete could contain much more entries.

Agree! Documentation would be good. We can create additional documentation pages but some integrated online hints would be preferable I think. I can take a look how this could be done.

PR #1581, opened 2023-02-21
Creation of thing's policy is atomic with creation the of thing itself

alstanchev:
This PR adds atomic behaviour to the creation of thing and its policy.
In the rare case of failing to create a thing, its policy will be rolled back (deleted) in order to not fail create retry.


alstanchev:
@thjaeckle can you take a look when you can please.

thjaeckle:
@alstanchev will take a look soon.
Do you work on a unit test? That can probably only be tested with mocking and producing the "thing creation failure".

internal/utils/persistent-actors/src/main/java/org/eclipse/ditto/internal/utils/persistentactors/AbstractPersistenceSupervisor.java
```diff
@@ -347,6 +347,18 @@ protected CompletionStage<Object> modifyTargetActorCommandResponse(final Signal<
         return CompletableFuture.completedStage(persistenceCommandResponse);
     }
 
+
+    /**
+     * Hook for handling unexpected PersistenceActor exceptions before response is sent back to the SupervisorActor.
+     *
+     * @param error the error
+     * @param enforcedSignal the ditto headers from the initial command
+     */
```
thjaeckle, line 356: `@return` is missing from javadoc

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/RollbackCreatedPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.ditto.base.model.common.ResponseType;
+import org.eclipse.ditto.base.model.signals.Signal;
+import org.eclipse.ditto.things.model.signals.commands.modify.CreateThing;
+import org.eclipse.ditto.things.model.signals.commands.modify.CreateThingResponse;
+
+
+public record RollbackCreatedPolicy(Signal<?> initialCommand, Object response, CompletableFuture<Object> responseFuture) {
```
thjaeckle, line 24: Some javadoc for this record would be good - as it is public.

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/RollbackCreatedPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public static RollbackCreatedPolicy of(final Signal<?> command, final Object response,
```
thjaeckle, line 26: also I would suggest to add some javadoc here

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/RollbackCreatedPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+            final CompletableFuture<Object> responseFuture) {
+        return new RollbackCreatedPolicy(command, response, responseFuture);
+    }
+
+    public static boolean shouldRollback(final Signal<?> command, final Object response) {
+        return  command instanceof CreateThing && response instanceof CreateThingResponse thingResponse &&
+                thingResponse.getResponseType() == ResponseType.ERROR;
```
thjaeckle, line 33: I don't think that this will work.
`CreateThingResponse` is only created for the "success case".
In case of an error, you won't get a `CreateThingResponse`,but probably a `DittoRuntimeException`, more specifically a `ThingException`.
Or you would get a `ThingErrorResponse` (I currently don't remember where a DittoRuntimeException is converted to a `ThingErrorResponse`, but I think at a later point).

I wonder if you tested this functionality and whether it did work with this current implemetation?

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/RollbackCreatedPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public void completeInitialResponse() {
```
thjaeckle, line 36: Please add Javadoc

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingPolicyCreated.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.eclipse.ditto.things.service.enforcement;
+
+
+import org.eclipse.ditto.base.model.headers.DittoHeaders;
+import org.eclipse.ditto.policies.model.PolicyId;
+import org.eclipse.ditto.things.model.ThingId;
+
+
+public record ThingPolicyCreated(ThingId thingId, PolicyId policyId, DittoHeaders dittoHeaders) {
```
thjaeckle, line 22: Please provide Javadoc for public record

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingPolicyCreated.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public static ThingPolicyCreated of(final ThingId thingId, final PolicyId policyId, final DittoHeaders dittoHeaders) {
```
thjaeckle, line 24: Javadoc, please

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/ThingSupervisorActor.java
```diff
@@ -315,6 +329,36 @@ protected CompletionStage<Object> modifyTargetActorCommandResponse(final Signal<
...
+            getSelf().tell(RollbackCreatedPolicy.of(createThing, error, responseFuture), getSelf());
+            return responseFuture;
+        }
+        return CompletableFuture.failedFuture(error);
+    }
+
+    private void handlerRollbackCreatedPolicy(final RollbackCreatedPolicy rollback) {
+        if (policyCreatedEvent != null) {
+            DittoHeaders dittoHeaders = rollback.initialCommand().getDittoHeaders();
+            final DeletePolicy deletePolicy = DeletePolicy.of(policyCreatedEvent.policyId(), dittoHeaders.toBuilder()
```
thjaeckle, line 345: I have a bad feeling using the dittoHeaders of the original command here, as eg a 'If-Match' header could be present and preserved leading to strange outcomes.
I would build new Ditto headers and only copy certain origin headers like eg the correlation-id.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/ThingSupervisorActor.java
```diff
@@ -315,6 +329,36 @@ protected CompletionStage<Object> modifyTargetActorCommandResponse(final Signal<
...
+        if (enforcedSignal instanceof CreateThing createThing) {
+            CompletableFuture<Object> responseFuture = new CompletableFuture<>();
+            getSelf().tell(RollbackCreatedPolicy.of(createThing, error, responseFuture), getSelf());
+            return responseFuture;
+        }
+        return CompletableFuture.failedFuture(error);
+    }
+
+    private void handlerRollbackCreatedPolicy(final RollbackCreatedPolicy rollback) {
+        if (policyCreatedEvent != null) {
```
thjaeckle, line 343: After the if/else the field `policyCreatedEvent` must be set to null again. Otherwise it will always stay non null

thjaeckle (changes requested):
I added a first round of review comments.

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/RollbackCreatedPolicy.java
alstanchev: I was not sure here either that is why a wanted to hear your opinion on that. Maybe i should have left a comment about that.

thjaeckle (approved):
Hey @alstanchev 
That looks great - I had a look at the changes + the tests and are confident that improves resiliency a lot.
Good stuff 👍

PR #1584, opened 2023-02-23
UI - removed old dropdown for searchfilters

thfries:
Hi @thjaeckle,

While I was writing my last comment, I decided to:
- add the examples to the autocomplete search instead
- extend the search for the group, so you can search for "all examples" or "all fields"
- remove the old dropdown button, because it is now fully covered by the autocomplete search

Sorry, you were far faster than me on the other PR - so I created this new one

Regarding documentation: https://bootstraptour.com looks promising and very easy.


thjaeckle (approved):
LGTM, thx @thfries

PR #1586, opened 2023-02-25
Added section for exception handling to CONTRIBUTING.md

jufickel-b:
Added a hint to use the `Result` type where appropriate to enhance overall code safety bit by bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,27 @@ We use the [Google Java Style Guide](https://github.com/google/styleguide) where
 
 The only adjustment: use longer lines ("line split") with 120 characters instead of only 100.
 
+## Strive for code safety
+
+This section mainly targets development and refactoring at the *service side.*
+We know: exception handling is cumbersome.
+We also know: not handling an exception could be fatal.
+This is why we introduced the `Result` type which can be found in module
+`ditto-utils-result`.
+The idea is that for each method which returns a value and which could fail a
+`Result` should be returned instead of the plain value.
+This clearly indicates that the result should be checked for an error before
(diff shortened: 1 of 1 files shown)


thjaeckle:
In general, I think it is good to document the "best practices" we have for Ditto code. 👍 
I would also expect to find the established "de facto" best practices we already have, like e.g.:
* apply immutability whenever possible
* add unit tests for testing units
* add integration-tests (I think we call them now that, or?) for testing several components interacting with each other
* add system-tests for testing complete functionalities via APIs
* use Java records where already on Java17 compile level?
* `var` keyword vs. writing types?

I would find it strange to start documenting the "new" exception handling approach (which is done almost nowhere currently - so our experience with that is also not yet "solid") as Contribution guideline and leave all other well established "de facto" guidelines (we developed as a team over the years) not declared.
What do you think @jufickel-b ?

jufickel-b:
I think every bit of documentation is helpful. Since no documentation of best practices exists right now, I thought that the CONTRIBUTING.md was the right place. The new exception handling approach is almost nowhere applied because it is new. :-) And it will never be applied because no one knows about it. 

Appart from that, I agree that a documentation of the best practices sure would be nice.

thjaeckle:
The contributing.md is *the* resource to look at before doing a contribution to a GitHub repo.
So if we decide to put coding  patterns to follow for new code in there, I think we should list all of them.

We can do that on a branch (and in draft PR where we can discuss) and start collecting those..
I suggest to use the branch you created for that. 

I would just propose to delay merging to master until we have a rather full set of rules we agree on.
Just thinking that it would be quite random to have the new exception handling pattern as only rule on master while having not a single other established patten/rule listed.

Issue #1592, opened 2023-03-04
UI: Provide Bearer authentication for devops user


thjaeckle:
Ditto is capable of securing the devops API with OAuth (as alternative to basic auth), which is always favorable security wise.
The Ditto UI currently only provides Bearer token based authentication for the main user.
It should be quite easy to add Bearer based also for the devops user.

thjaeckle:
I can work on that..

Issue #1590, opened 2023-03-04
UI: Enhance Ditto-UI to configure log levels of Ditto


thjaeckle:
Ditto provides a "devops" API to retireve and [dynamically adjust log levels](https://www.eclipse.org/ditto/installation-operating.html#dynamically-adjust-log-levels) on the fly without restating Ditto.

This is an extremely helpful thing for admins/operators of Ditto to analyze problems in production.

I suggest to provide a new UI functionality to configure the log levels, in a new "Admin" menu item.
The credentials are the same as for managing connections, so authentication is already handled by the UI.

@thfries what do you think? Other "Admin functionality", eg bypassing policies for accessing the manged things, policies could also be added in the future.

thfries:
Sure, sounds not too complicated. I'm not used to these use cases but I can try to make a proposal.

thfries:
I'm working on that

PR #1594, opened 2023-03-06
[#1592] provide "Bearer" authentication for devops resources controlled via UI

thjaeckle:
* added radiobutton groups so that "Authorize" modal clearly shows/states which authentication to use where
* reduced to a single "Authorize" button in modal
* simplified environment a bit

Resolves: #1592

Screenshot of modal:
Screen Shot 2023-03-06 at 20 12 36 PM

### DIFF
--- a/ui/index.css
+++ b/ui/index.css
@@ -46,6 +46,10 @@ body {
     padding-top: 60px;
 }
 
+#authorizationModal label {
+    display: inline;
+}
+
 .toast-header {
     color: #842029;
     background-color: #f8d7da;
--- a/ui/modules/api.js
+++ b/ui/modules/api.js
@@ -277,18 +277,25 @@ let authHeaderValue;
  * @param {boolean} forDevOps if true, the credentials for the dev ops api will be used.
  */
 export function setAuthHeader(forDevOps) {
-  if (forDevOps && Environments.current().useBasicAuth) {
-    authHeaderKey = 'Authorization';
-    authHeaderValue = 'Basic ' + window.btoa(Environments.current().usernamePasswordDevOps);
-  } else if (Environments.current().useBasicAuth) {
-    authHeaderKey = 'Authorization';
-    authHeaderValue = 'Basic ' + window.btoa(Environments.current().usernamePassword);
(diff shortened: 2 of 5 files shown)


thjaeckle:
@thfries could you please take a look and check if this is also in your opinion good? :)

Right after creating the PR I thought that the "Authorize" modal should mabye group by main / devops credentials and not by the authentication type.
So:
1. Main user
a) basic auth
b) bearer auth
c) pre-authenticated auth
2. Devops user
a) basic auth
b) bearer auth

What do you think?

thjaeckle:
I would plan the Ditto release 3.2.0 for tomorrow.
An approval of this PR would be nice to make it into the release.
/cc @thfries @alstanchev @jufickel-b

thfries:
The second approach is better. In the first case, the radio buttons do not fit, because you can work with both users in parallel (it is not either / or, you can enter both).

The separation of DevOps API and "Main" API is a bit confusing. There are two variants to retrieve the connections. The DevOps user is currently only needed for the "old" Ditto 2 way to retrieve connections.
So I think the separate authorize method for the DevOps user makes more sense for the new feature in #1590

Do you think we should remove the Ditto 2 support for connections?

thjaeckle:
> The separation of DevOps API and \"Main\" API is a bit confusing. There are two variants to retrieve the connections. The DevOps user is currently only needed for the \"old\" Ditto 2 way to retrieve connections.

Hm, actually no. There are 2 variants to retrieve connections, yes.
However both require the "DevOps authentication" (either via basic auth or OAuth, depending on how Ditto is configured).
The "main user" can never access connections in Ditto.

I think the proposed PR does correctly reflect the current possibilities.

Also no need to drop Ditto 2 support.

thfries:
Hi @thjaeckle,
I tested and reviewed and it looks good to me.
(in case of Things service, you need to enter the same bearer twice now? but it works)

Thanks!

thjaeckle:
Thanks @thfries .

Ah, yes. For Bosch IoT things the end user can have access to the connections.
The Bearer tokens could still differ of course.
The solution Admin does not have automatically access to things for example.

PR #1600, opened 2023-03-13
UI - New Operations tab

### LINKED ISSUE #1590: UI: Enhance Ditto-UI to configure log levels of Ditto

thfries:
please find this PR resolves #1590
* added new tab
* changed way to control right auth user
* extended api to allow devops path
* utils to add table row now returns row
* crud editor returns explicit cancel action
* crud editor now with option without delete or create
image

Please let me know if this is the right direction.

Pending topics:
* setting log level for all services is currently not possible
* I m thinking if we should allow to hide unused tabs by configuration (e.g. hide connections and operations for sandbox)
* Need to add validation to avoid invalid json to be used on update

Best regards, Thomas

### DIFF
--- a/ui/modules/operations/operations.html
+++ b/ui/modules/operations/operations.html
@@ -1,5 +1,5 @@
 <!--
-  ~ Copyright (c) 2022 Contributors to the Eclipse Foundation
+  ~ Copyright (c) 2023 Contributors to the Eclipse Foundation
   ~
   ~ See the NOTICE file(s) distributed with this work for additional
   ~ information regarding copyright ownership.
--- a/ui/modules/operations/operations.html
+++ b/ui/modules/operations/operations.html
@@ -1,0 +1,50 @@
+<!--
+  ~ Copyright (c) 2022 Contributors to the Eclipse Foundation
+  ~
+  ~ See the NOTICE file(s) distributed with this work for additional
+  ~ information regarding copyright ownership.
+  ~
+  ~ This program and the accompanying materials are made available under the
+  ~ terms of the Eclipse Public License 2.0 which is available at
+  ~ http://www.eclipse.org/legal/epl-2.0
+  ~
+  ~ SPDX-License-Identifier: EPL-2.0
+  -->
+<h5>Ditto Services Logging</h5>
+<hr />
+<div class="row resizable_pane" style="height: 80vh">
+    <div class="col-md-6 resizable_flex_column">
+        <div class="input-group input-group-sm mt-1 mb-1" role="group">
+            <div style="flex-grow: 1;"></div>
+            <button class="btn btn-outline-primary btn-sm button_round_left" id="buttonLoadAllLogLevels">
+                <i class="bi bi-arrow-clockwise"></i>
+                Refresh
+            </button>
+        </div>
+        <div class="input-group has-validation">
+            <input class="form-control" id="tableValidationOpsLogging" hidden="true"></input>
+            <div class="invalid-feedback"></div>
+        </div>
+        <div class="table-wrap">
+            <table class="table table-striped table-hover table-sm">
+                <thead>
+                    <tr>
+                        <th>Service</th>
+                        <th>Level</th>
+                        <th>Logger</th>
+                    </tr>
+                </thead>
+                <tbody id="tbodyOpsLogging"></tbody>
+            </table>
+        </div>
+    </div>
+    <div class="col-md-6 resizable_flex_column">
+        <crud-toolbar label="Service Logging" id="crudOpsLoggingJson" style="flex-grow: 1;" disableCreate disableDelete>
+            <div class="ace_container" style="flex-grow: 1;">
+                <div>
+                    <div class="script_editor" id="opsLoggingEditor"></div>
+                </div>
+            </div>
+        </crud-toolbar>
+    <div>
+</div>
--- a/ui/modules/operations/operations.html
+++ b/ui/modules/operations/operations.html
@@ -13,7 +13,7 @@
 <h5>Ditto Services Logging</h5>
 <hr />
 <div class="row resizable_pane" style="height: 80vh">
-    <div class="col-md-6 resizable_flex_column">
+    <div class="col-md-12 resizable_flex_column">
         <div class="input-group input-group-sm mt-1 mb-1" role="group">
             <div style="flex-grow: 1;"></div>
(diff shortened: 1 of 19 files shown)


thjaeckle:
Hi @thfries 
Thanks for the fast first draft 👍 

I thought that the different available log levels for each logger could be rendered with a [bootstrap button group](https://getbootstrap.com/docs/4.0/components/button-group/), e.g. also having different colors - DEBUG maybe light gray, INFO green, WARN yellow and ERROR red.
And the currently active log level could be highlighted in order to see at once glance what is currently configured.

That way, the buttons could show the currently configured log level and could also be used to directly change it.
Table + Editor for this use case seem a little bit too complicated, in my opinion.

thfries:
Hey @thjaeckle,
thanks for your proposal. Usability is indeed far better and the code is even simpler as before. This is how it looks like now:
image

I'm not sure about the refresh button: It should not scroll away or may be it is not needed at all?

Best regards, Thomas

thfries:
added another commit to keep the refresh button visible...

thjaeckle:
@thfries yes, that looks really great. :+1:
Will take a look at the code soon.

I think the api supports changing the log level of e.g. all connectivity instances at once and even of specific instances separately.
In my experience, only changing the levels of all instances at once is required.

ui/modules/operations/operations.html
```diff
@@ -0,0 +1,49 @@
+<!--
+  ~ Copyright (c) 2022 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: Please change to creation year 2023

ui/modules/operations/operations.js
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: Please change to creation year 2023

thjaeckle (changes requested):
Hi @thfries 
I just reviewed - looks good 👍 
Simple and adds a lot benefit for operating Ditto.

Just 2 newly created files have the wrong copyright year, I mentioned that inline.

Apart from that this would be good for me to merge.

thjaeckle:
@thfries I totally forgot .. There is another thing to do which would this make even more useful:

There should be for each service (e.g. "things", "gateway") a textfield in which you can enter a new arbitrary logger and also select a log level.
The API supports to add new loggers in the fly - that is e.g. very useful if you just need to have DEBUG logs for a single class - then you just add a logger on-the fly with DEBUG - and e.g. deactivate again after you have collected some DEBUG logs.

And they can even be removed again by setting the log level to "OFF" - which we could also think about.

thfries:
Hi @thjaeckle ,
thanks for your feedback and your thoughts.
Review comments are fixed.
Regarding your other comments:
* I found no API call to set all loggers of one service at once. But in general we can add a "select all" button on top of all columns to do that. If we need several API calls this would still be possible
* There is an API call to set the level for one logger for all services => therefor we would need a checkbox "apply for all services" on each row?
* adding and deleting loggers sounds interesting, too. If this is a relevant use case, we could add it.

The first one sounds very valuable to still add.
...and I still would like to add a feature to hide unused tabs by the environment definition 😉

If you like, I can add some more improvements to this PR.
I will remove draft state of the PR, so you can decide to merge... as you like

Thank you, @thjaeckle

thfries:
Hi @thjaeckle,
added the toggles for the main menu and found a bug that the auth header was keeping active after switching an environment.

thjaeckle:
> * I found no API call to set all loggers of one service at once. But in general we can add a "select all" button on top of all columns to do that. If we need several API calls this would still be possible

Your proposal already uses this API to switch all loggers of one service at once - so that works as intended 👍 

> * There is an API call to set the level for one logger for all services => therefor we would need a checkbox "apply for all services" on each row?

I don't think that this would be needed.

> * adding and deleting loggers sounds interesting, too. If this is a relevant use case, we could add it.

Yes, this is really valuable, as you can e.g. define for a specific class or package a new logger config without that it needs to be "visible" before in the logging overview.
I would prefer to already have it included in this PR - I could also try to contribute that if I find time ..

Please share your thoughts on that :)

ui/modules/operations/operations.js
```diff
@@ -73,5 +73,18 @@ function createLoggerView(allLogLevels) {
...
       dom.divLoggers.append(row);
     });
+    let newLoggerRow = document.createElement('div');
+    newLoggerRow.attachShadow({mode: 'open'});
+    newLoggerRow.shadowRoot.append(dom.templateLogger.content.cloneNode(true));
+    let inputLoggerElement = newLoggerRow.shadowRoot.getElementById('inputLogger');
+    inputLoggerElement.disabled = false;
+    inputLoggerElement.placeholder = 'Add new logger name and choose log level';
+    Array.from(newLoggerRow.shadowRoot.querySelectorAll('.btn-check')).forEach((btn) => {
+      btn.addEventListener('click', (event) => onUpdateLoggingClick(service, {
```
thfries, line 83: "service" is then part of the anonymous function (?) Above I put "service" as an attribute to the button explicitly. I would make that consistently

thfries:
Hi @thjaeckle ,

nice, thank you!
Yes I also observed issues with the bearer but I was not yet able to reproduce it.
Your addition looks nice. I made one comment and I already thought about to split into smaller local functions to improve readability a bit (I think this is best practice). Let me give it a try.

ui/modules/operations/operations.js
thjaeckle: Sounds good. I did not manage to do it the same as you did. For some reasons, that did not work.

thfries:
Done. I like your approach. It keeps the needed parameters just where they are needed. Works also for the other buttons. Strange that it didn't work with the event target...

thjaeckle (approved):
Lgtm

PR #1601, opened 2023-03-14
fix reading configured WoT TD "json template" from system property

thjaeckle:
* preserving the object structure

The problem currently is that when e.g. configuring the "json template" to add to a TD using system properties, like:
```
-Dditto.things.wot.to-thing-description.json-template="{\"security\":\"basic_sc\",\"securityDefinitions\":{\"basic_sc\":{\"in\":\"header\",\"scheme\":\"basic\"}},\"support\":\"https://www.eclipse.org/ditto/\"}"
```
This can't be parsed as a surrounding quote is inserted and the complete string is interpreted as JSON string.

Due to the nature of HOCON config I did not see another way other than stripping the `"` from e.g. a configured object: `"{}"` and also un-escaping all contained escaped strings.

### DIFF
--- a/documentation/src/main/resources/openapi/ditto-api-2.yml
+++ b/documentation/src/main/resources/openapi/ditto-api-2.yml
@@ -6822,6 +6822,7 @@ paths:
       description: Returns all connections.
       security:
         - DevOpsBasic: []
+        - DevOpsBearer: []
       tags:
         - Connections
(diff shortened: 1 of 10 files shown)


alstanchev (approved):
LGTM 👍

alstanchev (approved):
LGTM 👍

PR #1602, opened 2023-03-19
Fixed filtering desired properties signals. #1599

### LINKED ISSUE #1599: Change notification filtering doesn't work for desired properties changes
Hello. I'm trying to setup Kafka connection with target using following topic filter: `_/_/things/twin/events?extraFields=features/{{feature:id}}&filter=like(resource:path,'/features/RemoteAccessConnections*')`. It works ok for feature change events by changing whole RemoteAccessConnections feature on the thing. But no changes of desired properties are propagated to Kafka (e.g. by `PUT /api/2/things/test:device/features/RemoteAccessConnections/desiredProperties`). 

I did some debugging and came to conclusion that problem is in class [ThingEventToThingConverter](https://github.com/eclipse-ditto/ditto/blob/master/things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingEventToThingConverter.java#L111) because there is no mapping for FeatureDesiredPropertiesModified (an other) event defined and as a consequence signal filtering fails in this [method](https://github.com/eclipse-ditto/ditto/blob/master/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/SignalFilter.java#L178) for such events.

Am I doing something wrong or is it intended behavior? In this [article ](https://www.eclipse.org/ditto/2020-11-11-desired-properties.html) it is explicitly said that enrichment and filtering should be supported also for desired properties so I would consider it as a bug.

If it is a bug I would be happy to help with solving the issue.

kyberpunk:
This PR is fixing a bug which caused that signal filtering did not work for desired properties changes. There was missing implementation of mapping to Thing model for such events. More details described in #1599.

I've added few unit tests covering new code and the bug itself. Also tested it manually with my K8s deployment with Kafka and it works as expected now.

I didn't want to change a code too much but maybe would worth to move the conversion function to ThingEvent interface an use polymorphism; so no one will forget to implement it anymore in case of new events.

fixes #1599

### DIFF
--- a/connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/TestConstants.java
+++ b/connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/TestConstants.java
@@ -127,9 +127,11 @@ import org.eclipse.ditto.protocol.ProtocolFactory;
 import org.eclipse.ditto.protocol.TopicPath;
 import org.eclipse.ditto.protocol.adapter.DittoProtocolAdapter;
 import org.eclipse.ditto.things.model.Attributes;
+import org.eclipse.ditto.things.model.FeatureProperties;
 import org.eclipse.ditto.things.model.Thing;
 import org.eclipse.ditto.things.model.ThingId;
 import org.eclipse.ditto.things.model.signals.commands.modify.ModifyThing;
+import org.eclipse.ditto.things.model.signals.events.FeatureDesiredPropertiesModified;
 import org.eclipse.ditto.things.model.signals.events.ThingModified;
 import org.eclipse.ditto.things.model.signals.events.ThingModifiedEvent;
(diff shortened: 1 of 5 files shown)


kyberpunk:
I've did wrong by mixing my test with parametrized tests. Will fix it.

kyberpunk:
I've did wrong by mixing my test with parametrized tests. Will fix it.

thjaeckle (approved):
Thanks so much @kyberpunk 

I reviewed and your changes look great 👍 

> I didn't want to change a code too much but maybe would worth to move the conversion function to ThingEvent interface an use polymorphism; so no one will forget to implement it anymore in case of new events.

Indeed, that would have been and would still be way more cleaner to do it that way.
Noted for the next model changes ;)

PR #1604, opened 2023-03-22
don't use WARN log level when no policy is rolled back

thjaeckle:
I encountered a lot of "WARN" logs in the Ditto 3.2.0 release cause by the new "rollback policy on thing exception error during thing creation" stabilization done in #1581
I think a "WARNING" should only be logged if the policy is actually rolled back (deleted again).

Currently, it is always logged if e.g. the following 2 exception happen during thing creation:
* ThingConflictException (thing already present) - nothing to roll back
* ThingPreconditionFailedException - this is sent back when eg `If-Match` or `If-None-Match` headers were set to do a "conditional create" only if the thing does not yet exist

For those cases, I would not like to see a WARNING in the logs.
So I changed the log level to info instead and added another WARNING log when the policy is **actually** rolled back.


alstanchev:
I would even change them to debug and leave info only if the policy is deleted but I will add it with a later commit when I finish with the caching problem. For now it LGTM:)

Issue #1619, opened 2023-04-18
Connection metrics for "target only" connections can be reported+logged wrongly for message dropping


thjaeckle:
Based on the following [Stackoverflow question](https://stackoverflow.com/questions/76044974/eclipse-ditto-does-not-send-all-things-events-over-target-connection/76046710#76046710) I found out that in the `OutboundMappingProcessorActor` a wrong `monitor` is invoked for reporting failures:

https://github.com/eclipse-ditto/ditto/blob/a5862aad7decb8b9ea4187010af6a1b6d26b1430/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java#L352-L360

The `responseDispatchedMonitor` is used - however this method is also called for "target only" connections for outbound signals.  
Those are not a response (reported as the `"address": "_responses"`), but should be reported to the `getMonitorsForOutboundSignal()`:

https://github.com/eclipse-ditto/ditto/blob/a5862aad7decb8b9ea4187010af6a1b6d26b1430/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java#L679

It even seems that the metrics for e.g. the `"Message is dropped as a result of backpressure strategy!"` situation are counted and logged twice for the `dispatched` type:
* once as a `success` (with correct `target` address)
* once as a `failed` (with the wrong `"_responses"` address)

That must also be analyzed, it should per category only be counted once.

Check the provided target logs and search for the correlationId `6c73e2e6-defa-42fd-a7c4-53a3d8f6e2de` in order to reproduce:
https://pastebin.com/VSApKH0g

@desislava-marinova you were the author of this `messageDiscarded` method using always the `responseDispatchedMonitor`.
Do you still remember (already 9 months ago I fear :D) that addition?

PR #1620, opened 2023-04-18
#1619 adjusted monitor to use in OutboundMappingProcessorActor

thjaeckle:
* to not use `responseDispatchedMonitor`, but getMonitorsForOutboundSignal() with using `responseMappedMonitor` instead for responses

Fixes: #1619


thjaeckle:
@alstanchev could you take a look as reviewer? 
Is only a small fix, fixing the category of outbound mapped messages in case they are dropped - which can happen for both responses, but also "normal" outbound messages.

alstanchev (approved):
LGTM :)

PR #1641, opened 2023-05-24
Improve DittoProtocol `MessagePath` to be aware of message subject

thjaeckle:
* this was very cumbersome to determine before when using the `MessagePath` API
* also added `boolean isInboxOutboxMessage()` for determining if the message path represents an inbox/outbox message


kalinkostashki (approved):
LGTM :)

Issue #1644, opened 2023-05-29
Thing API does not support alternative way of specifying list query parameters


enenuki:
Ditto thing API supports querying for certain parts of thing, by specifying URL like:
`<BASE_URL>/api/2/things/<thingid>?fields=attributes/manufacturer,attributes/type,attributes/productkey`
 
but if we want to use alternative format:
`<BASE_URL>/api/2/things/<thingid>?fields=attributes/manufacturer&fields=attributes/type&fields=attributes/productkey`

it only returns first attribute (in this case attributes/manufacturer).

While this may seem like minor issue, some libraries (jawax.ws.rs.QueryParam in JAX-RS API) use only the alternative format to serialize their requests, which makes it a bit harder to bypass.

![Screenshot from 2023-05-24 16-45-05](https://github.com/eclipse-ditto/ditto/assets/752310/aeff0e6e-a370-41c0-a111-84ce0fabae30)
![Screenshot from 2023-05-24 16-45-18](https://github.com/eclipse-ditto/ditto/assets/752310/9d3e4ad5-97fb-44a7-81c5-22f90842f5dd)

Ditto version: 3.2.0

thjaeckle:
Thanks for reporting, @enenuki
That definitely would be a good and simple enhancement for Ditto.
Will work on that..

PR #1652, opened 2023-06-12
#1644 support query params in the style where not comma separation is used, but duplicated keys

thjaeckle:
* e.g.: `?fields=thingId&fields=policyId`

Resolves: #1644


kalinkostashki (approved):
LGTM :)

PR #1657, opened 2023-06-15
Fix that redeliveries for acknowledgeable connectivity messages were issued to often

thjaeckle:
* removed generic `acknowledgeableMessage.reject(true)` on encountered exceptions
* the logic where the status code of the DittoRuntimeException is handled did not even take a part in the redelivery


alstanchev:
Sure i'll look in to it right away

kalinkostashki (approved):
LGTM :)

Issue #1649, opened 2023-06-08
[Kubernetes,Gateway]: AskTimeoutException: Ask timed out on [Actor[...]] after [25000 ms]. Message of type [akka.cluster.ddata.Replicator$Update]


Altair-Bueno:
### Description

Due to a node crash on my Kubernetes cluster, some pods from Ditto were restarted. These included the NGINX proxy, Ditto Gateway and Ditto Policies. However, NGINX refused to start (init state).

Gateway shows multiple entries with the following contents:

```text
2023-06-08 11:39:08,649 ERROR [] o.e.d.i.u.p.a.SubUpdater akka://ditto-cluster/user/policy-announcement-aware-sub-supervisor/subUpdater2 - Failure updating Ditto pub/sub subscription - trying again next clock tick
java.util.concurrent.CompletionException: akka.pattern.AskTimeoutException: Ask timed out on [Actor[akka://ditto-cluster/user/policy-announcement-aware-replicator#1986903194]] after [25000 ms]. Message of type [akka.cluster.ddata.Replicator$Update]. A typical reason for `AskTimeoutException` is that the recipient actor didn't send a reply.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:636)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:29)
	at scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:26)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)
	at scala.concurrent.ExecutionContext$parasitic$.execute(ExecutionContext.scala:222)
	at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
	at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:335)
	at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
[...]

PR #1661, opened 2023-06-20
#1649 remove network policies from Ditto Helm Chart

thjaeckle:
Same as the conclusion drawn in #1649, I think it is dangerous for the Akka cluster to define networkPolicies, which prevents Akka Management to access the Kubernetes API in order to "heal itself".

This PR removes the network policies from the Ditto Helm Chart.

Fixes: #1649


thjaeckle:
@kalinkostashki @alstanchev for Ditto 3.3.0 this would also be good to review

kalinkostashki (approved):
LGTM :)

PR #1665, opened 2023-06-23
add CPU limits for nginx based deployments for Helm chart

thjaeckle:
The Helm chart lacks to define "limits" for the nginx deployments.
That could cause for machines with a a lot of available CPUs that nginx would create a lot of worker threads, resuling in much memory needs.

Applied the configured "request" also to be used as "limit".


thjaeckle:
Discarding the approach, as [defining CPU limits is not a good approach](https://home.robusta.dev/blog/stop-using-cpu-limits)

PR #1666, opened 2023-06-26
Fix blocking Akka dispatcher thread on WoT skeleton creation

thjaeckle:
During creating of Things and Features with a given WoT "ThingModel" `definition`, an API was used which uses `CompletableFuture` and which does blocking calls (e.g. in order to retrieve the model(s) via HTTP).

This however was directly called via the Akka "dispatcher" thread - so other processing was completely halted, even the Actor system of the `things ` node could be affected by not being able to schedule any actor related work.

This bugfix enables the internal mechanisms of Ditto to be also able to deal with asynchronously provided `event`s to persist - which will offload the blocking operations to the already provided `"wot-dispatcher"` instead.

### DIFF
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/events/PolicyActionEvent.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/events/PolicyActionEvent.java
@@ -28,5 +28,5 @@ public interface PolicyActionEvent<T extends PolicyActionEvent<T>> extends Polic
      * @param otherPolicyActionEvents the collection of policy action events.
      * @return the aggregated event.
      */
-    PolicyEvent<?> aggregateWith(Collection<PolicyActionEvent<?>> otherPolicyActionEvents);
+    PolicyEvent<?> aggregateWith(Collection<PolicyActionEvent> otherPolicyActionEvents);
 }
--- a/wot/integration/src/main/java/org/eclipse/ditto/wot/integration/provider/DefaultWotThingDescriptionProvider.java
+++ b/wot/integration/src/main/java/org/eclipse/ditto/wot/integration/provider/DefaultWotThingDescriptionProvider.java
@@ -17,7 +17,9 @@ import static org.eclipse.ditto.base.model.common.ConditionChecker.checkNotNull;
 import java.net.URL;
 import java.time.Duration;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 
@@ -119,7 +121,7 @@ final class DefaultWotThingDescriptionProvider implements WotThingDescriptionPro
     }
 
     @Override
-    public Optional<Thing> provideThingSkeletonForCreation(final ThingId thingId,
+    public CompletionStage<Optional<Thing>> provideThingSkeletonForCreation(final ThingId thingId,
             @Nullable final ThingDefinition thingDefinition,
             final DittoHeaders dittoHeaders) {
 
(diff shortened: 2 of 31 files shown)


thjaeckle:
@kalinkostashki @alstanchev this is quite important for me.
Could you squeeze in a review and a system test run?

alstanchev:
Sure will do my best. Will look in to it later today.

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/events/PolicyActionEvent.java
```diff
@@ -28,5 +28,5 @@ public interface PolicyActionEvent<T extends PolicyActionEvent<T>> extends Polic
      * @param otherPolicyActionEvents the collection of policy action events.
      * @return the aggregated event.
      */
-    PolicyEvent<?> aggregateWith(Collection<PolicyActionEvent<?>> otherPolicyActionEvents);
+    PolicyEvent<?> aggregateWith(Collection<PolicyActionEvent> otherPolicyActionEvents);
```
alstanchev, line 31: I didn't quite got the idea of removing the wildcard parameter type here. What is the reason for that?

alstanchev (commented):
I have left few comments in the code other then that looks good.
Tests are running currently had one failure in the Things unit tests which i believe is from the thing i have commented. For the sake of time i tried a local fix and am currently waiting on the builds.

wot/integration/src/main/java/org/eclipse/ditto/wot/integration/provider/DefaultWotThingDescriptionProvider.java
```diff
@@ -126,28 +135,31 @@ public Optional<Thing> provideThingSkeletonForCreation(final ThingId thingId,
...
+                    return thingModelFetcher.fetchThingModel(url, dittoHeaders)
+                            .thenApplyAsync(thingModel -> thingSkeletonGenerator
+                                            .generateThingSkeleton(thingId, thingModel, url, dittoHeaders),
+                                    executor
+                            )
+                            .thenApply(thingSkeleton -> {
+                                logger.debug("Created Thing skeleton for new Thing with id <{}>: <{}>", thingId,
+                                        thingSkeleton);
+                                return thingSkeleton;
+                            });
```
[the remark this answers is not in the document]
thjaeckle: Very good catch @alstanchev - totally right and this even makes the code nicer, thanks - will adjust 👍

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/events/PolicyActionEvent.java
thjaeckle: Yes, I "had to" do this as I had otherwise compile errors all over the place when adding the `CompletionStage` overloaded methods to e.g. `Result` and `ResultVisitor`.
So not really "nicer" that way, however I did not see another way.

thjaeckle:
@alstanchev I am still working on this as I encountered some more `.join()` calls which were done on the Akka main dispatcher thread.
I tried to replace them all with `CompletionStage` based APIs, I am however stuck when persisting in the `AbstractPersistenceActor` - there I have an open "TODO" to replace the explicit blocking.

Without that however the event is not persisted correctly (in case of WoT based Thing creations) ..

So still working on it :)

thjaeckle:
@alstanchev I am confident that I solved it now ..

I reduced the potential blocking on the Akka dispatcher thread to a minimum - as I did not manage to get the "persist" to work without it.

Let me sleep one night, maybe I have a better idea tomorrow :)

thjaeckle:
@alstanchev this morning I got rid of blocking on the Akka dispatcher thread completely 😁 
With that I would say it is ready to be reviewed.

Sorry that it is that many changes, I completely made the WoT integration async - I don't know why I did not do that when creating it in the first place.. :/

thjaeckle:
@alstanchev there is one unit test failing on GitHub action build - locally however this work both when starting in the IDE and when building via maven.

As you are the author of this test, maybe do you have a clue why it would fail?
`ThingPersistenceActorTest.policyShouldNotBeDeletedOnThingRetrieveAndActorFail`

One hint: I changed something in the "rollback created policy logic when thing creation fails" as this was IMO not done correctly:
https://github.com/eclipse-ditto/ditto/blob/7f96a643bbe0b81f3f06ab15f97d654da36c78da/internal/utils/persistent-actors/src/main/java/org/eclipse/ditto/internal/utils/persistentactors/AbstractPersistenceSupervisor.java#L786

Here I replaced `signal` with `transformed` - as an incoming `ModifyThing` command could be transformed to a `CreateThing` by the `handleOptionalTransformationException` method when the Thing did not yet exist before.
In such a case, the rollback of the policy was not working..

As the unit test tests something in that direction, maybe this is related?

alstanchev (approved):
LGTM ;)

Issue #1663, opened 2023-06-23
[Helm,Nginx,dittoui]: OOMKilled on Ditto 3.3.0


Altair-Bueno:
```yml
# values.yml
global:
  prometheus:
    enabled: true

gateway:
  devopsPassword: foobar

mongodb:
  persistence:
    enabled: false
``` 

```sh
helm install --create-namespace dt-ditto oci://registry-1.docker.io/eclipse/ditto -f values.yml --version=3.3.0
```

```text
dt-ditto-connectivity-56f78b486d-8r692   1/1     Running            0          2m10s
dt-ditto-dittoui-9f875458c-fkcq7         0/1     CrashLoopBackOff   4          2m10s
dt-ditto-gateway-565d4f4f6c-4z7tp        1/1     Running            0          2m10s
dt-ditto-mongodb-df4ffd99d-mps8x         1/1     Running            0          2m10s
dt-ditto-nginx-566ff8d696-c75dg          0/1     CrashLoopBackOff   3          2m10s
dt-ditto-policies-657c989997-p88v2       1/1     Running            0          2m10s
dt-ditto-swaggerui-77f4864955-2vfwm      1/1     Running            0          2m10s
dt-ditto-things-5886d99c7b-tgjn8         1/1     Running            0          2m10s
dt-ditto-thingssearch-654556bb8b-ml9hn   1/1     Running            0          2m10s
```

Changing the `nginx.resources.memoryMi` to 500 stops the pod from crashing. However, as an user of the Helm chart I would expect a sensible default ([currently is set to only 64 megabytes!!](https://github.com/eclipse-ditto/ditto/blob/2157ca59ea70caabf80140133cb8f367f968ae37/deployment/helm/ditto/values.yaml#L1399))

<details><summary>Logs nginx</summary>
<p>

```text
Defaulted container "ditto-nginx" out of: ditto-nginx, wait-for-gateway (init)
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
[...]

PR #1667, opened 2023-06-26
#1663 fix that nginx's worker_processes setting 'auto' causes problems

thjaeckle:
… when deploying Helm chart to worker with many CPUs

* configure the "dittoui"'s nginx to use 1 worker process (it only serves static content)
* add configuration option in values.yaml to configure the nginx's used worker_processes and default to 4

Fixes: #1663


kalinkostashki (approved):
LGTM :)

PR #1668, opened 2023-06-26
Support for ingress controller

vbricentic:
Support for ingress controller
Creation of ingress-nginx namespace


thjaeckle:
Thanks @vbricentic for this contribution.
Could you please sign your commit with the e-mail-address registered to your Eclipse account (and where you have signed the ECA)?
See also [CONTRIBUTING.md](https://github.com/eclipse-ditto/ditto/blob/master/CONTRIBUTING.md#legal-requirements)

vbricentic:
Thanks for the suggestion @thjaeckle . I have just signed the commit and eclipsefdn/eca. Is there a way to re-trigger the check or I should maybe recreate PR?

thjaeckle:
> Thanks for the suggestion @thjaeckle . I have just signed the commit and eclipsefdn/eca. Is there a way to re-trigger the check or I should maybe recreate PR?

It seems something is still off .. see the Eclipse ECA tool:
image

Did you configure your git email-address correctly to the one used in your eclipse account?

In addition, other checks fail as well:
* linting of the chart (the chart version needs to be increased - see the "Details")
* license headers are missing from files (see the "Details")

vbricentic:
It should be better now, but not sure if everything is now solved. @thjaeckle could you please re-trigger check again.

brvbosch:
Account issue finally resolved.

@thjaeckle I'm little bit confused with chart versioning. Lint and test Helm chart requires to change. In code it says it's effectively set by release-job. Version of helm currently matches application version, so I tried with 3.3.0.1 but the format was rejected. I set now to 3.3.1 but might be misleading.

thjaeckle:
@vbricentic The chart version is still invalid:
```
Error: validation: chart.metadata.version "3.3.0.1" is invalid
```

brvbosch:
@thjaeckle in that case it should be now fine, I set it to 3.3.2 to not conflict with #1667

deployment/helm/ditto/templates/nginx-ingress.yaml
```diff
@@ -0,0 +1,843 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nginx-node-health-check-conf
+  namespace: "{{ .Values.ingress.namespace }}"
+data:
+  nginx-node-health-check.conf: |
+    # config for health check container running together with the ingress controller
+    worker_processes auto; #auto-detect number of logical CPU cores
```
thjaeckle, line 28: If this nginx is only used for healh-check, I would limit the `worker_processes` to `1` instead of using `auto` which will use the amount of available CPU cores of the worker node it runs on as value.

deployment/helm/ditto/values.yaml
```diff
@@ -169,8 +169,15 @@ dbconfig:
 ingress:
   # enabled whether Ingress should be enabled as alternative to the contained nginx
   enabled: false
+  # namespace for ingress controller, must exist
+  namespace: ingress-nginx
+  # Ingress-NGINX version
+  nginxIngressVersion: v1.8.0
+  # Nginx Version
+  nginxVersion: "1.21.6"
```
thjaeckle, line 177: As suggested already, please move that under a new `controller` section as part of the `ingress`.
And add a flag whether to deploy the ingress-controller or not - by default it is probably ok to deploy it when `ingress.enabled = true`.

thjaeckle, line 177: This `nginxVersion` is already quite old.
Can we use a recent one?

deployment/helm/ditto/values.yaml
```diff
@@ -169,8 +169,15 @@ dbconfig:
...
+
   # className is the 'ingressClassName' to configure in the Ingress spec
-  className: nginx
+  className: ingress-nginx
```
thjaeckle, line 180: Is this an officially supported Ingress `className`? 
I did quickly search for it and only found that the nginx-ingress-controller uses `nginx` (as it currently also is configured).

thjaeckle (changes requested):
Hey @vbricentic 
I did a first round of reviews and added some comments.

Please move the ingress "controller" specific config to a separate YAML "object" - and make it configurable that the Ingress controller is deployed or not (if Ingress is enabled).

deployment/helm/ditto/values.yaml
brvbosch: I followed the supported version table from https://github.com/kubernetes/ingress-nginx but probably a newer one should also work fine

deployment/helm/ditto/templates/nginx-ingress.yaml
brvbosch: it is health check and there is no need for more then 1 worker process

deployment/helm/ditto/values.yaml
thjaeckle: > I followed the supported version table from https://github.com/kubernetes/ingress-nginx but probably a newer one should also work fine

Ah, I see .. then I would also stick to that version 👍

deployment/helm/ditto/templates/nginx-ingress.yaml
```diff
@@ -0,0 +1,843 @@
...
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+{{- if .Values.ingress.enabled -}}
```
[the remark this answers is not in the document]
brvbosch: done

deployment/helm/ditto/templates/nginx-ingress.yaml
```diff
@@ -0,0 +1,843 @@
...
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "{{ .Values.ingress.namespace }}"
```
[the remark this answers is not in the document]
brvbosch: done

deployment/helm/ditto/templates/nginx-ingress.yaml
```diff
@@ -0,0 +1,843 @@
...
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: "{{ .Values.ingress.namespace }}"
+    app.kubernetes.io/name: "{{ .Values.ingress.namespace }}"
+    app.kubernetes.io/part-of: "{{ .Values.ingress.namespace }}"
+    app.kubernetes.io/version: "{{ .Values.ingress.nginxIngressVersion }}"
```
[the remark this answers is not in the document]
brvbosch: done

deployment/helm/ditto/templates/nginx-ingress.yaml
brvbosch: done

deployment/helm/ditto/values.yaml
brvbosch: The configuration moved to controller section.

brvbosch: Ok, the suggested version than stays as default. The version table link is now included as description for the variables.

deployment/helm/ditto/values.yaml
brvbosch: Currently `ingress.className` must match `ingress.controller.namespace` to ingress really working, which I guess is not the best approach. They should probably be independent parameters, so I'm trying to separate them.

brvbosch: Fixed. Two independent parameters. Default className is nginx.

brvbosch: Actually changing controller class name not working at the moment. Reverted back.

brvbosch:
version bumped again

deployment/helm/ditto/values.yaml
brvbosch: Hm, different namespaces are now also working, looks like a glitch in k8s :confused:

Issue #1670, opened 2023-06-27
[Helm,Gateway] devopsPassword and statusPassword wrong values.yaml path


mladBlum:
In templates/gateway-secret.yaml the devopsPassword and statusPassword are read from `.Values.gateway.devopsPassword` and `.Values.gateway.statusPassword`.

In the deployment/helm/ditto/values.yaml these two values are located at `gateway.config.devops.devopsPassword` and `gateway.config.devops.statusPassword`

As a result, the two passwords are always regenerated and the passed passwords are not used.

thjaeckle:
Thanks for the hint, will fix in #1671

PR #1671, opened 2023-06-27
#1670 fix wrong config path for pulling devops and status password from helm values

thjaeckle:
fixes: #1670


thjaeckle:
@kalinkostashki would you mind taking a look?

Issue #1669, opened 2023-06-27
[Helm]: 500 Internal Server Error. 'node needs to be a replica set member to use read concern' on server dt-ditto-mongodb:27017"


Altair-Bueno:
### cURL Request 
```sh
# ditto:ditto
DITTO_AUTH=ZGl0dG86ZGl0dG8=
DITTO_ADDR=TBD
curl --location "$DITTO_ADDR/api/2/things" \
--header "Authorization: Basic $DITTO_AUTH" -v
```

### Expected response

Empty array. No things currently deployed on Ditto

### Actual response
```text
*   Trying 192.168.43.7:30531...
* Connected to node1.ertis.internal (192.168.43.7) port 30531 (#0)
> GET /api/2/things HTTP/1.1
> Host: node1.ertis.internal:30531
> User-Agent: curl/7.88.1
> Accept: */*
> Authorization: Basic ZGl0dG86ZGl0dG8=
> 
< HTTP/1.1 500 Internal Server Error
< Server: nginx/1.23.4
< Date: Tue, 27 Jun 2023 10:47:44 GMT
< Content-Type: application/json
< Content-Length: 304
< Connection: keep-alive
< version: 2
< correlation-id: 1d40a1c4-7b6e-4b7e-9b32-59e65ed97d07
< accept: */*
< Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
< Access-Control-Allow-Credentials: true
< Access-Control-Expose-Headers: *
< 
* Connection #0 to host node1.ertis.internal left intact
{"status":500,"error":"internalerror","message":"class com.mongodb.MongoQueryException: Query failed with error code 123 and error message 'node needs to be a replica set member to use read concern' on server dt-ditto-mongodb:27017","description":"Please contact the service team or your administrator."}
```

### Helm Values

```yml
# helm upgrade --install --create-namespace --dependency-update dt-ditto oci://registry-1.docker.io/eclipse/ditto -f values.yml --version=3.3.0 --wait --debug

global:
  jwtOnly: false
  hashedBasicAuthUsers: false
  basicAuthUsers:
    ditto:
      user: ditto
      password: ditto
  prometheus:
    enabled: true
    port: 9095

swaggerui:
  enabled: false

dittoui:
  enabled: false

gateway:
  config:
    authentication:
      enablePreAuthentication: true
  devopsPassword: foobar
  statusPassword: foobar

nginx:
  service:
    type: NodePort
[...]

PR #1672, opened 2023-06-27
#1669 by default, configure queryReadConcern: "local" for ditto things-search

thjaeckle:
* if used in a replicated MongoDB setup, this should be changed to `queryReadConcern: "linearizable"`

Fixes: #1669


thjaeckle:
@kalinkostashki would you mind taking a look?

PR #1674, opened 2023-06-30
fix that ImmutablePolicyImports.getPolicyImport with arbitrary CharSequence did not work

thjaeckle:
* reason is that the Map contained `CharSequence` as key, but was internally always `PolicyId` - which however was not "equal to" an arbitrary CharSequence provided to the `getPolicyImport(CharSequence)` method
* this method therefore never returned the expected entry


kalinkostashki (approved):
Nice catch.
LGTM :)

PR #1676, opened 2023-07-03
fix(helm): Allow connectivity inter cluster communication

Altair-Bueno:
Doesn't make much sense to block everything inside the cluster for connectivity, specially by default. You end up with confusing error responses such as this one, making it difficult to use the platform.

```json
{
    "status": 400,
    "error": "connectivity:connection.configuration.invalid",
    "message": "The configured host 'kafka.kafka' may not be used for the connection because the host is blocked: the hostname resolved to a site local address.",
    "description": "It is a blocked or otherwise forbidden hostname which may not be used."
}
```

Also, the regex original regex wasn't right:

```yml
# There is no scaping on plain strings
# https://yaml.org/spec/1.2.2/#plain-style
blockedHostRegex: ^.*\.svc\.cluster\.local$
```

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.2  # chart version is effectively set by release-job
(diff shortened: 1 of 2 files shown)


thjaeckle:
@Altair-Bueno thanks for the PR - true .. that was a leftover from a SaaS setup where connections could be created by non-admins, but by developers using the service.

But I agree that as default this does not make sense 👍 

You would also need to bump the Chart version in order to get Helm linting pass.

thjaeckle (approved):
Thanks for the contribution @Altair-Bueno 👍

PR #1678, opened 2023-07-05
fix non-correct WARN log about potentially failing connections status

thjaeckle:
* the logic was assuming that 1 Publisher actor will only be started if targets are available which is a wrong assumption - it is always started


alstanchev (approved):
LGTM

Issue #1614, opened 2023-04-13
When PATCHing a Thing, it should also be possible to implicitly convert to "Create Thing" if thing is not yet existing


thjaeckle:
The [RFC for HTTP PATCH](https://www.rfc-editor.org/rfc/rfc5789) states that:

> If the Request-URI does not
   point to an existing resource, the server MAY create a new resource,
   depending on the patch document type (whether it can logically modify
   a null resource) and permissions, etc.

So it would also make sense to create a Thing when e.g. doing a `PATCH /api/2/things/<thingId>` when this thing is not yet existing.
Using `If-Match: *` header it still would be possible to only update already existing things.

This would of course only make sense when "patching" a complete thing, so on path `/`.

The already existing `ModifyToCreateThingTransformer` could be enhanced to also handle `MergeThing` commands and transform to  `CreateThing` if the thing is not yet existing.

PR #1680, opened 2023-07-06
fix MergeThing command being used for creating new things did not respect "inlinePolicy"

thjaeckle:
* when processed via Ditto "Adaptable" - there the "inlinePolicy" was missing

Relates to #1614 - due to this bug, using the "_policy" does not work when sending "MergeThing" via DittoProtocol


thjaeckle:
@kalinkostashki @alstanchev could you please take a look on this bugfix?

alstanchev (approved):
LGTM

PR #1682, opened 2023-07-07
fix that CreationRestrictionPreEnforcer did not work

thjaeckle:
* reason was that wrong config was loaded - and e.g. system properties overwrites were not applied
* also cleaned up and improved documentation of that feature

### DIFF
--- a/policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/pre/CreationRestrictionPreEnforcer.java
+++ b/policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/pre/CreationRestrictionPreEnforcer.java
@@ -55,7 +55,9 @@ public final class CreationRestrictionPreEnforcer implements PreEnforcer {
      */
     @SuppressWarnings("unused")
     public CreationRestrictionPreEnforcer(final ActorSystem actorSystem, final Config config) {
-        this.config = DefaultEntityCreationConfig.of(config);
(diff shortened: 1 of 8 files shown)


policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/pre/CreationRestrictionPreEnforcer.java
```diff
@@ -55,7 +55,9 @@ public final class CreationRestrictionPreEnforcer implements PreEnforcer {
      */
     @SuppressWarnings("unused")
     public CreationRestrictionPreEnforcer(final ActorSystem actorSystem, final Config config) {
-        this.config = DefaultEntityCreationConfig.of(config);
+        // explicitly use the ActorSystem config instead of the PreEnforcer config - as the config is loaded from
+        // file "ditto-entity-creation.conf" and extending with system properties of that file should not be broken
+        this.config = DefaultEntityCreationConfig.of(actorSystem.settings().config());
```
kalinkostashki, line 60: why not remove the final Config config if it will not be used anymore.

thjaeckle: Because all extensions must have this constructor with an extension config as second parameter

thjaeckle:
@kalinkostashki did you already manage to review?

kalinkostashki (approved):
LGTM :)

PR #1684, opened 2023-07-10
Helm chart: add support for entity creation via Helm configuration

thjaeckle:


thjaeckle:
@kalinkostashki that's how I identified the bug fixed in #1682 - I was building an enhancement for the Helm chart, making it possible to configure the "entity-creation" restrictions.

Would be great if you could have a look - I would (if no other bugs occur) like to build a 3.3.4 bugfix release, as this is quite important for us to be fixed.

kalinkostashki (approved):
LGTM :)

PR #1688, opened 2023-07-12
UI - Migrate to typescript

thfries:
Hi @thjaeckle,

I switched the whole project to typescript.
By now, types were only introduced were needed. There are a lot of further improvements possible, that could be added over time.
Already some smaller issues were discovered, too. And I still want to investigate.

Sorry for not picking the issues with the features, but I m sure this pays off over time. I try to not block this for too long.


thjaeckle:
Hi @thfries
Cool, that is a good initiative.
I will look into it.
Also, we probably have to adjust the build, right?

I noticed that the Java sources are compiled when doing a UI PR. 
That should also be changed to only transpile Typescript.
I can have a look into that..

thfries:
Oh, where is the compilation of java files triggered?
The build should already work as esbuild is processing ts files by default.
I added a `tsc -noEmit` to the build step, but I think it is optional.

Did another change to the error display, because the ditto error description was ignored.

Rest seems good now.

thfries:
I forgot: ditto sandbox seems to be down?

thjaeckle:
> I forgot: ditto sandbox seems to be down?
> 

Thanks for notifying.. I renewed the certificate which was outdated.

thjaeckle (approved):
LGTM 👍

PR #1697, opened 2023-07-19
fix that in Helm config `jwtOnly=false` will enabled pre-authentication in Ditto

thjaeckle:
* was not done by default before, so would have to be activated by default

Related to discussion: #1685


alstanchev (approved):
LGTM

Issue #1521, opened 2022-10-27
Configure on a namespace basis the fields to index in the thing-search


thjaeckle:
Currently, the complete JSON of a thing (including all attributes+features) is automatically indexed in the search.
We learned from input of the community that this might not always be wished as this could lead to a lot of load on the search DB - which could e.g. be prevented if some fields are never used in the search "fields".

an1310:
This is of strategic interest to us -- when would be a good time to go over requirements/design ideas for a possible contribution?

thjaeckle:
@an1310 I will try to find out where this could be placed the best.
As far as I remember there was an extension point for just that purpose used by Bosch's commercial offering of Ditto.

thjaeckle:
@an1310 I took myself a closer look and I think I found the mentioned extension point to do index only custom fields in namespaces.

I think to remember that we explicitly added `CachingSignalEnrichmentFacade` extension point in Ditto to retrieve only selectively parts of "things" (based on their namespace "search index configuration") when processing - retrieving a "partial thing" in order to update the search index:
https://github.com/eclipse-ditto/ditto/blob/022dffee0c1a21cedd57b74310683db5d7ce881e/internal/models/signalenrichment/src/main/java/org/eclipse/ditto/internal/models/signalenrichment/DittoCachingSignalEnrichmentFacade.java#L113-L158

We would need a custom `JsonFieldSelector` configured per namespace name, defining which fields of a "thing" to retrieve (and cache) for updating the search index.
The configuration could look like:
* a list/map of namespaces
   * namespace
      * indexedFields as `JsonFieldSelector`

Example what I have in mind (to put in the `search.conf` as configuration of the mentioned extension point):
https://github.com/eclipse-ditto/ditto/blob/022dffee0c1a21cedd57b74310683db5d7ce881e/thingsearch/service/src/main/resources/search.conf#L11

```
namespaces = [
  {
    namespace-name = "org.eclipse.ditto.foo"
    indexed-fields = [ 
      "thingId", 
      "policyId", 
      "_revision", 
      "_created", 
      "_modified", 
      "attributes/indexed-one"
      "attributes/complex-jsonobj"
    ]
  }
]
```

I hope that helps a little. Maybe a Ditto committer still at Bosch can share more details how this can be done, based on the identified extension point.

thjaeckle:
@an1310 are you working on this? Any update or PR for providing early feedback?

an1310:
Yes I am.

PR #1698, opened 2023-07-20
[API]: Endpoint for retrieving all policies

Altair-Bueno:
See https://github.com/eclipse-ditto/ditto/issues/1683

### DIFF
--- a/thingsearch/model/src/main/java/org/eclipse/ditto/thingsearch/model/signals/commands/query/QueryPolicies.java
+++ b/thingsearch/model/src/main/java/org/eclipse/ditto/thingsearch/model/signals/commands/query/QueryPolicies.java
@@ -1,0 +1,251 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.ditto.thingsearch.model.signals.commands.query;
+
+
+import org.eclipse.ditto.base.model.headers.DittoHeaders;
+import org.eclipse.ditto.base.model.json.FieldType;
+import org.eclipse.ditto.base.model.json.JsonParsableCommand;
+import org.eclipse.ditto.base.model.json.JsonSchemaVersion;
+import org.eclipse.ditto.base.model.signals.commands.AbstractCommand;
+import org.eclipse.ditto.base.model.signals.commands.CommandJsonDeserializer;
+import org.eclipse.ditto.json.*;
+import org.eclipse.ditto.thingsearch.model.signals.commands.ThingSearchCommand;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Command for searching policies.
+ * @since 1.0.0
+ */
+@Immutable
+@JsonParsableCommand(typePrefix = ThingSearchCommand.TYPE_PREFIX, name = QueryPolicies.NAME)
+public final class QueryPolicies extends AbstractCommand<QueryPolicies> implements ThingSearchQueryCommand<QueryPolicies> {
+
+    /**
+     * Name of the command.
+     */
+    public static final String NAME = "queryPolicies";
+
+    /**
+     * Type of this command.
+     */
+    public static final String TYPE = TYPE_PREFIX + NAME;
+
+    static final JsonFieldDefinition<String> JSON_FILTER =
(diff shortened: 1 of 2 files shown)


thingsearch/model/src/main/java/org/eclipse/ditto/thingsearch/model/signals/commands/query/QueryPolicies.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Command for searching policies.
+ * @since 1.0.0
+ */
+@Immutable
+@JsonParsableCommand(typePrefix = ThingSearchCommand.TYPE_PREFIX, name = QueryPolicies.NAME)
+public final class QueryPolicies extends AbstractCommand<QueryPolicies> implements ThingSearchQueryCommand<QueryPolicies> {
```
Altair-Bueno, line 37: It might be wise to use an abstract class to avoid code duplication, but:

1. I'm not sure what level of abstraction you are aiming for
2. Not even sure if it makes sense, just doing some rough changes until I learn the codebase

thjaeckle: I would say to wait for abstracting it before it really provides a benefit, e.g. in reacting to the command in a similar way when applying the things/policies query.

Altair-Bueno:
Hey @thjaeckle . I've been struggling to comprehend the control flow of the `GET /search/things` endpoint, and I'll like to summarize the steps to make sure I'm not leaving anything important on the table:

1. An incoming request to said endpoint is transformed into a Ditto command (`QueryThings`).
2. The `SearchActor` handles the command by validating it and retrieving all matching IDs from the persistence layer.
3. The retrieved IDs are then passed to `QueryThingsPerRequestActor`.
4. `QueryThingsPerRequestActor` instructs `ThingsAggregatorProxyActor` to retrieve all the corresponding things.
5. Finally, the `ThingsAggregatorProxyActor` gathers the requested things and responds back.

However, I'm curious as I don't understand the reasoning behind this approach. It appears quite complex and involves multiple messages between distributed services. Have you considered the possibility of using a more straightforward approach? On paper, a database query seems to be more performant and resource-efficient, as it would only require one microservice and the optimised database (in the end, Mongo is designed for high performance reads).

PS: This comment really help me out to finally gasp where things came from

https://github.com/eclipse-ditto/ditto/blob/58ce86b14aef9c43af2d153d0e4d5fa134065911/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/SearchActor.java#L105

thjaeckle:
@Altair-Bueno this approach is basically done in order to not transfer 200 search results in a single message across the cluster (with a max. thing size of 100 KiB this would be a 2 MiB message).  
We configured a max. message size in the cluster of 256 KiB in order to e.g. have an optimized memory consumption and not need to "reserve" a lot of non-needed memory for transfering huge messages.

The solution to still be able to retrieve (max page size) 200 results in a single page at the search API is to
1. use the search index to find the "thingIds" which match the query
2. respond with the list of "thingIds"
3. then 1-by-1 ask for the data of the things from the "things" service
   * this also will "filter" the responses based on the policy to not have data included which the user is not allowed to see
4. aggregate them back at the gateway into a single "QueryThingsResponse"

So yes, it is complex - but it is required to not exceed messaging limits.  
Policies can also be 100 KiB big (default configuration) - so the same approach has to be applied here as well I fear.

Altair-Bueno:
Thanks for your response Thomas. I do understand now the motivation behind performing multiple asks to the things service instead a single big one. However, that doesn't explain why asking each actor was chosen over a database query from the gateway component itself.

The only reason I can think of is to avoid stale data to be served (e.g. a thing with changes that hasn't been persisted yet), but IMHO that seems acceptable given the drawbacks of the current approach: increased pressure on Akka, latency and software complexity. If one would like a more precise representation of the actual twin, a `GET /things/{thingid}` query could performed to retrieve the latest changes.

Please keep in mind that I have little experience with Akka and this might be common practice in actor systems. It just doesn't click with me to rely so heavily on actors when other approaches exists.

thjaeckle:
> However, that doesn't explain why asking each actor was chosen over a database query from the gateway component itself.

As I wrote:
>   - this also will "filter" the responses based on the policy to not have data included which the user is not allowed to see

The search index does not necessarily have all of the information which the thing contains - its purpose is "only" to return with a list of matching IDs to a query.  
With this approach, it is also possible to e.g. limit the fields which the search index should index as discussed in #1521

As things and policies apply the event sourcing pattern (e.g. in order to provide history and audit log capabilities), we cannot "simply" query their database from the gateway (as the actual state would have to be constructed based on persisted events) - which you should not do anyways in service oriented architectures - to use the database of another service without the other service being aware.

This is not going to change or be simplified as I don't see how this can work otherwise.

thjaeckle:
@Altair-Bueno this is no longer in work, or is it?

Altair-Bueno:
Hi Thomas. Indeed, we are no longer working on this. I've recently been reassigned, and unfortunately, this feature development slipped through the cracks. I apologize for any inconvenience this may have caused.

thjaeckle:
@Altair-Bueno ok, no problem - thanks for letting us know 👍

PR #1702, opened 2023-07-21
Support for basic nginx-ingress authentication

vladica:
Enable basic auth for basicAuthUsers with ingress

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -217,14 +217,12 @@ ingress:
         backendSuffix: gateway
       - path: /health
         backendSuffix: gateway
+    kubernetesAuthAnnotations: |
+      nginx.ingress.kubernetes.io/auth-type: basic
+      nginx.ingress.kubernetes.io/auth-secret: {{ .Release.Name }}-nginx-ingress-htpasswd
+      nginx.ingress.kubernetes.io/auth-realm: 'Authentication required to use HTTP API!'
     # annotations defines k8s annotations to add to the Ingress
     annotations:
-      # type of authentication
-      nginx.ingress.kubernetes.io/auth-type: basic
-      # name of the secret that contains the user/password definitions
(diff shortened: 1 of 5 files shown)


thjaeckle:
Hi @vladica and thanks for the PR.
That is a good addition for when using ingress :+1:

Could you bump the Helm chart version, the linting fails because of that.

deployment/helm/ditto/values.yaml
```diff
@@ -219,6 +219,12 @@ ingress:
         backendSuffix: gateway
     # annotations defines k8s annotations to add to the Ingress
     annotations:
+      # type of authentication
+      nginx.ingress.kubernetes.io/auth-type: basic
+      # name of the secret that contains the user/password definitions
+      nginx.ingress.kubernetes.io/auth-secret: my-ditto-nginx-ingress-htpasswd
```
thjaeckle, line 225: `my-ditto` is your release name, right?
That should not be a part of the values.yaml
Do you have another idea of how to solve this?

deployment/helm/ditto/values.yaml
```diff
@@ -217,14 +217,12 @@ ingress:
         backendSuffix: gateway
       - path: /health
         backendSuffix: gateway
+    kubernetesAuthAnnotations: |
+      nginx.ingress.kubernetes.io/auth-type: basic
+      nginx.ingress.kubernetes.io/auth-secret: {{ .Release.Name }}-nginx-ingress-htpasswd
```
thjaeckle, line 222: I don't think that it is possible to use Helm variables in the `values.yaml` ..
Are you sure about that?

thjaeckle: Ah, wow .. did not see that.
That is a nice hack, thx for explaining 👍

thjaeckle (approved):
LGTM, thanks a lot for the PR 👍

PR #1706, opened 2023-07-25
Support HTTP POST method for /search and /count

PJGitLan:
Additional POST method accepting the search filter as payload. Workaround for longer RQL filters until QUERY is possible to implement.

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
@@ -12,22 +12,26 @@
  */
 package org.eclipse.ditto.gateway.service.endpoints.routes.thingsearch;
 
-import akka.http.javadsl.server.Directives;
-import akka.http.javadsl.server.PathMatchers;
-import akka.http.javadsl.server.RequestContext;
-import akka.http.javadsl.server.Route;
+import java.util.Arrays;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import javax.annotation.Nullable;
+
 import org.eclipse.ditto.base.model.headers.DittoHeaders;
-import org.eclipse.ditto.base.model.signals.commands.Command;
 import org.eclipse.ditto.gateway.service.endpoints.routes.AbstractRoute;
 import org.eclipse.ditto.gateway.service.endpoints.routes.RouteBaseProperties;
 import org.eclipse.ditto.thingsearch.model.signals.commands.query.CountThings;
 import org.eclipse.ditto.thingsearch.model.signals.commands.query.QueryThings;
 
-import javax.annotation.Nullable;
-import java.util.*;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import akka.http.javadsl.server.Directives;
+import akka.http.javadsl.server.PathMatchers;
+import akka.http.javadsl.server.RequestContext;
+import akka.http.javadsl.server.Route;
 
 /**
  * Builder for creating Akka HTTP routes for {@code /search/things}.
@@ -78,25 +82,25 @@ public final class ThingSearchRoute extends AbstractRoute {
                 // GET things/count?filter=<filterString>&namespaces=<namespacesString>
                 get(() -> thingSearchParameterOptional(
                         params -> handlePerRequest(ctx,
-                            CountThings.of(calculateFilter(params.get(ThingSearchParameter.FILTER)),
-                            calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
-                                    dittoHeaders)))),
+                                CountThings.of(calculateFilter(params.get(ThingSearchParameter.FILTER)),
+                                        calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
+                                        dittoHeaders)))),
                 // POST things/count
                 post(() -> ensureMediaTypeFormUrlEncodedThenExtractData(
                         ctx,
                         dittoHeaders,
-                        payloadSource ->  handlePerRequest(
+                        formFields -> handlePerRequest(
                                 ctx,
-                                dittoHeaders,
-                                payloadSource,
-                                requestString -> getFormParameters(
-                                        requestString,
-                                        params ->  CountThings.of(
-                                                calculateFilter(params.get(ThingSearchParameter.FILTER)),
-                                                calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
-                                                dittoHeaders)
-                                )
-                        )))
+                                CountThings.of(
+                                        calculateFilter(
+                                                formFields.getOrDefault(ThingSearchParameter.FILTER.toString(),
+                                                        List.of())),
+                                        calculateNamespaces(
+                                                formFields.getOrDefault(ThingSearchParameter.NAMESPACES.toString(),
+                                                        List.of())),
+                                        dittoHeaders)
+                        )
+                ))
         );
     }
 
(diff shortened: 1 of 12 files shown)


gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
```diff
@@ -79,10 +79,28 @@ public Route buildSearchRoute(final RequestContext ctx, final DittoHeaders ditto
      */
     private Route countThings(final RequestContext ctx, final DittoHeaders dittoHeaders) {
         // GET things/count?filter=<filterString>&namespaces=<namespacesString>
-        return get(() -> thingSearchParameterOptional(params -> handlePerRequest(ctx,
-                CountThings.of(calculateFilter(params.get(ThingSearchParameter.FILTER)),
-                        calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
-                        dittoHeaders))));
+        // POST things/count?namespaces=<namespacesString>
```
thjaeckle, line 82: Please move that comment to where the `post()` is called

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
```diff
@@ -79,10 +79,28 @@ public Route buildSearchRoute(final RequestContext ctx, final DittoHeaders ditto
...
+        return concat(
+                get(() -> thingSearchParameterOptional(
+                        params -> handlePerRequest(ctx,
+                            CountThings.of(calculateFilter(params.get(ThingSearchParameter.FILTER)),
+                            calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
+                                    dittoHeaders)))),
+                post(() -> thingSearchParameterOptional(
+                        params -> ensureMediaTypeJsonWithFallbacksThenExtractDataBytes(
```
thjaeckle, line 90: By using `ensureMediaTypeJsonWithFallbacksThenExtractDataBytes` you enforce the POST payload to be of `application/json` - do you really want that?
Did you test that?

To me, this looks like it could not be what is wanted - as the `filter` below inserted into `CountThings` is certainly not a Json, but a plain text string.

I would expect the content-type to be `application/x-www-form-urlencoded` instead  and that e.g. also query params would be provided all as form values - instead of mixing those.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
```diff
@@ -96,13 +114,43 @@ private Route searchThings(final RequestContext ctx, final DittoHeaders dittoHea
...
         //           &namespaces=<namespacesString>
         //           &nextPageKey=<nextPageKey>
-        return get(() -> thingSearchParameterOptional(params -> handlePerRequest(ctx,
-                QueryThings.of(calculateFilter(params.get(ThingSearchParameter.FILTER)),
-                        calculateOptions(params.get(ThingSearchParameter.OPTION)),
-                        AbstractRoute.calculateSelectedFields(params.get(ThingSearchParameter.FIELDS))
-                                .orElse(null),
-                        calculateNamespaces(params.get(ThingSearchParameter.NAMESPACES)),
-                        dittoHeaders))));
+        // POST things?options=<optionsString>PUT
```
thjaeckle, line 117: `POST things?options=<optionsString>PUT` ? 
What is that supposed to mean?

As mentioned above, I would not "mix" query params with form params of the POST - and provide all query params as form values instead.

gateway/service/src/test/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRouteTest.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: Please adjust creation year of the file to 2023

gateway/service/src/test/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRouteTest.java
```diff
@@ -0,0 +1,75 @@
...
+                        JsonKey.of("filter"),
+                        filter
+                );
+    }
+
+    @Test
+    public void countThingsShouldGetFilterFromBody() {
+        final var filter = "and(like(definition,\"*test*\"))";
+        final var result = underTest.run(HttpRequest.POST("/search/things/count")
+                .withEntity(ContentTypes.APPLICATION_JSON, filter));
```
thjaeckle, line 61: Please also adjust to `application/x-www-form-urlencoded` content-type.

thjaeckle (changes requested):
Thanks for the PR.

Please adjust the `POST` search to take all parameters from form parameters and adjust the supported content-type for that to `application/x-www-form-urlencoded` (form encoding).

AFAIK that is the best practice do do "POST" as replacement for "GET" with too long query parameters.

We would then also need to update some documentation (OpenAPI and also Ditto docs) as part of the PR - but one step after another :)

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: This should stay at 2017 - the file creation year.
I referenced the other file, the unit test which was added.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
PJGitLan: Typo

gateway/service/src/test/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRouteTest.java
PJGitLan: done

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/thingsearch/ThingSearchRoute.java
PJGitLan: oops 😅

thjaeckle:
@PJGitLan I had a brief look at the implementation - and it looks quite good. 👍 
I will invest more time in manually testing ..

Now we would only need to provide documentation for this in order to get it merged.
As mentioned before:
> We would then also need to update some documentation (OpenAPI and also Ditto docs) as part of the PR - but one step after another :)

thjaeckle:
@PJGitLan wanted to ping you - are you still interested in getting this PR merged?

PJGitLan:
> @PJGitLan wanted to ping you - are you still interested in getting this PR merged?

yes, I am still want to get this merged.  Sorry for the long wait.

thjaeckle:
Status: Waiting for documentation - as stated before.

thjaeckle:
@PJGitLan thanks a lot - I will review soon.
Should then hopefully make it to the upcoming Ditto 3.4.0 :)

thjaeckle:
@PJGitLan I added a commit to your PR and hope you don't mind
I noticed that the form parameter parsing was handmade and quite complex + error prone, so I reworked that part to use Akka HTTP's `formFieldMultiMap()` API to extract all form parameters in a very similar way to how we extract the query parameters.

This implementation is now also way more tolerant how the `filter` and so on parameters are provided - as far as I see, they don't need to be URL encoded which makes it a lot easier.

Please check my changes and if you can live with them ;)

PJGitLan:
> @PJGitLan I added a commit to your PR and hope you don't mind I noticed that the form parameter parsing was handmade and quite complex + error prone, so I reworked that part to use Akka HTTP's `formFieldMultiMap()` API to extract all form parameters in a very similar way to how we extract the query parameters.
> 
> This implementation is now also way more tolerant how the `filter` and so on parameters are provided - as far as I see, they don't need to be URL encoded which makes it a lot easier.
> 
> Please check my changes and if you can live with them ;)

Changes look good to me. I didn't know there was a 'formFieldMultiMap()' method. 
Thanks for the help.

thjaeckle (approved):

PR #1716, opened 2023-08-11
fix providing Ditto Adaptable information in the "_context" of an SSE event

thjaeckle:
Previously, the optional `_context` of a SSE subscription only contained the `headers`.

For the `NormalizedMessageMapper` of connectivity, however, also additional information from the Ditto Adpatable is provided, like:
* topic
* path

This PR adds this information as well for the SSE `_context` and fixes that the `headers` included Ditto internal headers which should not be exposed.


alstanchev (approved):
LGTM

PR #1718, opened 2023-08-20
Correct documentation of RawMessage payload mapper headers

### LINKED ISSUE #1717: [docs] RawMessage payload mapper expects header "ditto-message-subject" not "subject"
[Documentation](https://eclipse.dev/ditto/connectivity-mapping.html#rawmessage-mapper) of RawMessage payload mapper refers to header "subject" whereas correct header name is "ditto-message-subject".

dimabarbul:
Resolves #1717


thjaeckle (approved):
Really good catch - indeed that is wrongly documented.
Thanks a lot for the contribution @dimabarbul 👍

PR #1724, opened 2023-08-23
fix using "/cloudevents" endpoint for other entities than things

thjaeckle:
* e.g. policy commands could not be created with the "POST /cloudevents" endpoint as "response-required" was always set to `false` in combination with setting a "requested ACK" (which is only available for thing commands)


thjaeckle:
@alstanchev @kalinkostashki it would be great if you could have a short look at this bugfix

alstanchev (approved):
LGTM

PR #1734, opened 2023-09-06
Spread ditto services equally among cluster nodes

vladica:
This PR delivers two improvements:
1. Add Pod Topology Spread Constraints to spread ditto services equally among cluster nodes. Each deployment has the ability to set constraints independently.
2. Extract nginx-ingress configuration hard-coded values to the values.yaml file:
    - replicaCount
    - updateStrategy
    - minReadySeconds
    - revisionHistoryLimit
    - resources


thjaeckle (approved):
Thx a lot @vladica for this contribution.

I also had this on my "todo / to look into" list - this is super useful for maintaining high availability in case of a zone crashing when e.g. 3 zones are available. 👍 
So: Much appreciated.

PR #1735, opened 2023-09-07
Enable configuration of size "limits" in Helm chart

thjaeckle:
This PR exposes the configuration of limits like "thing size, policy size, message size" to Helm values.


alstanchev (approved):
LGTM

PR #1737, opened 2023-09-08
#1731 use tpl function for otelExporterOtlpEndpoint value

### LINKED ISSUE #1731: [Helm chart] Allow otelExporterOtlpEndpoint value to contain Release/Values references
When deploying the ditto Helm chart as part of the IoT Packages cloud2edge chart, it would be useful to provide combined tracing support for Ditto and Hono. The straightforward thing to do for this, would be enabling tracing support in the Hono chart, and referencing the corresponding OTLP endpoint in the Ditto tracing configuration.
This would mean setting `ditto.global.tracing.otelExporterOtlpEndpoint` to `http://c2e-jaeger-collector:4317` for example. But this only works if the used release name when deploying the cloud2edge chart is indeed `c2e`.

To support arbitrary release names, `http://{{ .Release.Name }}-jaeger-collector:4317` would have to be used as `otelExporterOtlpEndpoint` value. 

In order for this to work, the ditto Helm chart would just need to use the `tpl` function in the place where `otelExporterOtlpEndpoint` gets used. This is a feature request for doing so.

calohmn:
Fixes #1731

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.8  # chart version is effectively set by release-job
+version: 3.3.9  # chart version is effectively set by release-job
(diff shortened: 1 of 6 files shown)


thjaeckle (approved):
LGTM, thanks for the enhancement @calohmn 👍

PR #1738, opened 2023-09-08
fix wrong exception message when policy could not be implicitly created when creating thing

thjaeckle:
* the exception message did not preserve the "root cause", e.g. a not existing policy import and e.g. always produced a text indicating to a policy "conflict"


thjaeckle:
I noticed that when making use of "implicit policy creation" - the error message was often wrong when creating the policy did not work.

alstanchev (approved):
LGTM

Issue #1739, opened 2023-09-08
Improve logging by adding the W3C `traceparent` header as MDC field to logs


thjaeckle:
Currently, only the "correlation-id" is logged to the MDC for each log-statement in Ditto.

When having [tracing](https://eclipse.dev/ditto/installation-operating.html#tracing) activated, it would be very useful to also provide the [traceparent](https://github.com/w3c/trace-context/blob/main/spec/20-http_request_header_format.md#traceparent-header), as defined by the [W3C TraceContext](https://www.w3.org/TR/trace-context/), as an MDC field.

PR #1740, opened 2023-09-08
#1739 provide the traceparent header as MDC value in logs

thjaeckle:
* generify the fields to provide to the MDC in CommonMdcEntryKey enum
* rename "x-correlation-id" in logs to just "correlation-id"
* exchange some places where only correlationId was extracted from a map of headers with parsing all the headers for MDC worthy fields

Resolves: #1739

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/RootRouteHeadersStepBuilder.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/RootRouteHeadersStepBuilder.java
@@ -15,7 +15,9 @@ package org.eclipse.ditto.gateway.service.endpoints.routes;
 import static org.eclipse.ditto.base.model.common.ConditionChecker.checkNotNull;
 
 import java.text.MessageFormat;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletionStage;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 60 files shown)


thjaeckle:
@alstanchev this PR would also be cool to be included in 3.4.0
I mentioned it briefly in today's Ditto community meeting, it provides the `traceparent` of [W3C Trace Context](https://www.w3.org/TR/trace-context/#traceparent-header) as MDC value in a generic way - similar to providing the `correlationId`.

I could simplify the loggers also quite a bit.

I also fixed that the `traceparent` header coming via an HTTP request was not used/forwarded.

alstanchev:
> #1739

I have mentioned the issue https://github.com/eclipse-ditto/ditto/issues/1739 or do you want to mention the PR specifically?

thjaeckle:
> I have mentioned the #1739 issue or do you want the PR to be mentioned specifically?

When there is an issue and a PR I normally only mention and link the issue ;) 

I also pinged you about reviewing, but you're right .. release notes would also be good - I could also do that however..

thjaeckle:
@alstanchev @kalinkostashki Could you please have a look? Still missing an approval here .. :)

alstanchev:
i am on it

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/RootRouteHeadersStepBuilder.java
```diff
@@ -122,7 +124,16 @@ QueryParametersStep withRequestContext(final RequestContext requestContext) {
 
         private Map<String, String> getFilteredExternalHeaders(final HttpMessage httpRequest) {
             final Iterable<HttpHeader> headers = httpRequest.getHeaders();
+            final Set<String> headerNames = new HashSet<>();
             final Map<String, String> externalHeaders = StreamSupport.stream(headers.spliterator(), false)
+                    .map(httpHeader -> {
```
alstanchev, line 129: Just out of curiocity, what is the reason behind this double duplicate headers check?

alstanchev (approved):
LGTM

PR #1741, opened 2023-09-10
Fix typo in documentation regarding changing inbox to outbox

dimabarbul:
The text below suggests that path in response has changed from /inbox/... to /outbox/... But the example JSON does not shows this change.

### DIFF
--- a/documentation/src/main/resources/pages/ditto/protocol-specification-things-messages.md
+++ b/documentation/src/main/resources/pages/ditto/protocol-specification-things-messages.md
@@ -109,8 +109,6 @@ The answer of the coffee machine has the same `topic` and `correlation-id`
 as the original message. As we can see, the response does not only contain a
 `value` but also the `status` of the response
 which is based on the [HTTP status codes](protocol-specification.html#status). 
-Notice, that the `path` of the Message has changed from *inbox* to *outbox*, 
-which means the Message was sent *from* the Thing.
 Ditto automatically added some headers that we can ignore for now.
 
 ### Receiving a message


thjaeckle:
In fact, this change from `inbox` to `outbox` is not required, it is more of a relict of wrong documentation - I think the example once had this change to `outbox` also contained.

I am wondering if it would make more sense to delete this hint about changing "inbox" to "outbox". It does not really make things clearer - instead more confusing.  
* a message response is already identified by the use of `status` in the payload
* it is correlated via the `correlation-id` header
* no permissions are checked - knowing the `correlation-id` is sufficient to answer to a message

documentation/src/main/resources/pages/ditto/protocol-specification-things-messages.md
```diff
@@ -99,7 +99,7 @@ The response we would get from our coffee machine could look something like this
         "content-type": "text/plain",
         "version": 1,
     },
-    "path": "/inbox/messages/ask",
+    "path": "/outbox/messages/ask",
```
thjaeckle, line 102:
```suggestion
    "path": "/inbox/messages/ask",
```

thjaeckle (changes requested):
I would rather remove the sentence:

```
Notice, that the `path` of the Message has changed from *inbox* to *outbox*, 
which means the Message was sent *from* the Thing.
```

documentation/src/main/resources/pages/ditto/protocol-specification-things-messages.md
dimabarbul: Can I force-push or is it better to make second commit?

thjaeckle:
LGTM, thanks 👍

documentation/src/main/resources/pages/ditto/protocol-specification-things-messages.md
dimabarbul: Question for future PRs: am I expected to mark conversation as resolved or author of the comment?

thjaeckle: Either way works .. ;)

PR #1742, opened 2023-09-10
Correct wording about need for URL-encoding of encryption key

dimabarbul:
I found this wording a bit misleading as it says that the key should be URL-encoded whereas it needs to be only base64-encoded but with URL-safe alphabet (link to RFC is added).
Also corrected line in link to method generateAESKeyAsString (currently it points to method generateAESKey).


thjaeckle (approved):
LGTM, thx 👍

PR #1743, opened 2023-09-10
Allow spaces inside placeholder in target issued acknowledgement label

dimabarbul:
Issued acknowledgement label for target must start with connection ID. This can be achieved by using `connection:id` placeholder. Usually placeholders allow using spaces inside of (right after opening and right before closing) curly brackets, for example, `{{ connection:id }}`. But validation does not allow spaces in placeholder in target issued acknowledgement label.


thjaeckle (changes requested):
@dimabarbul I added my review findings as comments.
Thanks a lot for finding and directly fixing 👍

PR #1753, opened 2023-09-18
update maven plugins to latest versions

thjaeckle:
* and update legal / dependencies documents accordingly


thjaeckle:
@alstanchev @kalinkostashki I updated all Maven plugins to their latest versions and also e.g. the used Scala version.
Found an issue when using `CompletableFuture.completedStage()` with new Scala 2.13.12 - so I replaced use of this with `CompletableFuture.completedFuture()` which works.

alstanchev (approved):
LGTM

PR #1754, opened 2023-09-19
provide a new function to join placeholder array pipeline elements into a single string

thjaeckle:
Named: `fn:join('delimiter')`

Can e.g. be used to join all affected features of a twin change being contained in the `{{ feature:id }}` placeholder to a single string by doing:
```
{{ feature:id | fn:join(',') }}
```


alstanchev (approved):
LGTM

PR #1756, opened 2023-09-21
Add MQTT message expiry interval

### LINKED ISSUE #1729: Add support for MQTT 5 message expiry interval property
[Message expiry interval](https://www.hivemq.com/blog/mqtt5-essentials-part4-session-and-message-expiry/#mqtt-5s-message-expiry-interval) property in MQTT 5 allows specifying that a message is valid for limited period of time. This allows a message broker to discard messages when the interval is up, and clients will not receive (and process) messages that are expired.

An example of such message can be live message to device that requires response. Processing such message after the timeout is unnecessary (and sometimes not desirable).

It would be nice to have special `mqtt.message-expiry-interval` (or, probably, something shorter) header that contains/sets message expiry interval property (for source and target mappings).

dimabarbul:
Resolves #1729

Please, be patient. As I'm C# developer, I assume, there are moments done in non-Java way. If you find some names or language constructs that feel strange or non-Java-like, I'd be grateful if you point them out to me.

My concerns are:

1. In class `ExternalMessageToMqttPublishTransformerTest` constant `RETAIN_FALL_BACK` was not used. It might have been used in method:

    ```java
    @Test
    public void transformExternalMessageWithoutRetainValueInHeadersFallsBackToFalse() {
        Mockito.when(externalMessage.getHeaders()).thenReturn(DittoHeaders.empty());
    
        assertThat(ExternalMessageToMqttPublishTransformer.transform(externalMessage, mqttPublishTarget))
                .isEqualTo(TransformationSuccess.of(externalMessage,
                        GenericMqttPublish.builder(MQTT_TOPIC_FALL_BACK, MQTT_QOS_FALL_BACK)
                                .retain(false)
                                .build()));
    }
    ```

    I believe, as long as the method is named as "*FallsBackToFalse" it's ok to use false instead of class constant.
2. I'm not sure about implementation of class `MessageExpiryInterval`, especially about method `getAsOptionalLong`. I was choosing between `Optional<Long>` and `OptionalLong`. I chose `OptionalLong` because it doesn't do boxing and it has only 2 states - does or doesn't have value, whereas `Optional<Long>` allows 3 states - doesn't have value, has `null` value and has not-`null` value.
3. Header name `mqtt.message-expiry-interval` seems a bit long. I thought about `mqtt.expiry` but decided to stick with MQTT 5 name.
4. In class `GenericMqttPublish` I didn't know what to do when we got message from broker with invalid message expiry interval. I decided to not throw any exception and just fall back to no message expiry interval. Probably, need to log something here. Throwing exception doesn't seem to be an option here, cause it'll be needed to be added to signatures of whole call tree of this method.
5. I added MQTT version to enum MqttHeader. Not sure if adding this kind of info to enum is OK in general.
6. I consider adding message expiry interval to last will message as out-of-scope.

### DIFF
--- a/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
+++ b/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
@@ -24,7 +24,7 @@ import javax.annotation.concurrent.Immutable;
  * The maximum seconds is {@value MAX_INTERVAL_SECONDS}.
  */
 @Immutable
-public class MessageExpiryInterval {
+public final class MessageExpiryInterval {
 
     public static final long MIN_INTERVAL_SECONDS = 1L;
 
@@ -36,6 +36,14 @@ public class MessageExpiryInterval {
         this.seconds = seconds;
     }
 
+    /**
+     * Returns an instance of {@code MessageExpiryInterval} for the specified interval.
+     *
+     * @param seconds the message expiry interval duration in seconds. The value must be between
+     * {@value #MIN_INTERVAL_SECONDS} and {@value #MAX_INTERVAL_SECONDS} (both inclusive).
+     * @return the MessageExpiryInterval for interval of {@code seconds}.
+     * @throws IllegalMessageExpiryIntervalSecondsException if the seconds is less than {@value MIN_INTERVAL_SECONDS} or greater than {@value #MAX_INTERVAL_SECONDS}.
+     */
     public static MessageExpiryInterval of(final long seconds)
             throws IllegalMessageExpiryIntervalSecondsException {
         if (seconds < MIN_INTERVAL_SECONDS || seconds > MAX_INTERVAL_SECONDS) {
@@ -50,10 +58,20 @@ public class MessageExpiryInterval {
         return new MessageExpiryInterval(seconds);
     }
 
+    /**
+     * Returns an instance of the empty {@code MessageExpiryInterval}.
+     *
+     * @return the empty message expiry interval.
+     */
     public static MessageExpiryInterval empty() {
         return new MessageExpiryInterval(null);
     }
 
+    /**
+     * Returns message expiry interval seconds.
+     *
+     * @return optional count of seconds in the interval.
+     */
     public OptionalLong getAsOptionalLong() {
         return seconds != null ?
                 OptionalLong.of(seconds) :
--- a/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
+++ b/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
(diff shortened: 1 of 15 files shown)


thjaeckle:
Thanks a lot for the contribution @dimabarbul 
I will try to look at it soon 👍

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Representation of the MQTT 5 message expiry interval.
+ * The minimum seconds is {@value MIN_INTERVAL_SECONDS}.
+ * The maximum seconds is {@value MAX_INTERVAL_SECONDS}.
+ */
+@Immutable
+public class MessageExpiryInterval {
```
thjaeckle, line 27: this class may be `final` - Ditto codebase applies final wherever possible

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public static final long MIN_INTERVAL_SECONDS = 1L;
+
+    public static final long MAX_INTERVAL_SECONDS = 4_294_967_295L;
```
thjaeckle, line 31: Those are ~136 years, is that correct?
Is this max value specified?

Seems a little too much to be reasonable.

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Nullable private final Long seconds;
+
+    private MessageExpiryInterval(@Nullable final Long seconds) {
+        this.seconds = seconds;
+    }
+
+    public static MessageExpiryInterval of(final long seconds)
```
thjaeckle, line 39: Could you please add javadoc to public methods?

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
```diff
@@ -0,0 +1,86 @@
...
+                            MIN_INTERVAL_SECONDS,
+                            MAX_INTERVAL_SECONDS,
+                            seconds),
+                    null);
+        }
+
+        return new MessageExpiryInterval(seconds);
+    }
+
+    public static MessageExpiryInterval empty() {
```
thjaeckle, line 53: Also, please add javadoc.

thjaeckle (changes requested):
Hey @dimabarbul 
Looks really good, thanks a lot. 👍 
I only added a few comments, mainly about missing javadoc on public methods.

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/mqtt/MessageExpiryInterval.java
dimabarbul: [Specification](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901112) says that the value is 4-byte integer value. Nothing about maximum value.
HiveMQ doc [says](https://github.com/hivemq/hivemq-community-edition/wiki/MQTT-Specific-Configuration#mqtt-specific-configuration-options) that maximum value for message expiry is 4.294.967.296. This seems strange to me, cause this takes more than 4 bytes.
So I took maximum 4-byte unsinged value.

thjaeckle (approved):
LGTM, thanks for the contribution @dimabarbul 👍

PR #1759, opened 2023-09-26
fix(helm): fix missing whitespace in JAVA_TOOL_OPTIONS

WatcherWhale:
In the JAVA_TOOL_OPTIONS environment variable, the different values get put together without any (white)space.
This could lead to some options not being processed by the JVM.

This PR fixes this by removing the "trim whitespace" character from the affected template.

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.8  # chart version is effectively set by release-job
+version: 3.3.9  # chart version is effectively set by release-job
 appVersion: nightly
 keywords:
   - iot-chart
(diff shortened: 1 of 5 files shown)


deployment/helm/ditto/Chart.yaml
```diff
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.7  # chart version is effectively set by release-job
+version: 3.3.8  # chart version is effectively set by release-job
 appVersion: 3.3.7
```
thjaeckle, line 20: @WatcherWhale could you please update to `appVersion: nightly` - that should fix the akka/pekko issues against master branch

thjaeckle, line 20:
```suggestion
appVersion: nightly
```

deployment/helm/ditto/Chart.yaml
```diff
@@ -16,8 +16,8 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.7  # chart version is effectively set by release-job
-appVersion: 3.3.7
+version: 3.3.8  # chart version is effectively set by release-job
```
thjaeckle, line 19:
```suggestion
version: 3.3.9  # chart version is effectively set by release-job
```

WatcherWhale:
First of all thanks for the help!
Yeah, it is not really readable at the moment. I think the root of the original problem is `{{- .Values.global.pekkoOptions }}` having the trim whitespace character, maybe just reduce this PR to only change that line to `{{ .Values.global.pekkoOptions }}`?

thjaeckle:
@WatcherWhale yes, sounds good to me, let's try that first

thjaeckle (approved):
@WatcherWhale looks good now, I would approve! :) 
Thanks for the contribution

PR #1760, opened 2023-09-29
Allow BASIC authentication for devops/status users while using Helm deployment and Ingress Controller 

RockyMM:
At the moment all `/devops` and `/status` paths are routed through `api` Ingress resource. Since the introduction of BASIC auth for ditto user(s), `/devops` and `/status` paths could not be authenticated with devops/status users.

This PR introduces a new Ingress resource to control devops and status paths. The authorization header is simply forwarded to the gateway.

Another update is parameterizing preStop wait time on the IngressController. More info in the comments in the `values.yaml`.

Also Ingress Controller version is updated to 1.9.0. With this all regex-type paths are given `ImplementationSpecific` pathType instead of Prefix.

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,8 +16,8 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.7  # chart version is effectively set by release-job
-appVersion: 3.3.7
+version: 3.3.8  # chart version is effectively set by release-job
+appVersion: nightly
 keywords:
   - iot-chart
   - digital-twin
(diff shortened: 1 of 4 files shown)


RockyMM:
Important thing to note - I could not test successfully with the `main` branch. None of the pods `gateway`, `things` or `thingssearch` could successfully start. What I see is:

```
Running CoordinatedShutdown with reason [JvmExitReason]
```

Seems like a akka/pekko problem. 

I cherry picked all commits over `release-3.3` branch and that worked well.

kalinkostashki (approved):
LGTM :+1:

deployment/helm/ditto/Chart.yaml
```diff
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.3.7  # chart version is effectively set by release-job
+version: 3.3.8  # chart version is effectively set by release-job
 appVersion: 3.3.7
```
thjaeckle, line 20: @RockyMM could you please update to `appVersion: nightly` - that should fix the akka/pekko issues against master branch

thjaeckle: That fixed the Helm issue, thx 👍

PR #1769, opened 2023-10-09
reduce Kamon startup warnings and logs

thjaeckle:
* made loading of "otel-trace-reporter" configurable and by default, don't load it
* introduce new env var `DITTO_TRACING_OTEL_TRACE_REPORTER_ENABLED` which can in addition to `DITTO_TRACING_ENABLED` be used to just switch off OTEL reporting of traces


thjaeckle:
@alstanchev what we talked about in today's meeting

This PR reduces the "noise" of Kamon during startup and also makes the "OTEL reporting" module load optionally.

A review would be great :)

alstanchev (approved):
LGTM

PR #1772, opened 2023-10-12
provide new option for "if-equal" header: "skip-minimizing-merge"

thjaeckle:
* which reduces the part of a merge/patch command to the actually changed elements, removing non-changed elements
* this reduces e.g. required storage in the MongoDB by a lot, if redundant data is sent often
* this also reduces the event payload to the actually changed parts of the thing

Would want to add this to 3.4.0 as it would be a really powerful feature in order to reduce:
* payload stored in MongoDB journal entries
* payload size in events

It is only supported:
* for "merge thing commands"
* and if the header `if-equal: skip-minimizing-merge` is used


thjaeckle:
@alstanchev I still want to provide some unit tests and documentation .. :)

alstanchev (approved):
LGTM 👍 👍 👍

PR #1773, opened 2023-10-13
several small Ditto UI improvements

### LINKED ISSUE #1712: Patching a thing with a null value in JSON body is not reflected in the UI in real time
When sending a patch request to Ditto in a JSON merge patch format and using a null specifier to request a deletion of a JSON key from the thing, if said thing was already selected in the UI console, it does **not** get updated right away to reflect this removal. In order to see whether the removal happened, one needs to either refresh the page, or deselect and select the thing in question again.

Considering that adding or modifying things via patch requests results in the selected thing in the UI being updated right away, I suspect this could be a bug.

For example, a patch request with the following body would result in the updated "brewingTemp" being visible right away in the UI for the selected thing, but "brewed-coffees" would also be visible until a page is refreshed, or the thing was deselected and selected again.

```
{
    "features" : {
        "coffee-brewer":{
           "properties":{
               "brewed-coffees": null, //remove key
               "brewingTemp":87 //update value
           }
        }
    }
}
```

thjaeckle:
* add a tab "Message to Thing" to send thing messages
* add a loading spinner to the "Send" (message) button and deactivate it while sending
* update a complete Thing using "PATCH" and with the new 3.4.0 header "if-equal: skip-minimizing-merge"
* only send eTag if it could be retrieved when updating complete thing
* added missing "ilike" predicate to the search slot
* made UI more responsive for small screens
* prevent browser for doing autocomplete in the "search" input field
* increase size of the "Things" JSON editor - keep sizes and position of other JSON editors as they were
* fixes #1712 - by doing the "merge" on Ditto server side and subscribing complete thing changes instead of only deltas in the UI

### DIFF
--- a/ui/modules/things/thingMessages.ts
+++ b/ui/modules/things/thingMessages.ts
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+/* eslint-disable require-jsdoc */
+import * as API from '../api.js';
+import * as Environments from '../environments/environments.js';
+import * as Utils from '../utils.js';
+import messagesHTML from './thingMessages.html';
+import * as Things from './things.js';
+
+const dom = {
+  inputThingMessageSubject: null,
+  inputThingMessageTimeout: null,
+  inputThingMessageTemplate: null,
+  buttonThingMessageSend: null,
+  buttonThingMessageFavorite: null,
+  ulThingMessageTemplates: null,
+  favIconThingMessage: null,
+};
+
+let acePayload;
+let aceResponse;
+
+/**
+ * Initializes components. Should be called after DOMContentLoaded event
+ */
+export async function ready() {
+  Environments.addChangeListener(onEnvironmentChanged);
+
+  Utils.addTab(
+      document.getElementById('tabItemsThing'),
+      document.getElementById('tabContentThing'),
+      'Message to Thing',
+      messagesHTML,
+  );
+
+  Utils.getAllElementsById(dom);
+
+  acePayload = Utils.createAceEditor('acePayloadThingMessage', 'ace/mode/json');
+  aceResponse = Utils.createAceEditor('aceResponseThingMessage', 'ace/mode/json', true);
+
+
+  dom.buttonThingMessageSend.onclick = () => {
+    Utils.assert(dom.inputThingMessageSubject.value, 'Please give a Subject', dom.inputThingMessageSubject);
+    Utils.assert(dom.inputThingMessageTimeout.value, 'Please give a timeout', dom.inputThingMessageTimeout);
+    dom.buttonThingMessageSend.classList.add('busy');
+    dom.buttonThingMessageSend.disabled = true;
+    messageThing();
+  };
+
+  dom.buttonThingMessageFavorite.onclick = () => {
+    const templateName = dom.inputThingMessageTemplate.value;
+    Utils.assert(templateName, 'Please give a name for the template', dom.inputThingMessageTemplate);
+    Environments.current().messageTemplates['/'] = Environments.current().messageTemplates['/'] || {};
+    if (Object.keys(Environments.current().messageTemplates['/']).includes(templateName) &&
+        dom.favIconThingMessage.classList.contains('bi-star-fill')) {
+      dom.favIconThingMessage.classList.replace('bi-star-fill', 'bi-star');
+      delete Environments.current().messageTemplates['/'][templateName];
+    } else {
+      dom.favIconThingMessage.classList.replace('bi-star', 'bi-star-fill');
+      Environments.current().messageTemplates['/'][templateName] = {
+        subject: dom.inputThingMessageSubject.value,
+        timeout: dom.inputThingMessageTimeout.value,
+        payload: JSON.parse(acePayload.getValue()),
+      };
+      acePayload.session.getUndoManager().markClean();
+    }
+    Environments.environmentsJsonChanged('messageTemplates');
+  };
+
+  dom.ulThingMessageTemplates.addEventListener('click', (event) => {
+    if (event.target && event.target.classList.contains('dropdown-item')) {
+      dom.favIconThingMessage.classList.replace('bi-star', 'bi-star-fill');
+      const template = Environments.current().messageTemplates['/'][event.target.textContent];
+      dom.inputThingMessageTemplate.value = event.target.textContent;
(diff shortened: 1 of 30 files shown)


thjaeckle:
@thfries I don't know if you are still working with the Ditto UI.
If you could, would you be so kind and have a short look on the improvements/fixes I did for the upcoming Ditto 3.4.0 release?

ui/modules/things/thingMessages.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (Object.keys(Environments.current().messageTemplates['/']).includes(templateName) &&
+        dom.favIconThingMessage.classList.contains('bi-star-fill')) {
+      dom.favIconThingMessage.classList.replace('bi-star-fill', 'bi-star');
+      delete Environments.current().messageTemplates['/'][templateName];
+    } else {
+      dom.favIconThingMessage.classList.replace('bi-star', 'bi-star-fill');
+      Environments.current().messageTemplates['/'][templateName] = {
+        subject: dom.inputThingMessageSubject.value,
+        timeout: dom.inputThingMessageTimeout.value,
+        payload: JSON.parse(acePayload.getValue()),
```
thfries, line 74: Storing a template does not work, if the payload is empty (same is true for the messages on features). I did a commit to fix that but I have no permissions to push. Would it work with a patch or how is the procedure here?

ui/modules/things/thingMessages.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function onEnvironmentChanged(modifiedField) {
+  Environments.current()['messageTemplates'] = Environments.current()['messageTemplates'] || {};
+
+  if (!modifiedField) {
+    clearAllFields();
+  }
+  if (modifiedField === 'messageTemplates') {
+    refillTemplates();
```
thfries, line 136: there is no event to fill the template list initially. This can be fixed to remove the check for the modified field and call refillTemplates just on any environment change.

thfries (changes requested):
Hey @thjaeckle,
thank you for all these improvements. That is amazing!
- I like the big screen for the things editor. Was about to add that to connections and everywhere, because on the connections, the small editor is annoying. Would even make the editor full screen. Unfortunately on the connections and on the features, there are still two columns so the editor for me is still not wide enough. But very cool idea!
- Cool you fixed the responsive behavior! I also realized that this was not working but I had no clue how to fix it. Thanks a lot!
- I like the progress spinner on message sending. One other issue is pending unfortunately, that is annoying: The messages input fields are overwritten in case of SSE incoming for the same feature. I would like to fix that, too.

Very happy to see all those improvements.

ui/modules/things/things.html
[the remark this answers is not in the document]
thjaeckle: Hi @thfries 
I removed the top headlines where they did not collapse the top block - for "Things" and "Overview" in Policies this was not the case. So they kind of just "stole" space on the page, as you already have the navigation telling you where you are ..

I also agree that collapsing search does not really make sense.

So I think I would keep it the way I changed it to ..

ui/modules/things/thingMessages.ts
thjaeckle: @thfries yes, a patch would be great 👍 - I unfortunately cannot selectively give you permissions to make changes in this repo.

ui/modules/things/thingMessages.ts
thjaeckle: @thfries do you also have a solution and can add it too a patch?

ui/modules/things/thingMessages.ts
thfries: let's try: 
[0001-UI-improvements-review.patch](https://github.com/eclipse-ditto/ditto/files/12910260/0001-UI-improvements-review.patch)

ui/modules/things/thingMessages.ts
thfries: Yes that is included in the patch. Let me know, if the patch file works.

thjaeckle:
> * I like the big screen for the things editor. Was about to add that to connections and everywhere, because on the connections, the small editor is annoying. Would even make the editor full screen. Unfortunately on the connections and on the features, there are still two columns so the editor for me is still not wide enough. But very cool idea!

I could of course make the width even more than 60% ... let's say 90%?
And I could also add the mode for the connection and feature editor. For connections definitely that makes sense.  
For editing a feature I kind of liked this "inline style" - but you are right, because of the 2 columns, more width would definitely help here as well.

I will add those changes - I think it is better that way as well.

> One other issue is pending unfortunately, that is annoying: The messages input fields are overwritten in case of SSE incoming for the same feature. I would like to fix that, too.

Oh, I did not yet notice .. would you like to fix that soon? 
I planned the Ditto 3.4.0 release for Monday (tomorrow :D ).
But this could also be put into a "bugfix" release 3.4.1.

Speaking about release - are there any news about the "policy imports" for the UI?
Just asking .. no pressure ;)

ui/modules/things/thingMessages.ts
thjaeckle: Yes, worked - thanks .. I added the patch as commit 👍

thjaeckle:
I also managed to fix reported bug #1712 .. quite easily :)
And was even able to remove the "lodash" dependency as a result from the UI.

thfries:
> > * I like the big screen for the things editor. Was about to add that to connections and everywhere, because on the connections, the small editor is annoying. Would even make the editor full screen. Unfortunately on the connections and on the features, there are still two columns so the editor for me is still not wide enough. But very cool idea!
> 
> I could of course make the width even more than 60% ... let's say 90%? And I could also add the mode for the connection and feature editor. For connections definitely that makes sense. For editing a feature I kind of liked this "inline style" - but you are right, because of the 2 columns, more width would definitely help here as well.
> 
> I will add those changes - I think it is better that way as well.
> 
> > One other issue is pending unfortunately, that is annoying: The messages input fields are overwritten in case of SSE incoming for the same feature. I would like to fix that, too.
> 
> Oh, I did not yet notice .. would you like to fix that soon? I planned the Ditto 3.4.0 release for Monday (tomorrow :D ). But this could also be put into a "bugfix" release 3.4.1.
> 
> Speaking about release - are there any news about the "policy imports" for the UI? Just asking .. no pressure ;)

Hi @thjaeckle,
it seems it could be fixed very easily but I did not continue to test yesterday. Btw.: the ditto webpage has some issue with the certificate...
Regarding the feature editor: true, the inline style also is nice. I m thinking about to have a button for the user to switch the mode. and then just have different default modes on different editors...
yes, not much time for coding currently unfortunately, sorry... Still plan to continue

thfries (commented):
Hi @thjaeckle, just wanted to test my fix for the SSE overwriting the message input fields. I realized that incoming messages throw an exception now, because "old" ditto servers do not send the new _context field. It would be good if the UI keeps some backward compatibility, because we use the UI for different ditto servers on different versions. No smart idea though, besides expect and handle both types of SSE payload formats.

thfries:
Hi @thjaeckle, please find a patch to fix the issue I mentioned.
[0001-UI-fix-SSE-overwriting-message-input-fields.patch](https://github.com/eclipse-ditto/ditto/files/12912480/0001-UI-fix-SSE-overwriting-message-input-fields.patch)

Issue #1631, opened 2023-05-02
put-metadata does not work as expected since 3.x


demetz:
We recently upgraded from 2.x to the latest version. 

I am no longer able to set metadata for a single property:

```
url = 'https://xxx/api/2/things/[thingid]/features/myfeature/properties/myproperty/temperature'

metadata = [
    {
        "key":"/unit",
        "value": "°C"
    }
]

r = requests.put(url, headers={"Content-Type":"application/json", "put-metadata":json.dumps(metadata)}, auth=(ditto_user, ditto_pass),json=64.5)
```

Expected behavior: value is written and metadata is available `/features/myfeature/properties/myproperty/temperature/unit`

Actual behavior: only value is written

thjaeckle:
Hi @demetz
Thanks for reporting this issue.

There were some enhancements to metadata in Ditto 3.0, I assume some behavior may have changed.

Could you please try using no trailing slash in the "key", so:
```
{
        "key":"unit",
        "value": "°C"
    }
```

I could imagine that this somehow changed if  Iook at the docs.

demetz:
unfortunately, same result: ditto returns a `204` and the actual value is written, however no metadata is created.

specifying a nested key (e.g. `/test/unit`) for _metadata works as expected.

thjaeckle:
Could reproduce via sandbox:
```shell
curl --location --request PUT 'https://ditto.eclipseprojects.io/api/2/things/org.eclipse.ditto:lamp01/features/LampFeature/properties/powerConsumption' \
--header 'Authorization: Basic ZGl0dG86ZGl0dG8=' \
--header 'put-metadata: [{"key":"unit","value": "Watts"}]' \
--header 'Content-Type: application/json' \
--data '23.23'
```

Retrieving the metadata:
```shell
curl --location --request GET 'https://ditto.eclipseprojects.io/api/2/things/org.eclipse.ditto:lamp01?fields=thingId%2Cfeatures%2C_metadata' \
--header 'Authorization: Basic ZGl0dG86ZGl0dG8='
```

Did not apply the metadata as expected (`"powerConsumption"` is an empty object instead of having the `"unit"` key):
```json
{
    "thingId": "org.eclipse.ditto:lamp01",
    "features": {
        "LampFeature": {
            "properties": {
                "powerConsumption": 23.23
            }
        }
    },
    "_metadata": {
        "features": {
            "LampFeature": {
                "properties": {
                    "powerConsumption": {}
                }
[...]

PR #1774, opened 2023-10-15
#1631 fix putting metadata on a single property did not work

thjaeckle:
fixes: #1631


thjaeckle:
fix is quite easy .. I verified with a unit test that it is now working

PR #1779, opened 2023-10-23
disable reduced pom generation

alstanchev:
Reduced pom generation leads to missing license headers and all compile dependencies in a target built/released pom for the services modules.


thjaeckle (approved):
LGTM, good catch 👍

Issue #1780, opened 2023-10-25
Retrieving closed connections by tag fails


alstanchev:
Trying to retrieve a closed connection by tag with SudoRetrieveConnectionIdsByTag fails because the "connectionClosed" event persisted in the "connection_journal" does not contain tags and leads to skipping them.

PR #1781, opened 2023-10-25
#1780 fix update headers of ConnectionClosed event

alstanchev:
ditto headers are used for persisting tags, when missing there are no tags in the event persisted in the journal collection


thjaeckle (approved):
LGTM 👍

PR #1782, opened 2023-10-26
fix that ThingFieldSelector did not allow specific paths of "_metadata"

thjaeckle:
* it only allowed to select top-level "_metadata" - which does not make sense to limit, as even selecting only specific _metadata is already supported
* also add missing support for featureId wildcard when selecting e.g. "_metadata/features/*/properties"


thjaeckle:
Could you please have a look, @alstanchev @kalinkostashki

alstanchev (approved):
LGTM 👍

PR #1787, opened 2023-10-30
Move x-ditto-pre-authenticated and X-Forwarded-User to ingress.api.annotations

### LINKED ISSUE #1778: [Helm] Put x-ditto-pre-authenticated and X-Forwarded-User to ingress.api.annotations
I would suggest to add 
```
proxy_set_header X-Forwarded-User $remote_user;
proxy_set_header x-ditto-pre-authenticated "nginx:$remote_user";
```
directly in the default values.yaml at ingress.api.annotations.

Currently they are configured within the ingress.controller. I think the majority of the Kubernetes clusters out there have already set up a working nginx ingress controller for multiple different applications. If an additional application will be installed its normally a straight forward process of editing the values.yaml and make some minor configurations at the ingress section.
But with ditto it is currently not as straight forward as it could be because you have to figure out that these two headers have to be added.

I am not a full-time Kubernetes admin and it is possible that I make incorrect assumptions, please correct me if that is the case.

mladBlum:
resolves #1778

### DIFF
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -92,7 +92,7 @@ jobs:
         # available for the docker.io/kindest/node image
         # https://hub.docker.com/r/kindest/node/tags
         k8s:
-          - v1.26.10
+          - v1.26.6
           - v1.27.3
           - v1.28.3
     steps:
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -40,7 +40,7 @@ jobs:
           python-version: ${{ env.VERSION_PYTHON }}
           check-latest: true
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.4.0
+        uses: helm/chart-testing-action@v2.6.1
         with:
           version: ${{ env.VERSION_CHART_TESTING }}
       - name: Run chart-testing (list-changed)
@@ -63,9 +63,9 @@ jobs:
         # which a folder exists at
         # https://github.com/yannh/kubernetes-json-schema/
         k8s:
-          - v1.26.6
+          - v1.26.10
           - v1.27.3
-          - v1.28.0
+          - v1.28.3
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -92,9 +92,9 @@ jobs:
         # available for the docker.io/kindest/node image
         # https://hub.docker.com/r/kindest/node/tags
         k8s:
-          - v1.26.6
+          - v1.26.10
           - v1.27.3
-          - v1.28.0
+          - v1.28.3
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -109,7 +109,7 @@ jobs:
           python-version: ${{ env.VERSION_PYTHON }}
           check-latest: true
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.4.0
+        uses: helm/chart-testing-action@v2.6.1
         with:
           version: ${{ env.VERSION_CHART_TESTING }}
       - name: Run chart-testing (list-changed)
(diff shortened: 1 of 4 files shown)


thjaeckle (changes requested):
Thanks for the contribution, @mladBlum 

Please update the Chart version in Chart.yaml to the next micro-version.

thjaeckle:
@mladBlum on "master" the version is already on `version: 3.4.1` - sorry, did not see that.
In that case you'll have to bump to `version: 3.4.2` - even if 3.4.1 is not yet released ..

mladBlum:
> @mladBlum on "master" the version is already on `version: 3.4.1` - sorry, did not see that. In that case you'll have to bump to `version: 3.4.2` - even if 3.4.1 is not yet released ..

Sorry, that's on me, I should have checked it myself.

thjaeckle (approved):
LGTM - thanks a lot @mladBlum 
I hope you verified that this works as you need it to .. ;) 

I had to bump the version of `helm/chart-testing-action` - otherwise the chart linting was failing.

RockyMM:
Hey @mladBlum, could you explain what were you trying to achieve with this change?

For us it broke the Ditto authentication as the pre-authenticated header is now not reaching the gateway pod. I am planning to make a PR that would revert this, but maybe there is an opportunity to achieve your goals and our goals at the same time? 😁

mladBlum:
> Hey @mladBlum, could you explain what were you trying to achieve with this change?
> 
> For us it broke the Ditto authentication as the pre-authenticated header is now not reaching the gateway pod. I am planning to make a PR that would revert this, but maybe there is an opportunity to achieve your goals and our goals at the same time? 😁
> 
> Oh, I see #1778. Hm, maybe there could be another way how to do it.

Would be nice if we could solve this issue together. The same text you wrote applied for us before. I just placed the ditto headers into the ingress so that every ingress controller is working. I am really curious about the cause of your problem because this should be a more generic solution.

What is your setup?

RockyMM:
> What is your setup? Do you use the included ingress controller in the chart?

Yes, we do. I am guessing you are not using the ingress controller included with the chart.

For us, what I see is that `proxy_set_header` directives ended up in the `nginx.ingress.kubernetes.io/configuration-snippet` of the ingresses and previously they were part of the `location-snippet` of the `nginx-configuration` ConfigMap. The end result is that the gateway is complaining with 401. When I returned the `proxy_set_header` directive manually to the ConfigMap the gateway was satisfied. 

The included ingress controller is `ingress-nginx/controller` and we've pinned `v1.9.1` because our kluster is v1.28.

In my perspective, the ConfigMap is the best place for this, so I will provide a rework of the Helm deployment that places these headers in the `nginx-configuration` ConfigMap. Hopefully you can work with that too.

mladBlum:
> > What is your setup? Do you use the included ingress controller in the chart?
> 
> Yes, we do. I am guessing you are not using the ingress controller included with the chart.
> 
> For us, what I see is that `proxy_set_header` directives ended up in the `nginx.ingress.kubernetes.io/configuration-snippet` of the ingresses and previously they were part of the `location-snippet` of the `nginx-configuration` ConfigMap. The end result is that the gateway is complaining with 401. When I returned the `proxy_set_header` directive manually to the ConfigMap the gateway was satisfied.
> 
> The included ingress controller is `ingress-nginx/controller` and we've pinned `v1.9.1` because our kluster is v1.28.
> 
> In my perspective, the ConfigMap is the best place for this, so I will provide a rework of the Helm deployment that places these headers in the `nginx-configuration` ConfigMap. Hopefully you can work with that too.

Ok, thats a little bit strange. 
My assumption was that also the included ingress controller uses the proxy_set_header part in the ingress configuration. I'm very curious why he doesn't do that. 
Probably we could add the headers in the place where they were before and also in the ingress itself. So both variants work. I am just a little bit uncomfortable to dupplicate this configuration.

Please add me to the merge-request with the rework so i can check the changes.

PR #1788, opened 2023-10-30
#1748 Add honoTenantId configuration for HonoConnection.

### LINKED ISSUE #1748: Hono connection can't be created for Hono tenant with '.' in name
When trying to create a Hono connection for a Hono tenant that contains a dot in the name (e.g. `org.eclipse.packages.c2e` as used in the IoT packages cloud2edge chart), there is an error.
```
PUT /api/2/connections/org.eclipse.packages.c2e
{"status":400,
"error":"connectivity:connection.id.invalid",
"message":"The Connection ID 'org.eclipse.packages.c2e' is not valid.",
"description":"It must be conform to the regular expression: <[a-zA-Z0-9-_:]{1,100}>."}
```

This limitation is currently not documented.

I wonder whether there could be an alternative way to pass on the Hono tenant ID  in the connection data.
The current way of using the connection ID also doesn't look very intuitive to me.
Wouldn't it be possible to use a connection configuration containing something like
```
"specificConfig": {
  "honoTenant": "org.eclipse.packages.c2e"
} 
```
here?

calohmn:
This fixes #1748.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/hono/DefaultHonoConnectionFactory.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/hono/DefaultHonoConnectionFactory.java
@@ -16,8 +16,6 @@ import java.net.URI;
 import java.text.MessageFormat;
 import java.util.Set;
 
-import org.eclipse.ditto.connectivity.model.Connection;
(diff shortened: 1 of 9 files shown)


thjaeckle (approved):
Thanks a lot @calohmn for this contribution.

Looks good to me. 👍

PR #1792, opened 2023-11-03
fix preserving DittoHeaders when encountering a JsonParseException

thjaeckle:
* e.g. containing correlation-id, trace-context, etc.


alstanchev (approved):
That is a very nice improvement.
LGTM 👍

Issue #1768, opened 2023-10-08
Ditto does not acknowledge QoS 1 or 2 messages for MQTT 5 when there are several sources


dimabarbul:
Ditto does not acknowledge QoS 1 or 2 messages for MQTT 5 when there are several sources.

Setup:

- running Ditto
- running MQTT 5 broker, e.g. mosquitto
- MQTT 5 connection that, for example, updates device twin, with following config:
    - static specificConfig.clientId
    - qos 2 in source

Steps:
- send MQTT 5 message that should lead to device twin updated, with QoS 1 or 2

    ```shell
    # data topic
    mosquitto_pub --id CLI -V mqttv5 -t data -m $RANDOM -q 1
    # ditto topic
    mosquitto_pub --id CLI -V mqttv5 -t ditto -m '{"topic":"test/thing-01/things/twin/commands/merge", "headers": {"content-type":"application/merge-patch+json"}, "path":"/features/ditto/properties/value", "value":'$RANDOM'}' -q 2
    ```

Relevant logs from mosquitto:

```text
1696746966: New connection from 127.0.0.1:36164 on port 1883.
1696746966: New client connected from 127.0.0.1:36164 as CLI (p5, c1, k60).
1696746966: No will message specified.
1696746966: Sending CONNACK to CLI (0, 0)
1696746966: Received PUBLISH from CLI (d0, q2, r0, m1, 'data', ... (5 bytes))
1696746966: Sending PUBREC to CLI (m1, rc0)
1696746966: Received PUBREL from CLI (Mid: 1)
1696746966: Sending PUBLISH to ditto (d0, q2, r0, m27, 'data', ... (5 bytes))
1696746966: Sending PUBCOMP to CLI (m1)
1696746966: Received DISCONNECT from CLI
1696746966: Client CLI disconnected.
...
1696747363: New connection from 127.0.0.1:44862 on port 1883.
1696747363: New client connected from 127.0.0.1:44862 as CLI (p5, c1, k60).
1696747363: No will message specified.
1696747363: Sending CONNACK to CLI (0, 0)
1696747363: Received PUBLISH from CLI (d0, q2, r0, m1, 'ditto', ... (170 bytes))
1696747363: Sending PUBREC to CLI (m1, rc0)
1696747363: Received PUBREL from CLI (Mid: 1)
1696747363: Sending PUBLISH to ditto (d0, q2, r0, m36, 'ditto', ... (170 bytes))
[...]

PR #1794, opened 2023-11-05
Fix handling messages sent while ditto offline

### LINKED ISSUE #1767: MQTT 5 messages sent when Ditto was not connected are not processed
Messages that are sent to MQTT 5 broker (I used mosquitto) are delivered to Ditto, but are not processed.

Setup:
- running Ditto
- running MQTT 5 broker, e.g. mosquitto
- MQTT 5 connection that, for example, updates device twin, with following config:
    - static specificConfig.clientId
    - specificConfig.cleanSession = false
    - qos 1 or 2 in source

Steps:
- close connection, for example, using HTTP API  
    ```shell
    curl -u devops:foobar http://localhost:8080/api/2/connections/<CONN_ID>/command -d connectivity.commands:closeConnection -H 'content-type: text/plain'
    ```
- send MQTT 5 message that should lead to device twin updated, with QoS 1 or 2
- open connection, for example, using HTTP API  
    ```shell
    curl -u devops:foobar http://localhost:8080/api/2/connections/<CONN_ID>/command -d connectivity.commands:openConnection -H 'content-type: text/plain'
    ```

Connection I used to test:

```json
{
    "connectionType": "mqtt-5",
    "connectionStatus": "open",
    "uri": "tcp://mosquitto:1883",
    "specificConfig": {
        "clientId": "ditto",
        "cleanSession": false
    },
    "sources": [
        {
            "addresses": [ "data" ],
            "qos": 2,
            "authorizationContext": ["nginx:ditto"],
            "payloadMapping": ["js-data"]
        },
        {
            "addresses": [ "ditto" ],
            "qos": 2,
            "authorizationContext": ["nginx:ditto"]
        }
    ],
    "targets": [],
    "mappingDefinitions": {
        "js-data": {
            "mappingEngine": "JavaScript",
            "options": {
                "incomingScript": " function mapToDittoProtocolMsg( headers, textPayload, bytePayload, contentType) { return Ditto.buildDittoProtocolMsg( 'test', 'thing-01', 'things', 'twin', 'commands', 'merge', '/features/mqtt/properties/data', { 'content-type': 'application/merge-patch+json' }, { headers, contentType, textPayload, }); }",
                "outgoingScript": ""
            }
        }
    }
}
```

Messages I sent:

```shell
# update device twin using JS payload mapper
mosquitto_pub -V mqttv5 -t data -m $RANDOM -q 2
# update device twin using Ditto payload mapper
mosquitto_pub -V mqttv5 -t ditto -m '{ "topic": "test/thing-01/things/twin/commands/merge", "headers": {"Content-Type": "application/merge-patch+js
on"}, "path": "/features/mqtt/properties/ditto", "value": { "random": '$RANDOM' } }' -D publish content-type 'application/merge-patch+json' -q 2
```

dimabarbul:
Resolves #1767

Current version of the PR contains only integration tests, so it's draft PR.


dimabarbul:
I have some concerns regarding the PR, but at least it's at a point when the functionality looks working, so I mark it as ready for review.

### Solution Description

There are 2 kinds of messages that are lost:
1) Unlikely RabbitMQ, messages in MQTT are not stored in topics (or queues), they are stored in per-client sessions. So, when client that has existing session connects, it receives all missed messages **before** it subscribed to any topic. HiveMQ client in this case just auto-acknowledges the messages;
2) Messages that are sent after the code has subscribed to topic but before it has attached listener to Flowable of publishes. In this case HiveMQ issues a warning "No publish flow registered for MqttStatefulPublish" and acknowledges the message (see related [ticket](https://github.com/hivemq/hivemq-mqtt-client/issues/254)).

The implemented solution extracts consuming messages into dedicated interface - GenericMqttConsumingClient. This interface subscribes to **all** messages with manual acknowledgement **before the connection is established**. It is used then as source for per-connection-source messages (including messages from previous session that match corresponding topics, they will be processed) and as source for messages from previous session that the connection is no longer interested in (they will be just acknowledged).

### My Concerns about Implementation

1. rxJava2 error

Sometimes during tests I see following error in console:

```text
io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | com.hivemq.client.mqtt.exceptions.MqttSessionExpiredException: Session expired as connection was closed.
	at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
	at io.reactivex.subjects.PublishSubject$PublishDisposable.onError(PublishSubject.java:314)
	at io.reactivex.subjects.PublishSubject.onError(PublishSubject.java:243)
	at org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq.client.BufferingFlowableWrapper.lambda$new$1(BufferingFlowableWrapper.java:62)
	at io.reactivex.internal.subscribers.LambdaSubscriber.onError(LambdaSubscriber.java:79)
	at io.reactivex.internal.subscribers.BasicFuseableSubscriber.onError(BasicFuseableSubscriber.java:101)
	at io.reactivex.internal.operators.flowable.FlowableOnErrorNext$OnErrorNextSubscriber.onError(FlowableOnErrorNext.java:90)
	at io.reactivex.internal.subscriptions.EmptySubscription.error(EmptySubscription.java:55)
	at io.reactivex.internal.operators.flowable.FlowableError.subscribeActual(FlowableError.java:40)
	at io.reactivex.Flowable.subscribe(Flowable.java:14935)
	at io.reactivex.Flowable.subscribe(Flowable.java:14882)
	at io.reactivex.internal.operators.flowable.FlowableOnErrorNext$OnErrorNextSubscriber.onError(FlowableOnErrorNext.java:115)
	at io.reactivex.internal.operators.flowable.FlowableObserveOn$BaseObserveOnSubscriber.checkTerminated(FlowableObserveOn.java:197)
	at io.reactivex.internal.operators.flowable.FlowableObserveOn$ObserveOnSubscriber.runAsync(FlowableObserveOn.java:399)
	at io.reactivex.internal.operators.flowable.FlowableObserveOn$BaseObserveOnSubscriber.run(FlowableObserveOn.java:176)
	at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
	at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: com.hivemq.client.mqtt.exceptions.MqttSessionExpiredException: Session expired as connection was closed.
Caused by: com.hivemq.client.mqtt.mqtt3.exceptions.Mqtt3DisconnectException: Client sent DISCONNECT
Exception in thread "RxComputationThreadPool-4" io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | com.hivemq.client.mqtt.exceptions.MqttSessionExpiredException: Session expired as connection was closed.
	at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
	at io.reactivex.subjects.PublishSubject$PublishDisposable.onError(PublishSubject.java:314)
	at io.reactivex.subjects.PublishSubject.onError(PublishSubject.java:243)
	at org.eclipse.ditto.connectivity.service.messaging.mqtt.hivemq.client.BufferingFlowableWrapper.lambda$new$1(BufferingFlowableWrapper.java:62)
	at io.reactivex.internal.subscribers.LambdaSubscriber.onError(LambdaSubscriber.java:79)
	at io.reactivex.internal.subscribers.BasicFuseableSubscriber.onError(BasicFuseableSubscriber.java:101)
	at io.reactivex.internal.operators.flowable.FlowableOnErrorNext$OnErrorNextSubscriber.onError(FlowableOnErrorNext.java:90)
	at io.reactivex.internal.subscriptions.EmptySubscription.error(EmptySubscription.java:55)
	at io.reactivex.internal.operators.flowable.FlowableError.subscribeActual(FlowableError.java:40)
	at io.reactivex.Flowable.subscribe(Flowable.java:14935)
	at io.reactivex.Flowable.subscribe(Flowable.java:14882)
```

The solution is described [at page mentioned in error](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling). I already check if the object is disposed before raising error, so solution seems to be adding global error handler which sounds awkward to me. Also previously the code used rxJava, but did not have the handler, so I guess, I might be using it wrong. Suggestions are welcome.

2. BufferingFlowableWrapper class

In general I'm not sure if using Subject is OK for rxJava. I've struggled many weeks to create the code that buffers items until it's told to stop, at least, it passes the tests. It might be better to create custom Subject, not sure.

3. Explicit timings in MqttClientActorIT

I don't like specifying timings explicitly. Not only they are specific to machine where the tests run, but also they disregard [duration scale](https://nightlies.apache.org/pekko/docs/pekko/main-snapshot/docs/testing.html#accounting-for-slow-test-systems).
I did that because default message timeout (3 seconds) did not work for me for connected/disconnected events. I believe, this is because here we use real MQTT broker and it takes more time than in unit tests (for my machine the events were raised in about 10 seconds).
So, I didn't find good solution. What I could do is to add multiplying the timings by `pekko.test.timefactor` but that seems more like hack than solution.

4. Amount of test cases in MqttClientActorIT

Probably, we could minimize number of tests by removing some test methods.

dimabarbul:
Looks like the changes fix #1768 as well

thjaeckle (approved):
Hey @dimabarbul 

Sorry, that it took a little, but not I had the chance to have a look at the PR.
I must say, that is the highest quality contribution to Ditto I ever saw. Thanks a lot for this.
The abstraction for the integration-tests, starting containers (Mongo/Mosquitto) is really great to have and very well done 👍 

So I am really glad you fixed the issue(s) and improved testability by doing so.

One class may be "final" (which I commented inline), but the rest: LGTM!

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/client/BufferingFlowableWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+import io.reactivex.subjects.PublishSubject;
+
+/**
+ * Wrapper around flowable that buffers the items until it is told to stop.
+ * When buffering is enabled, all subscribers get all missed items, when the
+ * buffering is disabled, subscribers will get only new items.
+ *
+ * @param <T> type of items
+ */
+public class BufferingFlowableWrapper<T> implements Disposable {
```
thjaeckle, line 29:
```suggestion
public final class BufferingFlowableWrapper<T> implements Disposable {
```

PR #1798, opened 2023-11-08
fix that JsonObjectBuilder.remove removes to much in certain cases

thjaeckle:
This looks like a really small bug, but is has a huge impact:
* Currently, for `ThingMerged` events the payload `value` can get "lost" before persisting a merge because of this bug

Here an inline policy is removed from the to-be-persisted JSON:
https://github.com/eclipse-ditto/ditto/blob/d008959c019a966af0c1b152c12f953002669c58/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/serializer/ThingMongoEventAdapter.java#L51-L53

The JsonPointer to be removed is `/value/_policy` - if this `_policy` however does not exist *and* if the `value` does not even contain a `JsonObject` (which can be the case when using "merge", e.g. only updating one specific path to e.g. a JsonNumber), *then* the current implementation of `ImmutableJsonObjectBuilder.remove` will remove the `value` from the event.  
That leads to the event missing its `value` in persistence and that leads to much pain and suffering when replaying events for that persistence actor, throwing exceptions on deserialization if the `value` is missing.

### DIFF
--- a/json/src/main/java/org/eclipse/ditto/json/ImmutableJsonObjectBuilder.java
+++ b/json/src/main/java/org/eclipse/ditto/json/ImmutableJsonObjectBuilder.java
@@ -175,7 +175,7 @@ final class ImmutableJsonObjectBuilder implements JsonObjectBuilder {
                         rootObject = rootObject.remove(nextPointerLevel);
                         set(JsonFactory.newField(jsonField.getKey(), rootObject,
                                 jsonField.getDefinition().orElse(null)));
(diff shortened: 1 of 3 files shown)


thjaeckle:
@alstanchev @kalinkostashki a review would be nice - this is a really fundamental issue when using "MergeThing" commands :/

alstanchev (approved):
LGTM 👍

PR #1802, opened 2023-11-10
#1801 Add trusted certificates configuration for HonoConnection.

### LINKED ISSUE #1801: HonoConnection config missing trusted certificates configuration
When configuring a Kafka connection, the `ca` property can be used to set the trusted certificates for server certificate validation.

For a Hono Connection, where the Kafka-server related configuration is done as part of the connectivity service configuration, a `ca` property currently can't be set.

To let the Hono Connection be used as an alternative to a Kafka connection for connecting to Hono, such a property configuration option should be added.

calohmn:
Also fix table of contents in Hono Connection documentation.

This fixes #1801.


thjaeckle (approved):
LGTM, thanks a lot for the contribution :+1:

PR #1805, opened 2023-11-12
UI - Allow editors in full screen mode

thfries:
Hello @thjaeckle,

I reworked the editor size behavior:
- all editors allow a new `fullscreen` attribute
- while editor is active, you can toggle the mode. The mode per editor is remembered by the component
- fixed also the scrolling behavior in full screen mode. Full screen is full screen - no scrolling.

Please let me know if that works for you.


thjaeckle (changes requested):
Hi @thfries 
For me, the "full screen" button (for Thing editor, but also Feature editor) apparently does not work. I am using Chrome and when I click it, I no longer see an editor at all, only the "greyed out" background.
With which browser did you test it?

thfries:
Hi @thjaeckle, that is strange. I am also testing with Chrome. Could that be caused by old cached css classes? Can you try shift-reload in Chrome? If that is the reason, we need some way to force the Browser to reload somehow...

The error on the console is the same as on the policy-Imports PR caused by things without features. Fix will be merged there.

thjaeckle (approved):
Aaah, it works now 👍 

I think I did not restart `npm run start` when switching reviewing from the other UI PR.
I did that now and it works.

So looks good to me, also a nice addition 👍

PR #1810, opened 2023-11-21
fix that non-successful message responses were not displayed in ACE editor

thjaeckle:
* only successful responses were displayed
* however, messages might e.g. be answered with a 400 (bad request) and useful information as response

### DIFF
--- a/ui/modules/api.ts
+++ b/ui/modules/api.ts
@@ -315,10 +315,16 @@ export function setAuthHeader(forDevOps) {
  * @param {Object} additionalHeaders object with additional header fields
  * @param {boolean} returnHeaders request full response instead of json content
  * @param {boolean} devOps default: false. Set true to avoid /api/2 path
+ * @param {boolean} returnErrorJson default: false. Set true to return the response of a failed HTTP call as JSON
  * @return {Object} result as json object
  */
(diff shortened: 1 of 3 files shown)


thjaeckle:
@thfries I noticed that only successful message responses were displayed in the "response" ace editor
We use e.g. 400 "bad request" for validation errors and it would be really helpful to display such responses.

Could you please have a short look on the improvement/fix?

I also thought that displaying the status code might also be helpful, but had no good idea to add it to the current layout.

thfries:
Hi @thjaeckle, I am missing the usual error toast now that is popping up in case of errors but I did not find a way to show it in addition to filling the response editor.
The toast with the error message did not work for you? By clicking on it you can prevent it from disappearing while you keep the mouse on it.
Can you please take a look at the toast behaviour? May be this can be / needs to be improved. That should work all errors from ditto in a similar way. Thanks for you feedback.

thjaeckle:
Thanks, will look into the toasts for other errors ..

thjaeckle:
@thfries I now additionally show the "error toast" and improved that, displaying also errors which are not in "Ditto error format".

thfries:
Nice! I like it. Thank you!

PR #1812, opened 2023-11-22
display attributes in UI in an aceEditor instead of a simple inputfield

thjaeckle:
* for attributes which were JsonObjects or JsonArrays, displaying them in a single "input" field was very umcomfortable

This PR shows all attributes also in an aceEditor, while still showing the string representation in the attributes table.

@thfries I would appreciate if you could have a look

### DIFF
--- a/ui/modules/things/attributes.ts
+++ b/ui/modules/things/attributes.ts
@@ -20,11 +20,11 @@ import * as Things from './things.js';
 const dom = {
   tbodyAttributes: null,
   crudAttribute: null,
-  inputAttributeValue: null,
   badgeAttributeCount: null,
 };
 
 let eTag;
+let attributeEditor;
 
 /**
  * Initializes components. Should be called after DOMContentLoaded event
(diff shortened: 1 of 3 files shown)


ui/modules/things/things.html
```diff
@@ -92,24 +92,26 @@ <h5 data-bs-toggle="collapse" data-bs-target="#collapseAttributes">Attributes <s
 <hr />
 <div class="collapse" id="collapseAttributes">
     <div class="row">
```
thfries, line 94: Please add class `resizable_pane` and `style="height: 20vh;"` so that a minimum height is given (if there is no or only one or two attributes). The height of the ace editor is then resizable.

ui/modules/things/attributes.ts
```diff
@@ -105,10 +110,10 @@ function refreshAttribute(thing, attributePath = null) {
 
   if (thing) {
     dom.crudAttribute.idValue = attributePath;
-    dom.inputAttributeValue.value = attributeToString(thing.attributes[attributePath]);
```
thfries, line 108: `attributeToString` is no longer called. I have seen that simple strings are displayed with double quotes and same for numbers (see e.g. a coffeebrewer Thing with attribute `42` or `Speaking coffee machine`). Can we achieve that you can still edit simple numbers or simple strings?

thfries (changes requested):
Hi @thjaeckle, 

I have two comments. Rest is very cool 👍. Thank you!

ui/modules/things/attributes.ts
thjaeckle: I wouldn't do that - the ACE editor is in "JSON mode" - if you put in a string without the quotes, this is no longer valid JSON (and the ACE editor will complain).
Numbers are displayed as numbers without quote (if they are saved as a JSON number in the attribute).

In JSON, `"42"` is something completely different to `42` - I think before you could not even see (and make) that difference.

The attributes table still shows the `attributeToString` representation - which is great to get a quick overview.

It is much more consistent with all other editors on the UI as well.

thfries: Fine for me. I just checked: The sample "coffeebrewer" sample really has a `"42"` as string and not as an integer. And you still can set an integer attribute with the json editor. So that confused me. So you just need the quotes for "string attributes". All good. Thank you.

PR #1813, opened 2023-11-22
enhance "Incoming Thing Updates" in UI to provide more information at a glance

thjaeckle:
* added 2 columns "Action" (e.g. showing "merged" or "modified") and "Path" (showing on which level) to the table
* prioritized "_context" over "features" and "attributes" to extract "Field(s)" from - only done for actions on path "/"
* added a selection to the "Thing Update Detail" view which information to display from the selected event

@thfries again, I would appreciate your input on this :)

I noticed you swapped reading from either "_context" or features/attributes in `getColumnValues()` method recently.
That does not work, for newer Ditto versions both are present, so we have to first check for existence of `["_context"].value` (which Ditto 3.4 provides) and can then fall-back to plain features/attributes to determining the updated "Field(s)".

This PR improves displaying "merge" and also "delete" events a lot - because "Action" and "Path" are shown in the table.

### DIFF
--- a/ui/modules/things/messagesIncoming.html
+++ b/ui/modules/things/messagesIncoming.html
@@ -15,7 +15,7 @@
 <hr />
 <div class="collapse" id="collapseMessages">
     <div class="row resizable_pane" style="height: 45vh;">
-        <div class="col-md-6 resizable_flex_column">
+        <div class="col-md-7 resizable_flex_column">
             <div class="input-group input-group-sm mb-1">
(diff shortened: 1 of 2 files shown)


thfries:
Works like a charm. Nice!
Happy to see that "growing".
We may show that tab uncollapsed per default?

LGTM

thjaeckle:
I think it is OK to be collapsed by default, as on new events there will be the indicator next to the heading of the section that there is something happening there.

Issue #1583, opened 2023-02-23
Search in history of a thing using RQL filter


thjaeckle:
With the addition of #1498 (to be released in Ditto 3.2.0), it is possible to get a stream of all modification events of a thing (up to the configured `history-retention-duration` (e.g. 30 days).

It would be quite easy to enhance this new API in order to find a "special condition" within the history of a thing.
Examples:
* stream all events matching a given condition (`filter` as RQL)
  * e.g. in which a specific property was modified (via the SSE API):  
    `/api/2/things/namespace:my-thing1?from-historical-revision=1&filter=exists(feature/myFeature/properties/targetProperty)`

Additional idea (optional):
* return the state of a twin when a certain condition (`filter` as RQL) matched
   * e.g. get state of complete twin when a given threshold was "first" reached:
     `GET /api/2/things/namespace:my-thing1?filter-first-historical-revision=gt(feature/myFeature/properties/value,42)`
   * e.g. get state of complete twin when a given threshold was "last" reached:
     `GET /api/2/things/namespace:my-thing1?filter-last-historical-revision=gt(feature/myFeature/properties/value,42)`
   * e.g. get timestamp when a given property was "last" modified:
     `GET /api/2/things/namespace:my-thing1?filter-last-historical-revision=exists(feature/myFeature/properties/value)&fields=_revision,_modified`

PR #1815, opened 2023-11-23
#1583 apply RQL based filtering when streaming "historical" thing events

thjaeckle:
Resolves: #1583

Open:

- [x] unit tests
- [x] system tests - see eclipse-ditto/ditto-testing#5
- [x] documentation


thjaeckle:
This one would be ready to be reviewed, @alstanchev @kalinkostashki 
If you find the time, that would be great.

alstanchev (approved):
LGTM 👍

PR #1816, opened 2023-11-23
fix that a "wrong" JsonPointer in a search RQL query lead to ERRORs

thjaeckle:
* instead, provide a BAD REQUEST (status 400) when doing a search with an RQL containing e.g. double slashes


thjaeckle:
@alstanchev @kalinkostashki could you please have a quick look?

alstanchev (approved):
LGTM 👍

PR #1817, opened 2023-11-25
UI - fix operations new empty logger

### LINKED ISSUE #1809: Enhancement Proposal: Addressing Operations Tab UI Issue with Logger Creation and Level Buttons
There is an issue with the Operations tab UI. I am uncertain whether you have intentionally retained this feature. In the text field for adding a new logger, I suggest disabling the level buttons. I observed that, even when the text box is empty, it is possible to send a PUT request using the buttons, thereby creating an empty logger with associated levels.

thfries:
Hi, please find this PR to fix #1809

@thjaeckle: btw: don't we need an option to delete a logger, too?

### DIFF
--- a/ui/modules/operations/servicesLogging.html
+++ b/ui/modules/operations/servicesLogging.html
@@ -34,18 +34,19 @@
                     crossorigin="anonymous" referrerpolicy="no-referrer" />
                 <link href="./dist/main.css" rel="stylesheet" />
 
-                <div class="input-group input-group-sm mb-1">
-                    <input class="form-control" type="text" disabled id="inputLogger" spellcheck="false"></input>
+                <div class="input-group input-group-sm mb-1 has-validation">
+                    <input class="form-control me-1" type="text" disabled id="inputLogger" spellcheck="false"></input>
                     <div class="btn-group" role="group">
-                        <input type="radio" class="btn-check" id="debug" autocomplete="off">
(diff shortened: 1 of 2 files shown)


thjaeckle:
@thfries when I remember correctly, deleting a logger or setting its log level to OFF was problematic. 
At least not all off loggers could be queried and displayed, because the logger api Ditto uses would return for log level "off" basically all loggers which don't have a defined log level.
And those are thousands.. 😕

Or did you mean something else?

thfries:
I was testing to create new loggers like "foobar" (which is invalid, too) and was not able to "undo" that operation. In general you can create additional loggers but there is no way to undo.
Log level "OFF" is currently not available in the UI.
May be this is exactly what I mean: if you would set the log level to "OFF" it would disappear from the list? We could show a confirmation modal in this case.

thjaeckle:
> I was testing to create new loggers like "foobar" (which is invalid, too) and was not able to "undo" that operation. In general you can create additional loggers but there is no way to undo. Log level "OFF" is currently not available in the UI. May be this is exactly what I mean: if you would set the log level to "OFF" it would disappear from the list? We could show a confirmation modal in this case.

Yes, that is a good idea:
* make "off" configurable as log level
* show a confirmation that - in order to show/activate the logger again - it must be explicitly added as a new logger again

PR #1822, opened 2023-11-27
UI - fix for #1821 (401 on startup)

### LINKED ISSUE #1821: UI - Bug: Piggyback commands cause 401 error when you start the UI
The piggyback commands try to load the services on startup of the UI. As piggyback requires devops user and devops is not used at startup, this always fails with one 401 error.

`await loadServicesAndInstances();` should not be called in `ready` directly. Instead it should be wrapped in a `TabHandler`. This makes sure that it will be called when you activate the tab and also makes sure that devops user is set.

thfries:
Pls find this PR to fix #1821


thfries:
Fix is tested now and it works. Removing draft state.

thjaeckle (approved):
LGTM 👍 

Thanks a lot

PR #1823, opened 2023-11-27
removed X-Original-URI

kalinkostashki:
- currently this header isn't handled by ditto
 - if someone were to make a large request with many thingIds in a GET request the X-Original-URI will be passed as a Ditto Header in the protocol message causing a DittoHeadersTooLarge exception -> http error 431


thjaeckle (approved):
LGTM 👍

PR #1824, opened 2023-11-27
preserve maxPid in journal aggregation

alstanchev:
in the rare case where all or most of the elements in an aggregation are pruned for being deleted it is not possible to preserve maxPid which in turn makes it impossible to iterate the whole collection.

the aggregation pipeline:

```
[{$match: {
 pid: {
  $gt: ''
 }
}}, {$sort: {
 pid: 1,
 sn: -1
}}, {$limit: 200}, {$group: {
 _id: '$pid',
 sn: {
  $first: '$sn'
 },
 __lifecycle: {
  $first: '$s2.__lifecycle'
 },
 _revision: {
  $first: '$s2._revision'
 },
 policyId: {
  $first: '$s2.policyId'
 },
 _modified: {
  $first: '$s2._modified'
 }
}}, {$sort: {
 _id: 1
}}, {$group: {
 _id: null,
 m: {
  $max: '$_id'
 },
 i: {
  $push: '$$ROOT'
 }
}}, {$redact: {
 $cond: {
  'if': {
   $ne: [
    '$__lifecycle',
    'DELETED'
   ]
  },
  then: '$$DESCEND',
  'else': '$$PRUNE'
 }
}}]
```

### DIFF
--- a/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/streaming/MongoReadJournal.java
+++ b/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/streaming/MongoReadJournal.java
@@ -28,6 +28,7 @@ import java.util.stream.Stream;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonInt32;
+import org.bson.BsonNull;
 import org.bson.BsonString;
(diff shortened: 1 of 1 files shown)


thjaeckle (approved):
Took a brief look.
Did not deeply look into the how and why this works and the former didn't, but I trust in you and the system tests

alstanchev:
Tests are green

PR #1832, opened 2023-11-30
migrate logging to SLF4J 2.0.x and Logback 1.4.x

thjaeckle:
Fixes "High" CVE reported in Logstash 1.2:
https://github.com/advisories/GHSA-vmq6-5m68-f53m


thjaeckle:
@alstanchev @kalinkostashki this would be ready to be reviewed
There is a new CVE for logback version 1.2, which Ditto still uses.

I had also to update to SLF4J 2.x - which seems to work just fine (after some adjustments in the `logback.xml` regarding conditions .. 

I think we should also backport to Ditto 3.4.x, what do you think?

thjaeckle:
The security issue was also backported to logback 1.2 (and the fix is already included in the latest 3.4.x bugfix release of Ditto), I however think that we should migrate to slf4j 2 and logback 1.4 eventually - and why not with Ditto 3.5? :)

PR #1834, opened 2023-12-01
#1883 return pre-authenticated configuration to location-snippet

### LINKED ISSUE #1833: Regression: When using BASIC auth the pre-authenticated header is not forwarded to gateway pod
The changes introduced in #1787 are seemingly incompatible with the built-in ingress controller, which is the [Ingress-Nginx Controller](https://kubernetes.github.io/ingress-nginx/) maintained by Kubernetes community. While somewhat compatible with the "official" [NGINX Ingress Controller](https://docs.nginx.com/nginx-ingress-controller/) maintained by the NGINX the company, there are some incompatibilities. The change from #1787 is one of them. Juding by the https://github.com/kubernetes/ingress-nginx repo, the `proxy-set-header` is always part of the location and is not used anywhere else, therefore it seems that it must be part of the `location-snippet`.

The clean way to configure `proxy-set-headers` in the *Ingress-Nginx Controller* is to create a [ConfigMap](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-set-headers) with the headers and to reference them in the central `nginx-configuration` ConfigMap.

I decided actually to simply return the strings where they were, but also to keep them in the location where #1787 introduce them, in order to keep Ditto working with the Ingress controller that @mladBlum is using.

RockyMM:
Necessary for the Ingress-Nginx Controller. Fixes #1833

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.4.2  # chart version is effectively set by release-job
+version: 3.4.3-0  # chart version is effectively set by release-job
 appVersion: 3.4.2
 keywords:
   - iot-chart
--- a/deployment/helm/ditto/templates/nginx-ingress.yaml
+++ b/deployment/helm/ditto/templates/nginx-ingress.yaml
@@ -132,6 +132,11 @@ data:
     # ignore X-Original-URI in the request
     proxy_hide_header X-Original-URI;
 
+
+    # set ditto-specific forwarded headers - needed in the location for registry.k8s.io/ingress-nginx/controller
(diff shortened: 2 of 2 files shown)


RockyMM:
@mladBlum take a look, I think this change should not affect you.

RockyMM:
> Hm, should I increase the chart version?

The next semver should be `3.4.2-1`. I kept the `appVersion` at `3.4.2`. Maybe there is a way to only release the Helm chart?

thjaeckle:
> @RockyMM do you know if we can use version 3.4.2.1 in Helm? The forth digit would be nice in this case ..

Probably not .. so `3.4.3-0` would be fine for me (if that passes linting) - as a "pre-release of 3.4.3".
Otherwise, just bump it to `3.4.3` - then Chart and appVersion are no longer in sync until the next minor release `3.5.0`.

thjaeckle:
@kalinkostashki could you review? I know you also were looking into solutions..

kalinkostashki (approved):
LGTM :+1:

thjaeckle (approved):

PR #1835, opened 2023-12-05
update logback 1.2 to mitigate CVE

thjaeckle:


thjaeckle:
@alstanchev @kalinkostashki could you approve?
I would create a 3.4.3 bugfix release with just this change and the fix for the Helm chart ingress.

alstanchev (approved):
LGTM

Issue #1844, opened 2023-12-18
Fetching a thing at a non-existing historical revision (e.g. in future) results in status 503


thjaeckle:
Example request:
```
curl 'https://ditto.eclipseprojects.io/api/2/things/foo:bar?fields=thingId,_revision,attributes&at-historical-revision=785'
```
Response after 15 seconds:
```json
{"status":503,"error":"ask.error","message":"There was an internal timeout.","description":"Please retry the performed action in order to improve resiliency."}
```

And even worse than that - during the 15 seconds, all other requests to that Thing will fail, as all access is prevented as part of the `at-historical-revision` calculation.

PR #1845, opened 2023-12-18
#1844 fix resolving revision or timestamp from the future was not directly denied

thjaeckle:
Fixes: #1844


thjaeckle:
@alstanchev @kalinkostashki Could you please review?
Small fix, huge impact when using the history api.

alstanchev (approved):
LGTM 👍

Issue #1656, opened 2023-06-15
Provide "debug" option when creating a Thing with a referenced WoT TM, providing error feedback


thjaeckle:
When developing new WoT TMs and testing if the JSON skeleton created by Ditto is as expected it would be very helpful to get meaningful error messages, e.g. when:
* the model or some references submodels / referenced models cannot be retrieved
* when the TM is in invalid format / cannot be parsed
* etc.

Currently, Ditto "fails silently" and just does not create the JSON skeleton (which is also a good approach to be fault tolerant).

However, for this specific case it would make sense to configure a "strict WoT mode", e.g. via query-parameter or header field.

PR #1847, opened 2023-12-19
Enhance WoT skeleton creation with more options

thjaeckle:
* whether to generate defaults for "optional" WoT TM properties
* whether to throw exceptions related to WoT models instead of silently swallowing them

Also resolves: #1656


thjaeckle:
@alstanchev @kalinkostashki  if you find the time to review, it would be great

alstanchev (approved):
LGTM 👍

PR #1848, opened 2023-12-21
fixed that "condition" query param could not be provided as form field

thjaeckle:
* which prevented long conditions to be used via POST /search/things


thjaeckle:

When #1706 was done, it was forgotten to also expose the `condition` as form parameter - which leads to the same problem #1706 aimed to solve: potentially too long query parameter when passing a "condition" via query-param or HTTP-header.

PR #1849, opened 2023-12-22
benchmark tool improvements

vvasilevbosch:
Fix maxWait value to match the newest version of xk6-kafka
Use TestRun kubernetes resource, instead of k6, since it will be deprecated 
Add max retries on connection status fetching(CONNECTION_OPEN_MAX_RETRIES), after which the test aborts execution 
Add Service for mmock
Use better naming for some env variables
Add sample helm values.yaml for ditto
Add --quiet false option for full logging when running k6 test inside kubernetes
Remove AUTH_CONTEXT env var, value is equal to DITTO_PRE_AUTHENTICATED_HEADER_VALUE


thjaeckle (approved):
Looks good to me, thanks a lot @vvasilevbosch for improving this further 👍

Issue #1854, opened 2024-01-02
Provide time placeholders in RQL with ability to specify plus and minus "duration"


thjaeckle:
Currently, there is the `time:now` placeholder which can be used in RQL queries, e.g. like this:
```
ge(features/ConnectionStatus/properties/status/readyUntil,time:now)
```

The currently implemented time placeholders are:
* `time:now`
* `time:now_epoch_millis`

It would be very useful to be able to also specify a plus and minus offset to "now":
* e.g. querying all of the things which were modified in the last 1 day
* e.g. counting all of the things which have a ISO8601 timestamp persisted which is more recent than 2 hours

Syntax idea:
To keep it simple, the existing placeholders could support optionally the following syntax:
```
<sign> := + | -

<unit> := "ms" | "s" | "m" | "h" | "d"
<dittoDuration> := <integer><unit>

<offset> := <sign><dittoDuration>

time:now<offset>
time:now_epoch_millis<offset>
```
Where Ditto duration is defined by implementation class `DittoDuration`.

Examples:
* `time:now-5h` - current ISO8601 timestamp minus 5 hours
* `time:now_epoch_millis+2d` - current epoch timestamp plus 2 days
* `time:now+30s` - current ISO8601 timestamp plus 30 seconds

thjaeckle:
Additional idea: Rounding up/down to the next full `"ms" | "s" | "m" | "h" | "d"`.

Syntax could be e.g.:
```
time:now_round-d       # rounding down to "todays" start of the day
time:now-1d_round-d    # rounding down to yesterday's start of the day
time:now-1d_round-h    # rounding down to yesterday's start of the current hour

time:now_round+m       # rounding up to the next full minute
```

PR #1856, opened 2024-01-04
#1854 enhance time:now* placeholders to calculate plus and minus from now

thjaeckle:
* and to optionally truncate to a unit (rounding down)

Resolves: #1854

Open:

- [x] documentation

### DIFF
--- a/placeholders/src/main/java/org/eclipse/ditto/placeholders/ImmutableTimePlaceholder.java
+++ b/placeholders/src/main/java/org/eclipse/ditto/placeholders/ImmutableTimePlaceholder.java
@@ -13,13 +13,15 @@
 package org.eclipse.ditto.placeholders;
 
 import java.time.Instant;
-import java.util.Arrays;
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.List;
 
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 import org.eclipse.ditto.base.model.common.ConditionChecker;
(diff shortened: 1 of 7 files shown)


placeholders/src/main/java/org/eclipse/ditto/placeholders/ImmutableTimePlaceholder.java
```diff
@@ -42,8 +44,12 @@ final class ImmutableTimePlaceholder implements TimePlaceholder {
     private static final String NOW_PLACEHOLDER = "now";
     private static final String NOW_EPOCH_MILLIS_PLACEHOLDER = "now_epoch_millis";
 
-    private static final List<String> SUPPORTED = Collections.unmodifiableList(
-            Arrays.asList(NOW_PLACEHOLDER, NOW_EPOCH_MILLIS_PLACEHOLDER));
+    private static final String TRUNCATE_START;
+    private static final String TRUNCATE_END = "]";
+
+    static {
```
alstanchev, line 50: Is there a need for this static block?

alstanchev (approved):
LGTM 👍

placeholders/src/main/java/org/eclipse/ditto/placeholders/ImmutableTimePlaceholder.java
thjaeckle: Yeah, that was kinda strange. I could not declare `"["` directly to a static constant.
IbtelliJ was complaining.
But when I moved the assignment into the static block, it worked..

Could also be an IDE bug.. 😕

Issue #1818, opened 2023-11-26
UI: Add search filter option to incoming messages and connection logs


thfries:
In order to analyze and navigate in long lists of incoming message table and also in connection logs, it would be good to have a search filter in top of the table

example filters for connection logs:
- `direction:inbound type:consumed level:success`
- `timestamp<1.1.2900`
- `correlationId:4711`
- `thingId:234235`

example filters for incoming messages:
- feature path that was updated if you want to watch a specific feature
- ?

The filter should be easily filled by e.g clicking on the connection metrics or somehow on the correlationId/thingId of one log entry.
In case of incoming messages the filter should be set somehow by selecting a feature.

thfries:
@thjaeckle, pls assign this issue to me. Thanks

thjaeckle:
@thfries good idea, I also was thinking about that.
For the incoming messages, you could even provide an RQL expression and do server side filtering (when opening the SSE / EventSource).

I assume you want to do client side filtering?

thfries:
Yes, I just thought about client side filtering - for the same reason as you mentioned. I see the purpose more like a log for debugging. So you don't know before what you are searching for. I thought already that it is a drawback that you need to select the thing before that you want to "watch".

Since the fix for #1712, the SSE has become verbose. I remember now that before we merged the "delta" into the known state of the thing on the client-side.

Can you nowadays still see what part of the Thing was modified? We have a lot of different components on the edge that in parallel update different features of a thing, so it was nice to see the deltas only. So you can better debug what the edge components are doing (if they update full features or modify only the relevant part of a feature.

So I would argue that also for that reason it would make sense to reduce the SSE events as much as possible to the modification that happened to the Thing. Don't know if #1712 can still be fixed with that condition.

Btw. I am evaluating different options for the filtering. Next to plain javascript, Lodash would provide some support (which is no longer used) and also JSONata looks interesting. Currently I tend to plain javascript. Any recommendation or opinion from your side?

thjaeckle:
> Can you nowadays still see what part of the Thing was modified? We have a lot of different components on the edge that in parallel update different features of a thing, so it was nice to see the deltas only. So you can better debug what the edge components are doing (if they update full features or modify only the relevant part of a feature.

> So I would argue that also for that reason it would make sense to reduce the SSE events as much as possible to the modification that happened to the Thing. Don't know if #1712 can still be fixed with that condition.

[...]

PR #1858, opened 2024-01-06
UI - Add filter for tables

thfries:
Hello please find this PR that implements #1818

- The filter on incoming messages and connection logs is based on filtering the json arrays by using JSONPath
- There are new simplified filters that are more user friendly and convert to a JSONPath
- If that works on the two tables, the same approach can be used to generate RQL queries from that
- There is a switch to change to the JSONPath query instead of the basic filters.
- Basic filters support
   - equal to a property
   - like a property
   - like search in full json  
- Introduced Jest for unit tests
- Upgraded libraries

Please let me know how that works for you. I would like to add it to the connections list and for the things search. On things search we would get rid of the autocomplete approach then.


thjaeckle:
Thanks a lot @thfries 
Will take a look soon - there is one merge conflict which needs resolving.
And I saw that new files were added, for those we would need "2024" license header date ..

I need to let the license header check also run for .ts files .. :D

thfries:
Thanks @thjaeckle, I will take care of your points

PR #1860, opened 2024-01-11
Provide same behavior for filtering in arrays via search's "filter" for specified "condition"

thjaeckle:
* applying the `condition` also on "any" elements of an array of JsonObjects
* providing new `JsonObject` APIs: `containsFlatteningArrays(key)` and `getValueFlatteningArrays`

As this also adds new public methods to `JsonObject` it cannot be a bugfix, but a feature intead :)

### DIFF
--- a/things/model/src/main/java/org/eclipse/ditto/things/model/ImmutableFeatureProperties.java
+++ b/things/model/src/main/java/org/eclipse/ditto/things/model/ImmutableFeatureProperties.java
@@ -26,6 +26,7 @@ import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
+import org.eclipse.ditto.base.model.json.JsonSchemaVersion;
 import org.eclipse.ditto.json.JsonArray;
 import org.eclipse.ditto.json.JsonCollectors;
 import org.eclipse.ditto.json.JsonFactory;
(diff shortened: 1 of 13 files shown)


thjaeckle:

things/model/src/main/java/org/eclipse/ditto/things/model/ImmutableFeatureProperties.java
```diff
@@ -212,11 +212,21 @@ public boolean contains(final CharSequence key) {
         return wrapped.contains(key);
     }
 
+    @Override
+    public boolean containsFlatteningArrays(final CharSequence key) {
+        return wrapped.contains(key);
```
alstanchev, line 217: Shouldn’t this call "containsFlatteningArrays" on the wrapped object?

alstanchev (commented):
Other than the one comment in things/model/src/main/java/org/eclipse/ditto/things/model/ImmutableFeatureProperties.java
LGTM 👍

things/model/src/main/java/org/eclipse/ditto/things/model/ImmutableFeatureProperties.java
thjaeckle: @alstanchev very good catch - stupid copy&paste error :/
Will fix, thanks a lot

PR #1862, opened 2024-01-14
UI - Add filter for tables

thfries:
Hello @thjaeckle,
I again totally messed up the other PR, so please find this new PR for #1818.
Merge conflict should be resolved. Headers to new files added.

As mentioned, if this works well, I would like to extend that to RQL and use it also for Things search.
Any feedback also well come. I expect there can be improvements.

Details on commits:
- introduce jest for unit testing
- created new tableFilter web component
- upgraded dependencies
- fixed error in tsconfig
- introduced basic filter to create queries in UI
- dropdown entries now can attach data
- split adding dropdown entries for single and multiple entries
- moved utility methods to utils
- added table filter to connection logs
- added table filter to incoming thing messages
- connection metrix now shows red numbers for failures
- removed unused symbol in Things tab

Resolves: #1818

### DIFF
--- a/ui/__tests__/utils/basicFilter.test.ts
+++ b/ui/__tests__/utils/basicFilter.test.ts
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import { BasicFilters, Term, FilterType } from "../../modules/utils/basicFilters";
+
+describe("create JSONpath", () => {
+  test("One property equals one keyword", () => {
+    expect(new BasicFilters()
+        .addPropEq('quick', 'brown')
+        .createJsonPath())
+        .toBe('$[?(@.quick=="brown")]');
+  });
+
+  test("One property equals two keywords", () => {
+    expect(new BasicFilters()
+        .addPropEq('quick', 'brown')
+        .addPropEq('quick', 'fox')
+        .createJsonPath())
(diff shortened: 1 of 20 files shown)


thjaeckle:
Thanks a lot @thfries
Will have a look, looking forward to new features in the UI 😍

thfries:
Hi @thjaeckle,

thanks for your feedback. Highly appreciated.
I would like to consider some of the discussed improvements in this PR. Is there a tight deadline for 3.5?

- I'll check how you could filter merges to "/" path. I m thinking about searching in `_context/value`
- I was already struggling with the label for the advanced toggle, because of missing space. I was hoping that you can cover most cases with "basic" filters and you need the advanced mode only in rare exceptional cases...
- The input autocomplete behavior should be fixed in any case
- Yes, adding the star and storing filters was something I had in my mind, too. I would even also store the basic filters and provide option to give a name to a filter. (btw: one crazy idea was, to store the environment in a thing so you can share your filters and fields etc. with the rest of the team... 🤪)
- do you have an idea how to process RQL filtering on the browser side? my assumption was that any use of RQL involves the ditto server?

So I will investigate to those improvements...

thjaeckle:
Ditto 3.5 was planned to be released next week.
So too big changes to the PR are probably unrealistic.

So I would eg not try to cover the rql filtering.
RQL filtering on client side could be possible, but I don't think we should invest in that. JsonPath is mightier.
Maybe instead we can switch between server side and client side filtering at a later point.

thjaeckle:
@thfries If you don't mind, I would like to work on fixing the one point - so that we can get this PR ready to be merged for the Ditto 3.5 release:
> The filter input field suggests to autocomplete my stored credentials for the website

I already did that for the other search bar, so should not be a big thing..

thjaeckle (approved):
I added this little fix - preventing the browser from suggesting credentials for the input field. And also for the policyId.

From my point of view good to be merged, as we can tackle the other mentioned issues as follow-ups.

thfries:
Hi @thjaeckle, thank you for stepping in and considering the PR for 3.5! I did not find the time. Happy to have that in and thank you for the fix.

PR #1864, opened 2024-01-16
Enhance things search slot by displaying the amount of matching things

thjaeckle:
That shows directly, how many things were matched by the filer of by the "pinned" things.

Looks like this:
image

@thfries if you could please have a look at this small enhancement? :)

### DIFF
--- a/ui/modules/things/things.html
+++ b/ui/modules/things/things.html
@@ -24,6 +24,9 @@
                     <input autocomplete="false" name="hidden" type="text" style="display:none;">
                     <input type="search" class="form-control form-control-sm" id="searchFilterEdit" name="search"
                            spellcheck=false autocomplete="off" />
+                    <span class="input-group-text" id="searchFilterCount" data-bs-toggle="tooltip"
+                          title="Amount of things matching the filter">
+                    </span>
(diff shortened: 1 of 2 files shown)


thfries:
Hi @thjaeckle,

nice, thank you for this extension. Looks good to me and it works. (I assume with empty filter what I see is the total amount of things?)

One thought I had:
- I may have returned back the headline for the Things section and show the number of things there (like below). Then it would be consistent with the other collapsable sections and you would still see the number of Things if the things are collapsed. Also we would keep some more space for the filter input field. It would be overall a bit more consistent how to show "amount of items in a table". But if you see the count value as a property of the filter, it fits well like you did it. Do you expect to have the same on the new table filter? 
image

thjaeckle:
Yes, it shows the result of the count API which takes the same RQL as the search.
If no filter is provided, it shows the total count of things your user is allowed to see.

I would therefore (as it also directly is influenced by the filter) keep it close to the filter, that's why I put it there.

In the section heading might be misleading.
For the incoming thing Mesages I think it is good. There you see when some new messages arrive.

thfries:
Understood. I came to the same conclusion.
So I keep the new table filter as is. We may add that in future too.

Looking forward to have that merged 😀. Thank you

thfries:
@thjaeckle, I just realized that before there was no way to count things in the UI. Nice useful extension!

PR #1866, opened 2024-01-17
Provide option to provide negative numbers to historical event streaming

thjaeckle:
* in order to get e.g. "last 10" persisted events


thjaeckle:

Issue #1839, opened 2023-12-09
Stabilize Ditto on k8s pod rescheduling and rolling updates


thjaeckle:
We got feedback via our Gitter chat room that after pod rescheduling, Ditto cluster messaging sometimes gets broken:
https://matrix.to/#/!NgApgwRGamBVNEqzNU:gitter.im/$u8zsUQZHfuBIrIp-LdnIta8TajCGMoNVrEXSqxFdw14?via=gitter.im&via=matrix.org&via=othermo.ems.host

Ditto clusting uses Apache Pekko, formerly Akka.
I found this blogpost from the Lightbend team that since k8s 1.22 the default behavior to downscale pods changed:
https://www.lightbend.com/blog/faster-and-smoother-rolling-updates-for-akka-clusters-in-kubernetes

This is the related Akka Management  issue: https://github.com/akka/akka-management/issues/1130

Akka management fixed this issue in 1.3, which we however cannot use because of its license.

So there are a few options IMO to stablisize this in Ditto:
* disable on k8s cluster level the `LogarithmicScaleDown` flag
* change pod deletion cost to have the most cost for the cluster role leader for each Ditto service role
   * either as part of Ditto codebase (however Ditto is not aware of k8s)
   * or as contribution to Pekko management
   * or maybe as sidecar container / script?
   * or as custom k8s operator for Ditto

Just to provide some ideas..

A custom operator might also for other reasons be useful, eg in order to rolling update Ditto service by service instead of all at services at once.

@kalinkostashki maybe you also have an idea?

kalinkostashki:
Hey @thjaeckle,

From what I looked a bit there are several approaches:

1. Create a per service Helm charts and one wrapper chart like it was in eclipse packages:
    Benefits:

    -  you enable per service upgrades via hooks post-install, pre-upgrade, etc.
    -  this should fundamentally solve the issue you are talking about

    Negatives:

    - you create a more complex chart structure which is harder to maintain and will include code duplications I assume
    - Helm hooks and the way they behave may be a problem from the reading that I've done so it would be bad to avoid them -> for example what happens if a post install step fails for some reason...

2. Create per service Helm charts only
    Benefits:

    -  you enable per service upgrades via manually calling helm upgrade of a service
    -  this should fundamentally solve the issue you are talking about 

     Negatives:

    - you create a more complex chart structure which is harder to maintain and will **for sure** cause code duplications I assume

[...]

PR #1871, opened 2024-01-23
#1839: introduce script to Helm chart updating pod-deletion-cost

thjaeckle:
* script is executed as "pre-upgrade" hook and as cron-job
* script updates the pod-deletion-cost for the "oldest" members in the Ditto cluster

Resolves: #1839

Or it should at least improve stability during cluster upgrade


thjaeckle:
@alstanchev this would be ready to review - would appreciate a look :)

This basically implements what is described here https://www.lightbend.com/blog/faster-and-smoother-rolling-updates-for-akka-clusters-in-kubernetes - but purely via Helm and scripting.

kalinkostashki (approved):
LGTM :+1:

Issue #1869, opened 2024-01-22
Resolving imported policies in Ditto's "things-search" does not use a cache to load same policy


thjaeckle:
During updating a single policy which is imported in ~5.000 other Policies, I noticed many "ask timeout exceptions" being thrown in Ditto's "things-search" service.

It is intended that - when a imported policy is changed - all things using policies importing from that single policy are updated in the search index (to reflect the changed permissions).

It is however quite unexpected that e.g. the same policy `org.eclipse.ditto:my-root-policy` is loaded for all policies which use that exact same policy. But this seems to be the case:
https://github.com/eclipse-ditto/ditto/blob/aba53adf779a5dd6cad8c8548457097f7905969d/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/streaming/ResolvedPolicyCacheLoader.java#L49-L50

This will of course cause a huge amount of  `SudoRetrievePolicy` commands issued at the exact same time against the `policies` shard region, leading to many timeouts.
This place should make use of the Policy cache, so that for each cluster instance of "things-search" a heavily "imported" policy is only loaded once.

thjaeckle:
Resolved by: #1872

PR #1872, opened 2024-01-23
#1869 use cache in order to load imported policies when policies are …

thjaeckle:
…loaded after invalidation in search

This fixes an issue we encountered when using policy imports, when importing from a single policy a lot (thousands of policies importing from that one) and changing this policy, no cache was used and it was tried to load the imported policy thousands times. 

* cache is used in ResolvedPolicyCacheLoader
* it is now added that an invalidation caused by a policy contains the "causingPolicyTag" - which is then also invalidated
* however, the "causingPolicyTag" is only invalidated once (per search cluster node) to not overwhelm the cluster again due to too many invalidations


thjaeckle:
@alstanchev I would appreciate if you could review.
I paired with a colleague working on this, so we already had 4 eyes on it.

alstanchev (approved):
LGTM 👍

PR #1875, opened 2024-01-25
Fix helm chart open shift identation

mladBlum:
This PR fixes the indentation error in the helm chart for the securityContext. The problem just pops up if openshift deployment is enabled.
I am very confident, but a helm expert should take a quick look at it.


thjaeckle (approved):
LGTM, thx

PR #1881, opened 2024-01-29
fixed ResolvedPolicyCacheLoader loading several policy imports leading in PolicyEntryInvalidException

thjaeckle:
* cause was that `policy.withResolvedImports(` was used - but not needed in order to just find out referenced policy revisions


thjaeckle:
Invalid - Closing again ..

Issue #1882, opened 2024-01-29
When using Policy Imports, the Ditto search index might loose Data with Ditto 3.5.0


thjaeckle:
We just encountered, when updating to Ditto 3.5.0, that Ditto's search index might loose ("empty out") things with such a message:
```
WARN   Computed - due to missing enforcer - 'emptied out' ThingWriteModel for metadata <Metadata [thingId=...
```
That means that the thing is not "findable" afterwards with any Ditto search query, but only accessible directly via HTTP `GET /api/2/things/<thingId>`.

The reason seems to be extensive use of "Policy Imports" (the "Policy Label" is getting too long > 256 chars) - because of transitive lookup of policy imports:
```
org.eclipse.ditto.policies.model.PolicyEntryInvalidException: The Policy Label imported-io.......:a91d48bc-268c-4836-9ee2-059777c6ea50-imported-io........:a91d48bc-268c-4836-9ee2-059777c6ea50-main-imported-io........-imported-io........-LABEL is invalid
	at org.eclipse.ditto.policies.model.PolicyEntryInvalidException$Builder.doBuild(PolicyEntryInvalidException.java:127)
	at org.eclipse.ditto.policies.model.PolicyEntryInvalidException$Builder.doBuild(PolicyEntryInvalidException.java:113)
	at org.eclipse.ditto.base.model.exceptions.DittoRuntimeExceptionBuilder.build(DittoRuntimeExceptionBuilder.java:176)
	at org.eclipse.ditto.policies.model.ImmutableImportedLabel.of(ImmutableImportedLabel.java:65)
	at org.eclipse.ditto.policies.model.PoliciesModelFactory.newImportedLabel(PoliciesModelFactory.java:82)
	at org.eclipse.ditto.policies.model.PolicyImporter.lambda$rewriteImportedLabels$4(PolicyImporter.java:70)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.stream.Streams$StreamBuilderImpl.forEachRemaining(Streams.java:411)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
[...]

PR #1883, opened 2024-01-29
#1882 fixed ResolvedPolicyCacheLoader loading policy imports transitively

thjaeckle:
* cause was that by recursively using the cache, also policy imports were resolved transitively
* that could e.g. lead to a "PolicyLabelInvalidException" for scenarios with very nested policy import paths
* in either case, it was breaking the not available transitiveness of policy imports

Fixes: #1882


thjaeckle:
@alstanchev we found that today after our community call :/
Quite a blocker for running Ditto 3.5.0 - so a review would be highly welcome :D

alstanchev (approved):
LGTM 👍

PR #1886, opened 2024-01-30
Helm: fix hook scripts uses release name instead of fullname

WatcherWhale:
I found another small issue with the latest Helm Chart. The hook-scripts config map used the release name instead of the fullname.


thjaeckle (approved):
Good find @WatcherWhale - thanks a lot 👍 

I am currently building Ditto 3.5.1 - that's why the build failed, as I already updated the appVersion, but the docker containers are not yet deployed.

PR #1889, opened 2024-02-02
fix wrongly done invalidation of non-existing cached policy entry in search

thjaeckle:
* that could cause that e.g. a deleted imported policy which is re-added did not cause a reindexing of affected things


alstanchev (approved):
Awesome 👍

Issue #1888, opened 2024-02-02
Retrieving JSON arrays via `fields` selector is broken in Ditto 3.5.1 


AkshayKumarCB:
The GET /api/2/search/things API is currently broken on the release Ditto 3.5.1 when we attempt to fetch an array property.
A sample scenario is given below.

Sample data stored in Ditto:
```
{
  "thingId": "sample.eclipse.com:test_1",
  "policyId": "sample.eclipse.com:policy-default",
  "attributes": {
    "a": "b"
  },
  "features": {
    "values": {
      "properties": {
        "a": [
          1,
          2,
          3
        ]
      }
    }
  }
}
```

API query to fetch only the array 'a' from the Ditto thing:
`/api/2/search/things?namespaces=sample.eclipse.com&fields=features/values/properties/a`

Response on Ditto 3.4.4: (Returns array with values for "a")
`{"items":[{"features":{"values":{"properties":{"a":[1,2,3]}}}}],"nextPageOffset":-1}`

Response on Ditto 3.5.1: (Returns an empty array for "a")
`{"items":[{"features": { "values": { "properties": { "a": [] }}}}],"nextPageOffset": -1}`

However, if we were to update the fields query parameter in the request from `features/values/properties/a` to `features/values/properties`, we get the correct response - similar to what was observed in 3.4.4.

Looks like we have a bug for search/things API for arrays in 3.5.1 that needs to be fixed.

Reproducibility rate: 100% (5/5)

AkshayKumarCB:
Looks like the same behavior is also observed for **/api/2/things** API:

Request: /api/2/things/sample.eclipse.com:test_1?fields=**features/values/properties/a**
Response: `{"features":{"values":{"properties":{"a":[]}}}}`

Request: /api/2/things/sample.eclipse.com:test_1?fields=**features/values/properties**
Respnose: `{"features":{"values":{"properties":{"a":[1,2,3]}}}}`

thjaeckle:
Thanks a lot @AkshayKumarCB for the bugreport.
I will look into it - I remember having changed something around the field selection.

thjaeckle:
I assume this bug was introduced in PR #1860

PR #1890, opened 2024-02-02
#1888 fix retrieving simple arrays via field selectors

thjaeckle:
Fixes: #1888


thjaeckle:
@alstanchev if you could review, please - that would be great :)

alstanchev (approved):
LGTM 👍

PR #1891, opened 2024-02-04
UI small fixes and improvements

thfries:
Hi @thjaeckle,

please find some small updates on the UI:
    - send message with timeout 0 and no payload showed incorrect error in the response field. Reason was that on 202 `response.json()` was called that throwed an error. I added now 202 as "no content" case.
    - added category column on connection logs, so you can distinguish source and target log entries
    - connections js editors had wrong read-only behavior
    - connection log details ace editor now with word wrap, otherwise it is hard to read long message in one line. Btw.: is there a way to avoid that ditto transfers all HTTP headers into the ditto headers? The messages by default look polluted and with a lot of overhead in size.
    - Bug: filter on incoming messages does not work on new incoming messages

### DIFF
--- a/ui/modules/api.ts
+++ b/ui/modules/api.ts
@@ -370,7 +370,7 @@ export async function callDittoREST(method,
       throw new Error('An error occurred: ' + response.status);
     }
   }
-  if (response.status !== 204) {
+  if (response.status !== 204 && response.status !== 202) {
     if (returnHeaders) {
       return response;
     } else {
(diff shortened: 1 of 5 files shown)


thjaeckle (approved):
Thanks a lot @thfries 
LGTM 👍 

> Btw.: is there a way to avoid that ditto transfers all HTTP headers into the ditto headers? The messages by default look polluted and with a lot of overhead in size.

Not by default, the "header passthrough" is a feature - you can however configure which headers to ignore and not "pass through" via Ditto service configuration.

I think this was the place: https://github.com/eclipse-ditto/ditto/blob/ae8dc01bfa712209b9ee7d00a850ae9bb5561b3c/internal/utils/config/src/main/resources/ditto-protocol.conf#L3

PR #1901, opened 2024-02-09
fix that "Incoming Thing Updates" always repeated the first entry

thjaeckle:
* new entries to the table were not added correctly, but always only the first received entry was added again

@thfries I encountered that the "Incoming Thing Updates" table is broken - it only adds the first element of the `filteredMessage` array over and over again - and not the "latest one"

See e.g. screenshot: 
image

Could you review this small fix?


thjaeckle:
Discussed the issue with @thfries via chat - and approve on his behalf.

PR #1902, opened 2024-02-12
Added option to add custom annotations to podDeletionCostPatching jobs.

kyberpunk:
Hi. I would like to add capability for adding custom annotation to new pod deleting hook jobs in Ditto Helm chart. Main use case is to be able to adjust configuration (e.g. disabling or configuring of Istio sidecars) to make job working properly in special setups.
Thank you.

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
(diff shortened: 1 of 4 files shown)


thjaeckle (approved):
Hi @kyberpunk 

Looks good to me. 👍 
Could you provide a configuration example (just in this issue) how such an Istio sidecar configuration would look like?

kyberpunk:
> Hi @kyberpunk
> 
> Looks good to me. 👍 Could you provide a configuration example (just in this issue) how such an Istio sidecar configuration would look like?

Sure. We need to disable Istio sidecar for this particular job, because it prevents the job from finishing (know [issue](https://github.com/istio/istio/issues/11045) of Istio sidecars).

```
global:
  podDeletionCostPatching:
    annotations:
      sidecar.istio.io/inject: "false"
```

PR #1903, opened 2024-02-14
fix time:now placeholder truncation

thjaeckle:
* an exception was thrown regarding special regex char "[" used in "split"
* also fixed that fallback DittoJsonException for unknown exceptions could not be deserialized


thjaeckle:
Would want to release a 3.5.3 soon

alstanchev (approved):
LGTM 👍

Issue #1893, opened 2024-02-05
Ensure the consistency of signal enrichment for connections


kyberpunk:
Hello. I have a specific case where I use following connection `topic _/_/things/twin/events?extraFields=features/{{feature:id}},thingId&filter=like(resource:path,'/features/SoftwareUpdate*')` to consume change notifications by stateless service, which are enriched for other contextual data from the same feature even in case of partial feature modification. Currently it happens in case of concurrent modify commands, that `extra` field contains different values than actual changes in `value` field (see screenshot).

Since ditto supports retrieving historical revisions, the proposal is to provide deterministic and strong consistency by not looking up the current state, but the state at the revision number of the processed event. However, in some other use cases it may be more efficient to keep the caching implementation.

Thank you.

![ditto-mismatch-between-value-and-extra-status_ditto-twin-software-update-topic](https://github.com/eclipse-ditto/ditto/assets/12064920/3b511ef3-c172-4942-b55e-378fbefd26a8)

thjaeckle:
I try to take a look soon.

PR #1904, opened 2024-02-16
#1893 fix ensuring the consistency when doing signal enrichment

thjaeckle:
Resolves: #1893


kyberpunk:
Hi @thjaeckle you are very fast, looks perfect. Just one question: how exactly we can enforce the new behavior? Should we somehow setup ditto headers? Thank you.

thjaeckle:
@kyberpunk I would make this the default behavior - we want to ensure consistency here
There is also an optimisation in place, which does not apply the rather cost-expensive "history lookup" - when there is no newer modification of the thing, the "history lookup" is skipped and the in-memory representation is used for retrieving "extraFields".

That should for 95% of cases optimise and only for high-frequency cases add some additional cost.

thjaeckle:
Would want to release a 3.5.3 soon

Could you please run system-tests for this PR as part of a review? That would be super

kalinkostashki:
> @alstanchev @kalinkostashki could you please have a look? Would want to release a 3.5.3 soon
> 
> Could you please run system-tests for this PR as part of a review? That would be super

hey @thjaeckle,
sure will do :)

kalinkostashki (approved):
LGTM :+1:

PR #1921, opened 2024-03-29
PolicyAnnouncementTopicExtractor pub/sub namespaced topics support

### LINKED ISSUE #1920: Policy anounsments not working when connection is using namespace filtering
When a connection is subscribed to _/_/policies/announcements?namespaces=<some-namespace> the connection does not receive events with latest ditto version. If the namespace filter is removed events are received by the connection.
Strangely if the connection has two targets configured one with ns filtering and one without both publish the anouncement.
@thjaeckle i belive this is related to the latest pub-sub optimisations

alstanchev:
This adds namespaced subjects to the PolicyAnnouncementTopicExtractor

resolves #1920


thjaeckle (approved):
Thanks a lot, @alstanchev
Lgtm

PR #1922, opened 2024-04-04
Update README.md

Arijitdutta19910601:
Changing the directory before building Ditto with Maven

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ In order to first build Ditto and then start the built Docker images.
 
 #### 1. Build Ditto with Maven
 ```bash
+cd ~/ditto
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,6 +86,7 @@ In order to first build Ditto and then start the built Docker images.
 
 #### 1. Build Ditto with Maven
 ```bash
+cd ~/ditto
```
kalinkostashki, line 89: this would only work if you assume that you do the git clone in home(~) dir

thjaeckle:
As @kalinkostashki mentioned, this not only works for very few cases, it should be obvious to build from the checked out repo.

PR #1923, opened 2024-04-05
fix nested lookup of tm:refs in WoT model extension resolving

thjaeckle:


thjaeckle:
@alstanchev or @kalinkostashki could you please check this bugfix?
Really a small thing to fix recursively looking up `tm:ref` references in WoT models.

PR #1924, opened 2024-04-05
fix weak eTag handling of If-Match and If-None-Match headers

thjaeckle:
* they were not supporting both variants, but "If-Match" header assumed strong etags and "If-None-Match" assumed weak etags


thjaeckle:
@alstanchev or @kalinkostashki could you please check this bugfix?
Really a small thing to fix `If-Match` and `If-None-Match` header comparison using the provided "eTag".

kalinkostashki (approved):
LGTM :+1:

Issue #1931, opened 2024-04-29
Using some placeholders in RQL "filters" for connection "targets" is not possible


thjaeckle:
I would have expected that something like this filter for messages should work:
```
_/_/things/live/messages?filter=and(exists(feature:id),like(topic:subject,'fetch*'))
```

In words:
* Select all messages sent to a feature (containing a `feature:id` which can be resolved to a non-empty value=
* And having a subject beginning with "fetch"

But that fails with an exception:
```json
{
  "status":400,
  "error":"rql.expression.invalid",
  "message":"Unknown property name: feature:id",
  "description":"Ensure that all opening parentheses have also closing ones and that the RQL expression is valid."
}
```

I think that likely all placeholders which are placed instead of a "property name" might not be resolved as expected, so this would probably also fail:
```
eq(feature:id,'importantFeature')
```

Placeholders not supported here currently: 
* `feature:id`
* `thing:id`
* `entity:id`

PR #1932, opened 2024-04-29
#1931: fix that not all placeholders were supported in connection target filtering

thjaeckle:
Added support for placeholders in Connection "targets":
* `feature:id`
* `entity:id`
* `entity:namespace`
* `entity:name`
* `thing:id`
* `thing:namespace`
* `thing:name`

Resolves: #1931


thjaeckle:
@alstanchev @kalinkostashki Could you please have a look?
Some "inconsistency" I noticed when trying to formulate e.g. that I only want to be notified about "feature messages" via:
```
_/_/things/live/messages?filter=exists(feature:id)
```

alstanchev (approved):
LGTM 👍

PR #1934, opened 2024-05-02
Make nginx timeout configurable in helm chart

### LINKED ISSUE #1928: nginx timeout is not configurable in helm chart
Our use case is that we want to extend timeout for live messages to about 2 minutes. We configured all required timeouts for gateway service to be 115 seconds (slightly lower than 2 minutes due to limitation discussed at [How to increase message timeout to 5 minutes?](https://github.com/eclipse-ditto/ditto/discussions/1723)).

We install Ditto using helm chart and it looks like maximum timeout that nginx provides in hardcoded to [70 seconds](https://github.com/eclipse-ditto/ditto/blob/f6e156d869a97532f158ca60e05a6243a1bdb59e/deployment/helm/ditto/templates/nginx-configmap.yaml#L33-L45):

```text
      # timeouts are configured slightly higher than ditto-eclipse-ditto-gateway read-timeout of 60 seconds
      proxy_connect_timeout 70; # seconds, default: 60
      proxy_send_timeout 70; # seconds, default: 60
      proxy_read_timeout 70; # seconds, default: 60
      # will try another upstream if an error or timeout occurred during the connection
      # or if the upstream returns 502 response
      proxy_next_upstream error timeout http_502;
      # will retry up to 3 times to find another upstream to connect to
      proxy_next_upstream_tries 3;
      # will try for max. 20s to find another upstream to connect to
      proxy_next_upstream_timeout 20;

      send_timeout 70; # seconds, default: 60
```

It would be great if it was configurable.

dimabarbul:
Resolves #1928

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.5.4  # chart version is effectively set by release-job
+version: 3.5.5-0  # chart version is effectively set by release-job
 appVersion: 3.5.4
 keywords:
   - iot-chart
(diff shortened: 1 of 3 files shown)


thjaeckle (approved):
LGTM, thanks @dimabarbul 👍

Issue #1946, opened 2024-05-27
Configuring "oauth2" based authentication for "devops" access does not allow to use a different OpenID connect provider


thjaeckle:
When configuring to secure access to [DevOps commands](https://eclipse.dev/ditto/installation-operating.html#devops-commands) and to connections using `oauth2` via Helm: 
https://github.com/eclipse-ditto/ditto/blob/523985928dba2b0b4f63baa0b83c11293a16e1ba/deployment/helm/ditto/values.yaml#L1533-L1546

It is currently not possible to choose in the `openidConnectIssuers` a different OpenID connect provider than already defined in the "normal" oauth2 configuration at:
https://github.com/eclipse-ditto/ditto/blob/523985928dba2b0b4f63baa0b83c11293a16e1ba/deployment/helm/ditto/values.yaml#L1515-L1525

E.g. with a `-ops` suffix as in the comments .. Or even a completely different "issuer" endpoint.

The reason seems to be that this is loaded as "Extension" in Ditto (so basically a Singleton) and that only the first configuration is applied (which is the "normal" oauth2 config).  

So e.g. having admin users in a separated OpenID connect provider is not possible.

I also figured that this is currently not at all documented at [DevOps commands](https://eclipse.dev/ditto/installation-operating.html#devops-commands) - so adding some documentation about the option to use oauth for securing admin access would also be good as part of the bugfix.

PR #1948, opened 2024-05-29
#1946 fix that alternative OIDC provider for "devops" authorization could not be configured

thjaeckle:
* improved logging correlationId also for devops auth in gateway as well

Resolves: #1946


thjaeckle:
@alstanchev @kalinkostashki if you could have a look, it would be great 👍 

This issue happens when using `oauth2` for securing devops access and wanting to configure a different OIDC provider than already configured for the "main" authentication with oauth2.

kalinkostashki (approved):
LGTM :+1: 
and sorry for the delay :)

PR #1954, opened 2024-06-07
Add possibility to customize image used for pod-deletion-cost annotation patching job

kyberpunk:
Hi, I would like to add possibility for using custom images and private docker registries also for pod-deletion-cost annotation patching jobs. It works similar way as for Ditto deployments. Image is still the same by default so it is backwards compatible.
Thank you.


thjaeckle (approved):
@kyberpunk that's a great addition to the chart, thanks a lot 👍 

LGTM

PR #1956, opened 2024-06-10
Fix performance regression issue when running against MongoDB 6

thjaeckle:
* providing the Mongo query planner an index hint for aggregation queries

I prepared basically all aggregation queries done in `MongoReadJournal` to provide an optional `hint` which index to choose.

We faced the issue that with MongoDB 5 the aggregation done in `listNewestActiveSnapshotsByBatch` was very quick, as the MongoDB query planner chose to use the default index `"_id_"`.  
However, after updating to MongoDB 6, the query planner chose to use another index: `things_snaps_index` (`{ pid: 1, sn: -1, ts: -1 } `). This turns out to be really bad when having much snapshot data.  
In our case, the query time with MongoDB 5 was ~200ms and with MongoDB 6 it e.g. was ~100.000ms.

Configuring a hint which index name to use should provide the option to fine-tune as needed.

### DIFF
--- a/connectivity/service/src/main/resources/connectivity.conf
+++ b/connectivity/service/src/main/resources/connectivity.conf
@@ -4,6 +4,17 @@ ditto {
   mongodb {
     database = "connectivity"
     database = ${?MONGO_DB_DATABASE}
+
+    read-journal {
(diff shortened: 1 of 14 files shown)


alstanchev (approved):
Haven't run system tests on that but LGTM 👍

PR #1961, opened 2024-06-11
enable configuring even more flexibly indexes to `hint` for in Cleanup aggregations

thjaeckle:
* also create indexes speeding up the aggregation on snapshot collection


thjaeckle:
@kalinkostashki @alstanchev follow-up of bugfix #1956 .. it is not sufficient to provide a single `hint` - but conditionally on the `$match` done in the aggregation.
This fix also adds 2 indexes which should improve the aggregation queries.

By default I now left all index `hint`s as `null` - to assume the MongoDB Query Planner does its job well.

alstanchev (approved):
LGTM:)

Issue #1959, opened 2024-06-11
Generated Thing Description base URI should contain `/` at the end


egekorkan:
When I obtain a TD from a Ditto instance or look at the test TDs at https://github.com/w3c/wot-testing/blob/main/data/input_2022/TD/Ditto/TDs/ditto_floor-lamp-1.td.jsonld , the `base` has a URI that does NOT end with a `/`. When a standard URI parser is used and the base is combined with the relative URI in the forms, the absolute URI is wrong. This causes interoperability issues with other WoT tooling, such as Eclipse Thingweb node-wot (https://github.com/eclipse-thingweb/node-wot).

The easiest way to observe this is to open the browser console and use the standard `URL()` method (see [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)). The following function calls demonstrate it:

- `console.log(new URL("/attributes/manufacturer{?channel,timeout}","http://127.0.0.1/api/2/things/io.eclipseprojects.ditto:floor-lamp-0815"))` 
- Log: `http://127.0.0.1/attributes/manufacturer%7B?channel,timeout}`

If I am not wrong, the relative URLs should not start with `/` either since that will go to root of the URI.

thjaeckle:
Hi @egekorkan 
Nice to see you here :) 

Totally makes sense what you describe - I was "simply" appending the `href` to `base` in the implementation - but then starting with a `/` will of course select the "root" of the HTTP endpoint.

I will provide a fix for that - thanks a lot for the report.

egekorkan:
That was fast, thank you very much!

egekorkan:
Also @relu91, @danielpeintner and @jkrhb some nice interop between eclipse projects!

PR #1962, opened 2024-06-11
#1959: fix `base` and `href` of generated WoT ThingDescriptions

thjaeckle:
Resolves: #1959


thjaeckle:
@alstanchev @kalinkostashki a short review would be great :)

alstanchev (approved):
LGTM:)

PR #1963, opened 2024-06-12
make logstash buffer sizes configurable

kalinkostashki:
- introduced LOGSTASH_WRITE_BUFFER_SIZE and LOGSTASH_RING_BUFFER_SIZE envs in logback.xml of the respective services
 - added option to configure from helm chart


thjaeckle (approved):
LGTM 👍

PR #1964, opened 2024-06-13
fix performance regression with MongoDB 6 regarding snapshot aggregations

thjaeckle:
* next round: a total of 3 additional indexes need to be in place to have MongoDB 6 aggregation queries perform the same as with MongoDB 5
* documented in "MongoDB tuning" section in documentation
* removed/replaced index which is not needed
* fixed Helm configuration which caused empty config values


alstanchev (approved):
LGTM 👍

PR #1971, opened 2024-06-21
fix MongoReadJournalIT

kalinkostashki:
- recent changes in MongoReadJournalIT caused the setup method to fail with NPE
 - added mongodb.uri as a valid string to fix that


thjaeckle (approved):
LGTM

Issue #1525, opened 2022-10-28
Provide "aggregation" functions in Ditto search


thjaeckle:
Utilising the MongoDB search index, Ditto could support aggregation queries via its search API, e.g. in order to find out the mean temperature value of hundreds/thousands of selected things.

Our used [RQL syntax](https://github.com/persvr/rql) also provides "aggregations":
> aggregate(<property|function>,...) - Aggregates the array, grouping by objects that are distinct for the provided properties, and then reduces the remaining other property values using the provided functions

..
> sum(<property?>) - Finds the sum of every value in the array or if the property argument is provided, returns the sum of the value of property for every object in the array
mean(<property?>) - Finds the mean of every value in the array or if the property argument is provided, returns the mean of the value of property for every object in the array
max(<property?>) - Finds the maximum of every value in the array or if the property argument is provided, returns the maximum of the value of property for every object in the array
min(<property?>) - Finds the minimum of every value in the array or if the property argument is provided, returns the minimum of the value of property for every object in the array
recurse(<property?>) - Recursively searches, looking in children of the object as objects in arrays in the given property value

If any adopter of Ditto already had a need for such kind of aggregations, please provide feedback and your use-cases for such a feature.

thfries:
Hi @thjaeckle,
I recently thought about this feature. We have a clear need to have aggregations of important thing states to manage a large fleet of devices. To have an overview of groups of devices we need to aggregate the connectivity or health or update state of the devices within a group efficiently. Current solution is to replicate the important states in a relational db and update the replica based on a Ditto Kafka connection. Aggregation is then done efficiently in the RDB (home turf for that). But not good to having need for two databases.

So if there is an easy way to use aggregation capabilities of the MongoDB this would be beneficial.

- aggregations should be possible based on the values of an attribute (e.g. "groupId")
- policies of the things should be considered for aggregation (only aggregate value that the subject is allowed to see)

thjaeckle:
@thfries very interesting, thx for the input 

> aggregations should be possible based on the values of an attribute (e.g. "groupId")

You mean that eg the "groupId" would be used as "group by" in an aggregation?
But having another specified field? 

From your comment I assume that you would need a "district count" aggregation?
I can currently not imagine how to declare such a aggregation with the RQL syntax. 



For other queries and interpreting the sparse RQL documentation (and the lack of a specified count aggregation), I would assume that e.g. a "count" aggregation could look like:
```
[...]

PR #1973, opened 2024-06-24
Aggregation based custom metrics

### LINKED ISSUE #1972: Extended search based operator metrics
It would be nice to be able to define search metrics based on the search API.

It should be possible to tag devices with predefined values in the config as well as with resolved such from the things themselves.
All devices that match a filter and have the same tags should be aggregated under one metric instrument.
A cleanup functionality may be needed as there might be left out instruments for things that are deleted.
The configuration would as an extension to the already existing operator-metrics functionality.

```
operator-metrics {
      enabled = true
      scrape-interval = 30m
      custom-metrics {
        ...
      }
      custom-search-metrics {
        online_status {
          enabled = true
          scrape-interval = 20m # override scrape interval, run every 20 minute
          namespaces = [
            "org.eclipse.ditto"
          ]
          tags: {
            "online" = "{{online_placeholder}}"
            "location" = "{{attributes/Info/location}}"
          }
          filters = {
            online-filter = {
              filter = "gt(features/ConnectionStatus/properties/status/readyUntil/,time:now)"
              inline-placeholder-values = {
                // inline-placeholder-values are used to define hardcoded values to be used in the tags values if the placeholders are not json paths to an actual field in the thing
                // this is used to define different tags values based on the filter that matched the thing
                "online_placeholder" = true
              }
              // in order to do placeholder resolving from thing fields, the fields should be defined in the fields array
              // by default only the thingId is available for placeholder resolving
              fields = ["attributes/Info/location"]
              // The metric-value is used to define the value of the metric for the thing that matched the filter.
              // It does not support placeholders and should be a numeric value
            }
            offline-filter = {
              filter = "lt(features/ConnectionStatus/properties/status/readyUntil/,time:now)"
              inline-placeholder-values = {
                "online_placeholder" = false
              }
              fields = ["attributes/Info/location"]
            }
          }
        }
      }
    }
```

alstanchev:
Adds the posibility to define search based custom metrics.

Things can be tagged with predefined values in the config as well as with resolved ones from the things themselves.
All devices that match a filter and have the same tags ar aggregated under one metric instrument.
A cleanup is implemented so that no metric instruments are left after they are no longer used based on last time used

Resolves #1972


thjaeckle:
@alstanchev I will try to look at the PR soon, thanks :) 

In the meantime: the GitHub unit tests build fail because of these command registry tests.
There you probably added a new command + a new response + a new exception - and those must be added to the unit test.

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/OperatorSearchMetricsProviderActor.java
```diff
@@ -0,0 +1,398 @@
...
+                                .debug("Received QueryThingsResponse for custom search metric <{}>: {} - " +
+                                                "duration: <{}ms>",
+                                        metricName, queryThingsResponse.getSearchResult().getItems().getSize(),
+                                        Duration.ofNanos(System.nanoTime() - startTs).toMillis()
+                                );
+                        aggregateResponse(queryThingsResponse, filterConfig, metricName, config, dittoHeaders);
+                        if (queryThingsResponse.getSearchResult().hasNextPage() &&
+                                queryThingsResponse.getSearchResult().getCursor().isPresent()) {
+
+                            QueryThings nextPageSearch = QueryThings.of(searchThings.getFilter().orElse(null),
```
thjaeckle, line 212: variable can be `final`

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/OperatorSearchMetricsProviderActor.java
```diff
@@ -0,0 +1,398 @@
...
+                            } else {
+                                // update the timestamp
+                                log.debug("Updating custom search metric Gauge for metric <{}> with tags <{}>",
+                                        metricName, tags);
+                                return Tuple2.apply(tupleValue._1(), System.currentTimeMillis());
+                            }
+                        })
+                        ._1().set(value));
+
+        aggregatedByTagsValues.clear();
```
thjaeckle, line 331: I don't think this works reliably .. by using the class field `aggregatedByTagsValues`, this field is exposed to access by different threads.

The reason of this is because `recordMetric()` is called within a `Patterns.ask`, in the `whenComplete` - which does not run in the actor's main thread and therefore is not secured by Pekko against multi-threaded access.

So eventually, this will lead to race conditions and very unexpected behavior (e.g. when multiple metrics are gathered and e.g. a lot of results have to be paginated with a cursor over and over again).

I am afraid this has to be changed and re-thought ..

thjaeckle (changes requested):
Hi @alstanchev 

I did a first round of review.

I found some potential concurrency issues in the added `OperatorSearchMetricsProviderActor`.
Also, the added dependency to `ditto-edge-service` is not in line with keeping services only to the minimum dependencies they really need to know about.

Please have a look at my remarks, the concurrency thing is concerning me ..

thjaeckle:
The overall approach works of course, it however has some significant problems when the provided `filter` e.g. targets a huge dataset (e.g. millions of things - for which Ditto is built).  
Then, the approach will do a lot of "pagination" and put a lot of stress to Ditto, because all of the targeted things will have to be loaded into memory.

It is of course better than doing this "outside" of Ditto, using the HTTP API and doing the aggregation outside .. but it still is not ideal.
I would at least provide a disclaimer to the documentation to use this feature with care.

I would however suggested another way to do those "aggregations", which would be a lot more efficient and resource friendly:
We could use MongoDB aggregations in order to do those, as also suggested in #1525

So on the MongoDB `search` collection do an aggregation query (sticking to your provided example) of:
```
db.search.aggregate([
  {
    $group : {
      _id: "$t.attributes.Info.location",
      "sumOnline": {
        $sum: {
          $cond: [ { $gt: ["$t.features.ConnectionStatus.properties.status.readyUntil", "2024-07-03T08:10:00.000Z" ] }, 1, 0]
        }
      },
      "sumOffline": {
        $sum: {
          $cond: [ { $lt: ["$t.features.ConnectionStatus.properties.status.readyUntil", "2024-07-03T08:10:00.000Z" ] }, 1, 0]
        }
      }
    }
  }
])
```

Which would provide a DB result (without any need to page internally) like:
```json
[
  {
    "_id": "Berlin",
    "sumOnline": 6
  },
  {
    "_id": "Immenstaad",
    "sumOnline": 8
  }
]
```

So FMPOV this could even simplify the PR a lot .. e.g. the "operator" of Ditto could "just" configure the MongoDB aggregation query to perform in the configuration file.
Maybe with some "sugar" like using Ditto RQL to be able to e.g. use `time:now` placeholder.

What do you think, @alstanchev ?

alstanchev:
Awsome review, @thjaeckle, thanks. Sorry for the late reply i was out of office. I realy like the idea to move the aggregation to mongodb. Will look in to your suggestions and will do a refactor.

alstanchev:
@thjaeckle i have refactored the whole PR and the aggregation is now done in the db.
Please have a look when you can.

thjaeckle:
@alstanchev awesome, will do a review..

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultCustomSearchMetricConfig.java
```diff
@@ -0,0 +1,284 @@
...
+    private final Map<String, String> tags;
+    private final List<FilterConfig> filterConfigs;
+
+    private DefaultCustomSearchMetricConfig(final String key, final ConfigWithFallback configWithFallback) {
+        this.metricName = key;
+        enabled = configWithFallback.getBoolean(CustomSearchMetricConfigValue.ENABLED.getConfigPath());
+        scrapeInterval = configWithFallback.getDuration(CustomSearchMetricConfigValue.SCRAPE_INTERVAL.getConfigPath());
+        namespaces = Collections.unmodifiableList(new ArrayList<>(
+                configWithFallback.getStringList(CustomSearchMetricConfigValue.NAMESPACES.getConfigPath())));
+        groupBy = Collections.unmodifiableMap(new HashMap<>(
```
thjaeckle, line 54: shouldn't this be a `new LinkedHashMap` instead? I assume the "grouping by" is sensitive to the order?

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultCustomSearchMetricConfig.java
```diff
@@ -0,0 +1,284 @@
...
+                configWithFallback.getObject(CustomSearchMetricConfigValue.GROUP_BY.getConfigPath()).unwrapped()
+                        .entrySet()
+                        .stream()
+                        .collect(Collectors.toMap(Map.Entry::getKey, e -> String.valueOf(e.getValue())))));
```
thjaeckle, line 58: collect also to `LinkedHashMap`?

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/OperatorSearchMetricsProviderActor.java
```diff
@@ -0,0 +1,318 @@
...
+        final String metricName = gatherMetricsCommand.metricName();
+        final CustomSearchMetricConfig config = gatherMetricsCommand.config();
+        final DittoHeaders dittoHeaders = DittoHeaders.newBuilder()
+                .correlationId("gather-search-metrics_" + metricName + "_" + UUID.randomUUID())
+                .build();
+
+        final Map<String, String> namedFilters = config.getFilterConfigs().stream()
+                .collect(Collectors.toMap(CustomSearchMetricConfig.FilterConfig::getFilterName,
+                        CustomSearchMetricConfig.FilterConfig::getFilter));
+        AggregateThingsMetrics
```
thjaeckle, line 146: var can be `final`

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/OperatorSearchMetricsProviderActor.java
```diff
@@ -0,0 +1,318 @@
...
+                aggregateThingsMetrics = AggregateThingsMetrics.of(metricName, config.getGroupBy(), namedFilters,
+                Set.of(config.getNamespaces().toArray(new String[0])), dittoHeaders);
+        thingsAggregatorActorSingletonProxy.tell(aggregateThingsMetrics, getSelf());
+    }
+
+
+    private void handleAggregateThingsResponse(AggregateThingsMetricsResponse response) {
+        log.withCorrelationId(response).info("Received aggregate things response: {} thread: {}", //TODO debug
```
thjaeckle, line 154: what about this TODO?

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/OperatorSearchMetricsProviderActor.java
```diff
@@ -0,0 +1,318 @@
...
+                .collect(Collectors.toMap(CustomSearchMetricConfig.FilterConfig::getFilterName,
+                        CustomSearchMetricConfig.FilterConfig::getFilter));
+        AggregateThingsMetrics
+                aggregateThingsMetrics = AggregateThingsMetrics.of(metricName, config.getGroupBy(), namedFilters,
+                Set.of(config.getNamespaces().toArray(new String[0])), dittoHeaders);
+        thingsAggregatorActorSingletonProxy.tell(aggregateThingsMetrics, getSelf());
+    }
+
+
+    private void handleAggregateThingsResponse(AggregateThingsMetricsResponse response) {
```
thjaeckle, line 153: param should be `final`

thingsearch/model/src/main/java/org/eclipse/ditto/thingsearch/model/signals/commands/query/AggregateThingsMetricsResponse.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public Map<String, Long> getResult() {
+        return result;
+    }
+
+    public String getMetricName() {
+        return metricName;
+    }
+
+    private Supplier<RuntimeException> getJsonMissingFieldExceptionSupplier(String field) {
```
thjaeckle, line 90: param should be `final`

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationPredicateVisitor.java
```diff
@@ -0,0 +1,205 @@
...
+import org.bson.BsonValue;
+import org.bson.Document;
+import org.bson.conversions.Bson;
+import org.eclipse.ditto.placeholders.PlaceholderResolver;
+import org.eclipse.ditto.rql.model.ParsedPlaceholder;
+import org.eclipse.ditto.rql.query.criteria.Predicate;
+import org.eclipse.ditto.rql.query.criteria.visitors.PredicateVisitor;
+
+/**
+ * Creates Bson of a predicate.
```
thjaeckle, line 38: Copy&Paste of Javadoc .. seems wrong to me - "aggregation" is missing from it.

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationPredicateVisitor.java
```diff
@@ -0,0 +1,205 @@
...
+        return instance;
+    }
+
+    /**
+     * Creates a new instance of {@code CreateBsonPredicateVisitor} with additional custom placeholder resolvers.
+     *
+     * @param additionalPlaceholderResolvers the additional {@code PlaceholderResolver} to use for resolving
+     * placeholders in RQL predicates.
+     * @return the created instance.
+     * @since 2.3.0
```
thjaeckle, line 72: This `@since` is wrong

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationPredicateVisitor.java
```diff
@@ -0,0 +1,205 @@
...
+        return createInstance(Arrays.asList(additionalPlaceholderResolvers));
+    }
+
+    /**
+     * Creates a new instance of {@code CreateBsonPredicateVisitor} with additional custom placeholder resolvers.
+     *
+     * @param additionalPlaceholderResolvers the additional {@code PlaceholderResolver} to use for resolving
+     * placeholders in RQL predicates.
+     * @return the created instance.
+     * @since 2.3.0
```
thjaeckle, line 85: wrong

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationPredicateVisitor.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    public static CreateBsonAggregationPredicateVisitor createInstance(
+            final PlaceholderResolver<?>... additionalPlaceholderResolvers) {
+        return createInstance(Arrays.asList(additionalPlaceholderResolvers));
+    }
+
+    /**
+     * Creates a new instance of {@code CreateBsonPredicateVisitor} with additional custom placeholder resolvers.
```
thjaeckle, line 80:
```suggestion
     * Creates a new instance of {@code CreateBsonAggregationPredicateVisitor} with additional custom placeholder resolvers.
```

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationPredicateVisitor.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    public static CreateBsonAggregationPredicateVisitor getInstance() {
+        if (null == instance) {
+            instance = new CreateBsonAggregationPredicateVisitor(Collections.emptyList());
+        }
+        return instance;
+    }
+
+    /**
+     * Creates a new instance of {@code CreateBsonPredicateVisitor} with additional custom placeholder resolvers.
```
thjaeckle, line 67:
```suggestion
     * Creates a new instance of {@code CreateBsonAggregationPredicateVisitor} with additional custom placeholder resolvers.
```

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+import org.bson.conversions.Bson;
+import org.eclipse.ditto.placeholders.PlaceholderFactory;
+import org.eclipse.ditto.placeholders.TimePlaceholder;
+import org.eclipse.ditto.rql.query.criteria.Criteria;
+import org.eclipse.ditto.rql.query.criteria.Predicate;
+import org.eclipse.ditto.rql.query.expression.FilterFieldExpression;
+import org.eclipse.ditto.thingsearch.service.persistence.read.expression.visitors.GetFilterBsonVisitor;
+
+/**
+ * Creates the Bson object used for querying.
```
thjaeckle, line 29: Copy&paste of Javadoc is missing "aggregation" difference of this visistor

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/criteria/visitors/CreateBsonAggregationVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public Bson visitField(final FilterFieldExpression fieldExpression, final Predicate predicate) {
+        final Function<String, Bson> predicateCreator = predicate.accept(
+                CreateBsonAggregationPredicateVisitor.createInstance(
+                        PlaceholderFactory.newPlaceholderResolver(TIME_PLACEHOLDER, new Object())
+                )
+        );
+        return GetFilterBsonVisitor.apply(fieldExpression, predicateCreator, null);
```
thjaeckle, line 57: This could (instead of `null`) use the superclass' `authorizationSubjectIds` field - which you could make protected for that.
This would allow the class to be future-proof when e.g. used with specified `authorizationSubjectIds`.

thjaeckle (changes requested):
@alstanchev I did a first deeper review

This looks like a really great approach - I like how you reused the `Visitor` for just creating an "aggregation visitor" 👍 

I added remarks/requests inline.

In general, I think, that it would make sense to update configuration, code and documentation to not name this "custom search metrics" - but e.g. "custom aggregated metrics".
As the Ditto search is not used any longer for obtaining the metrics.

What do you think?

documentation/src/main/resources/pages/ditto/installation-operating.md
```diff
@@ -585,6 +585,101 @@ In Prometheus format this would look like:
...
+          group-by:{
+            "location" = "attributes/Info/location"
+            "isGateway" = "attributes/Info/gateway"
+          }
+          tags: {
+            "online" = "{{ inline:online_placeholder }}"
+            "health" = "{{ inline:health }}"
+            "hardcoded-tag" = "hardcoded_value"
+            "location" = "{{ group-by:location | fn:default('missing location') }}"
+            "altitude" = "{{ group-by:isGateway }}"
```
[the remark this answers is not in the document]
alstanchev: Oh, forgot to remove it. Used it to check boolean values.

thingsearch/model/src/main/java/org/eclipse/ditto/thingsearch/model/signals/commands/query/AggregateThingsMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+package org.eclipse.ditto.thingsearch.model.signals.commands.query;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import org.eclipse.ditto.base.model.headers.DittoHeaders;
+import org.eclipse.ditto.base.model.headers.WithDittoHeaders;
+
+public class AggregateThingsMetrics implements WithDittoHeaders {
```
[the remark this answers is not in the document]
alstanchev: It is not intentional. What would be the benefit of making it a command?

thjaeckle:
Thanks a lot for the adjustments @alstanchev 
I will try to do a last review round the next days :)

thjaeckle:
[1973-review.patch](https://github.com/user-attachments/files/16993666/1973-review.patch)

alstanchev:
Great thanks, will look into the WARN logs. There was a double call on stop but i tought i fixed it.

alstanchev:
@thjaeckle i have applied the changes you have requested and also minor other fix.

thjaeckle:
@alstanchev did you fix the WARN logs as well?

alstanchev:
should be, i am not getting it locally. The problem was that stop was called on each element of the stream. Now it is in a .watchTermination() of the stream.

thjaeckle (approved):
Cool, had another look - all fine for me now :)

Thanks again and a lot for this great addition 👍 

Merge whenever you are ready to :)

PR #1979, opened 2024-07-09
fix stopConsumingOnRequest test

kalinkostashki:
- change verify to account for 2 shutdown commands


thjaeckle (approved):
LGTM 👍 Thanks :)

PR #1986, opened 2024-07-22
add possibility to add extra volumes and volume mounts for extending …

vvasilevbosch:
…services in helm chart


kalinkostashki (approved):
LGTM :+1:

PR #1987, opened 2024-07-22
Integrate AWS IAM Role-based Authentication for MongoDB Atlas

hu-ahmed:
This PR introduces the integration of AWS IAM Role-based authentication for MongoDB Atlas in our application ([Passwordless authentication of MongoDB with AWS IAM](https://www.mongodb.com/docs/atlas/security/passwordless-authentication/)). By leveraging the AWS SDK's DefaultCredentialsProvider, we ensure secure and seamless authentication to MongoDB Atlas using the IAM role associated with the Kubernetes service account. The environment variables set by the service account are utilized to fetch the necessary credentials dynamically within the application. This implementation enhances security by eliminating the need for hardcoded credentials and leverages AWS's robust IAM policies for access control.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -59,6 +59,8 @@
         <!-- Keep these version consistent with pekko-persistence-mongodb.version's build.sbt -->
         <mongo-java-driver.version>5.1.1</mongo-java-driver.version>
 
+        <mongo-driver-sync.version>5.1.2</mongo-driver-sync.version>
+
         <jjwt.version>0.12.5</jjwt.version>
         <asm.version>9.2</asm.version>
         <qpid-jms-client.version>1.11.0</qpid-jms-client.version>
@@ -111,6 +113,7 @@
         <scalatest.version>3.2.17</scalatest.version>
         <docker-java.version>3.3.3</docker-java.version>
         <system-rules.version>1.19.0</system-rules.version>
+        <awssdk.version>2.26.21</awssdk.version>
     </properties>
 
     <dependencyManagement>
@@ -205,7 +208,11 @@
                 <artifactId>mongo-scala-driver_2.13</artifactId>
                 <version>${mongo-java-driver.version}</version>
             </dependency>
-
+            <dependency>
+                <groupId>org.mongodb</groupId>
+                <artifactId>mongodb-driver-sync</artifactId>
+                <version>${mongo-driver-sync.version}</version>
+            </dependency>
 
             <!-- Pekko Management -->
             <dependency>
@@ -801,6 +808,18 @@
                 <groupId>org.atteo.classindex</groupId>
                 <artifactId>classindex</artifactId>
                 <version>${classindex.version}</version>
+            </dependency>
+
+            <!-- ### aws sdk ### -->
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>sts</artifactId>
+                <version>${awssdk.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>auth</artifactId>
+                <version>${awssdk.version}</version>
             </dependency>
 
             <!-- ### Provided - own artifacts ### -->
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -59,6 +59,8 @@
         <!-- Keep these version consistent with pekko-persistence-mongodb.version's build.sbt -->
         <mongo-java-driver.version>5.1.1</mongo-java-driver.version>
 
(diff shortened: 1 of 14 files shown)


bom/pom.xml
```diff
@@ -111,6 +113,7 @@
         <scalatest.version>3.2.17</scalatest.version>
         <docker-java.version>3.3.3</docker-java.version>
         <system-rules.version>1.19.0</system-rules.version>
+        <awssdk.version>2.26.21</awssdk.version>
```
thjaeckle, line 115: Could you please move this declared version above the "test" dependency versions, maybe directly below the mongodb version - and adding a comment that this is needed for MongoDB AWS IAM authentication.

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -25,18 +25,12 @@
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.net.ssl.SSLContext;
 
+import com.mongodb.*;
```
thjaeckle, line 28: I would rather like to prevent wildcard imports, that makes it to "hidden" what actually is used from this package in the import section ..

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/config/MongoDbConfig.java
```diff
@@ -203,6 +210,11 @@ enum OptionsConfigValue implements KnownConfigValue {
              */
             RETRY_WRITES("retryWrites", true),
 
+            /**
+             * Determines whether IAM role should be used for authentication.
+             */
+            USE_AWS_IAM_ROLE("useAwsIamRole", false),
```
thjaeckle, line 216: Please also add this as configuration key (same as the above `retryWrites`) to this file, adding a one-line documentation in the config as well:
https://github.com/eclipse-ditto/ditto/blob/d5f43d0c7262154ae724b39d42ed9582b862f33b/internal/utils/config/src/main/resources/ditto-mongo.conf#L50-L52

This is then also the place to define the environment variable `USE_AWS_IAM_ROLE` which you already use in the Helm chart:
```hocon
useAwsIamRole = false
useAwsIamRole = ${?USE_AWS_IAM_ROLE}
```

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -476,6 +484,18 @@ private static SSLContext createAndInitSslContext() throws NoSuchAlgorithmExcept
             return result;
         }
 
-    }
+        private void applyAwsIamRoleSettings() {
+            AwsCredentialsProvider credentialsProvider = getAwsCredentialsProvider();
+            AwsCredentials credentials = credentialsProvider.resolveCredentials();
+            MongoCredential credential = MongoCredential.createAwsCredential(
```
thjaeckle, line 490: Please declare all 3 variables as `final` - that's the "Ditto codestyle" to use `final` wherever possible.

thjaeckle (changes requested):
Hi @hu-ahmed 
Thanks a lot for this contribution.
You should mention in the PR that this is about [Passwordless authentication of MongoDB with AWS IAM](https://www.mongodb.com/docs/atlas/security/passwordless-authentication/)

The build contains a test error:
```
Error:  Failures: 
Error:    DefaultOptionsConfigTest.testHashCodeAndEquals:63 EqualsVerifier found a problem in class org.eclipse.ditto.internal.utils.persistence.mongo.config.DefaultOptionsConfig.
-> Significant fields: equals does not use useAwsIamRole, or it is stateless.
```

And when updating the Helm chart, please update the Chart version - there is a linting in place which verifies that.

I added some inline comments, please check them.

internal/utils/config/src/main/resources/ditto-mongo.conf
```diff
@@ -50,6 +50,9 @@ ditto.mongodb {
     retryWrites = true
     retryWrites = ${?MONGO_DB_RETRY_WRITES}
 
+    useAwsIamRole = false
```
thjaeckle, line 53: please add a comment in the line before what this configuration is for .. (enabling MongoDB passwordless authentication via AWS IAM)

thjaeckle (approved):
Thanks a lot for the quick reaction on the review.

And thanks a lot for your first contribution to Ditto, I am happy to approve the PR 👍 🎉

PR #1994, opened 2024-08-08
Execute cronjob with non-root user

gbudinov:


alstanchev (approved):
@thjaeckle in some systems the security advisor warns on running containers with root user. 
This PR fixes that.

thjaeckle:
Shouldn't the user and group be configurable? AFAIR we also made the image configurable to run for the cron job, so a different user might be wanted.

alstanchev:
Good point. I will make it configurable.

alstanchev:
Closing this PR and will provide the changes in another one.

PR #1996, opened 2024-08-19
Add handling nested properties during resolving jwt placeholders

### LINKED ISSUE #1985: JWT placeholder resolving does not correctly expand nested arrays
Question #1984 brought up an issue with the `jwt` placeholder.

Given the following JWT:
```json
{
  "authorization": {
    "permissions": [
      {
        "scopes": [ "device.read" ],
        "resource_set_id": "123456"
      },
      {
        "scopes": [  "device.read", "device.write" ],
        "resource_set_id": "987654"
      }
    ]
  }
}
```

And given the following placeholder:
```
{{ jwt:authorization/permissions/scopes }}
```

It would be expected that this resolves to:
```
issuer:device.read
issuer:device.write
```

Due to the nature of "flattening" arrays.
This however does not work, it resolves to no subjects at all..

The JWT placeholder implementation currently seems to only "flatten" arrays at top level.

It should work "recursively", in arbitrary nested JsonObject / JsonArray constellations.

dimabarbul:
Resolves #1985

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authentication/jwt/JwtPlaceholder.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authentication/jwt/JwtPlaceholder.java
@@ -15,10 +15,10 @@ package org.eclipse.ditto.gateway.service.security.authentication.jwt;
 import static org.eclipse.ditto.base.model.common.ConditionChecker.argumentNotEmpty;
 import static org.eclipse.ditto.base.model.common.ConditionChecker.checkNotNull;
 
-import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
+import java.util.stream.Stream;
 
+import org.eclipse.ditto.json.JsonPointer;
 import org.eclipse.ditto.json.JsonValue;
 import org.eclipse.ditto.jwt.model.JsonWebToken;
 import org.eclipse.ditto.placeholders.Placeholder;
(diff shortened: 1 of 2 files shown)


thjaeckle:
@dimabarbul thanks a lot for the PR.
Will review next week once back from my vacation.

dimabarbul:
@thjaeckle Give me some time (couple of days or so), please, I want to make another version with HashSet instead of Stream for comparison. I haven't checked performance, but I'm pretty sure, HashSet version should be better.

dimabarbul:
HashSet version is:

```java
    private static Set<String> resolveValues(final JsonValue jsonValue, final JsonPointer jsonPointer) {
        final Set<String> result = new HashSet<>();

        resolveValuesInner(jsonValue, jsonPointer, result);

        return result;
    }

    private static void resolveValuesInner(
            final JsonValue jsonValue, final JsonPointer jsonPointer, final Set<String> result) {
        jsonPointer.getRoot()
                .ifPresentOrElse(
                        root -> {
                            if (jsonValue.isArray()) {
                                jsonValue.asArray().stream()
                                        .forEach(item -> resolveValuesInner(item, jsonPointer, result));
                            } else if (jsonValue.isObject()) {
                                jsonValue.asObject().getValue(root)
                                        .ifPresent(v -> resolveValuesInner(v, jsonPointer.nextLevel(), result));
                            }
                        },
                        () -> {
                            if (jsonValue.isArray()) {
                                jsonValue.asArray().stream().map(JsonValue::formatAsString).forEach(result::add);
                            } else {
                                result.add(jsonValue.formatAsString());
                            }
                        });
    }
```

I've roughly measured performance of HashSet and Stream versions (using `System.currentTimeMillis`) and found that time is relatively the same (with stream version slightly better). So I'll leave the PR as is.

thjaeckle (approved):
@dimabarbul LGTM 👍 

I just rebased from `master` branch and removed the "mutablity" unit test - as this is no longer possible to do with the updated Java version 21 (which was merged in the meantime on master).

Thanks a lot.

PR #2000, opened 2024-08-20
Execute cronjob with non-root user

alstanchev:
Executes cronjob with non-root user.
Also adresses the comments from #1994 and is now configurable from values.


thjaeckle (approved):

thjaeckle:
Thanks a lot, LGTM

PR #2001, opened 2024-08-20
delete duplicate nginx body style

neottil:
delete duplicate nginx body style in all deployments nginx configurations


neottil:
@thjaeckle how can I fix the lint-chart check? I see in the code that the 2"hart version is effectively set by release-job".

thjaeckle:
@neottil thanks for the PR - you just need to bump the `version` to next patch version - the linting enforces that the version is updated (and checks against `master` branch).

neottil:
@thjaeckle now it's ok.
Thanks

thjaeckle (approved):
LGTM, thanks for the PR 👍

PR #2004, opened 2024-08-26
adjust gid and uid in Docker images for "ditto" user and group

thjaeckle:
* base image suddenly uses gid and uid 1000


thjaeckle:
@alstanchev @kalinkostashki could you check this small PR?
The nightly Docker image build fails because the base image `docker.io/eclipse-temurin:17-jdk` added a group "ubuntu" with gid `1000` - and also a user "ubuntu" with uid `1000`.
This caused an error when trying to create "ditto" group with `gid` 1000.

That's why I increased the Ditto group and user id to `1005`.

alstanchev:
LGTM 👍

PR #1997, opened 2024-08-19
fix aws role credential

hu-ahmed:

PR #2006, opened 2024-08-27
provide AWS credential for MongoDB IAM authentication via supplier

thjaeckle:
* in order to allow refreshing the credentials
* also make AWS region configurable

Follow-up of and related to #1987 and #1997


hu-ahmed:
LGTM!

PR #2009, opened 2024-09-05
updating pekko to 1.1.0

thjaeckle:


thjaeckle:
As e.g. `pekko-http`, `pekko-management` are not yet released, I would not include this in 3.6.0, but move to 3.7.0 instead.

PR #2012, opened 2024-09-09
removed unused DITTO_POLICIES_ENFORCER_CACHE_MAX_SIZE in thingssearch

kalinkostashki:
- this property is only used for the policies service in CachingPolicyEnforcerProvider.java and PolicyEnforcerProviderExtension.java


deployment/helm/ditto/templates/thingssearch-deployment.yaml
```diff
@@ -254,8 +254,6 @@ spec:
               value: "{{ .Values.thingsSearch.config.updater.backgroundSync.throttle.throughput }}"
             - name: BACKGROUND_SYCN_THROTTLE_PERIOD
               value: "{{ .Values.thingsSearch.config.updater.backgroundSync.throttle.period }}"
-            - name: DITTO_POLICIES_ENFORCER_CACHE_MAX_SIZE
-              value: "{{ .Values.thingsSearch.config.updater.stream.policiesEnforcer.cache.maxSize }}"
             - name: DITTO_POLICIES_ENFORCER_CACHE_EXPIRE_AFTER_WRITE
```
thjaeckle, line 257: @kalinkostashki shouldn't those keys then also be deleted?

thjaeckle (changes requested):
Thanks @kalinkostashki for the PR - you're right - those are not used in things-search.

However, there are also more config keys to the `DITTO_POLICIES_ENFORCER_CACHE_*`- we should also remove them, right?

And what about the Helm `values.yaml`? 
`.Values.thingsSearch.config.updater.stream.policiesEnforcer.cache.` can likely be removed from that as well, right?

deployment/helm/ditto/templates/thingssearch-deployment.yaml
kalinkostashki: yes, you're right I forgot about them.
will remove them as well :)

kalinkostashki:
> Thanks @kalinkostashki for the PR - you're right - those are not used in things-search.
> 
> However, there are also more config keys to the `DITTO_POLICIES_ENFORCER_CACHE_*`- we should also remove them, right?
> 
> And what about the Helm `values.yaml`? `.Values.thingsSearch.config.updater.stream.policiesEnforcer.cache.` can likely be removed from that as well, right?

I don't we should remove Values.thingsSearch.config.updater.stream.policiesEnforcer.cache. as those are used to set THINGS_SEARCH_UPDATER_STREAM_POLICY_CACHE* options.

thjaeckle:
> I don't we should remove Values.thingsSearch.config.updater.stream.policiesEnforcer.cache. as those are used to set THINGS_SEARCH_UPDATER_STREAM_POLICY_CACHE* options.

Oh, good catch - then they were used by 2 places .. which is already a bad smell :D

Your changes LGTM 👍

thjaeckle (approved):

PR #2013, opened 2024-09-09
Helm: create empty dir for heapdumps and more

thjaeckle:
* empty dir for heapdumps survives pod restarts - configured JVM to do a heapdump on OOM by default
* adjust owner uid and gid tp 1005 for log folder permission change
* add more config options for cleanup


thjaeckle:
@kalinkostashki could you have a look at this PR?
We noticed that creating a heapdump causes k8s to kill the pod - and as a result, the heapdump is lost.

Mounting an empty directory solves that, as this survives pod recreation.

kalinkostashki (approved):
LGTM :+1:

PR #2014, opened 2024-09-09
monitoring connection pool config key fix

alstanchev:
Fixes the overiding of "connection-pool" property with env var.


kalinkostashki (approved):
LGTM :+1:

PR #2017, opened 2024-09-11
fix search background sync being stuck

thjaeckle:
* reason was a change in behavior of MongoDB reactive API which does no longer throw an exception for already created capped collections
* changed our logic to first lookup existing collections and only try creating the capped collection if it does not yet exist


alstanchev (approved):
LGTM 👍 
Luckily for me the latest version fixes the connection leek we faced.

Issue #2022, opened 2024-09-19
Support logging WARNing logs when WoT model enforcement failed, but still apply change


thjaeckle:
Make it configurable whether to, for an "invalid update" (not confirming to the WoT model):
* fail the "modify command" (so the current behavior, completely rejecting the update by Ditto)
* log a WARN / ERROR (in Ditto's "things" service), but still apply the "modify command"
* completely ignore validation errors / don't even validate in the first place

PR #2025, opened 2024-09-20
#2022 provide the option to log a warning instead of failing modifying API calls in WoT based validation

thjaeckle:
Resolves: #2022


thjaeckle:
@alstanchev could you have a look at this PR, please?

alstanchev (approved):
LGTM 👍

Issue #1951, opened 2024-06-03
Default value for `date-time` format in referenced WoT ThingModel is generated as empty string


thjaeckle:
Currently, when Ditto generates the JSON Skeleton for a Thing when a WoT "ThingModel" is referenced in the `"definition"` field on creation, non-optional properties with:
```json
{
  "type": "string",
  "format": "date-time"
}
```

Are generated as:
```json
"propertyName": ""
```
-> which is the "neutral element" for Strings.

We should however generate the "neutral element" for dates, which IMO would be EPOCH:
```json
"propertyName": "1970-01-01T00:00:00Z"
```

PR #2026, opened 2024-09-20
#1951 generate EPOCH timestamp as neutral element for string "format" date-time, date, time

thjaeckle:
Fixes: #1951


thjaeckle:

alstanchev (approved):
LGTM 👍

Issue #1591, opened 2023-03-04
UI: Authenticate with OAuth2 client ID / secret


thjaeckle:
Currently the UI accepts a Bearer token for authenticating.
More comfortable however would be if the UI would authenticate against a configured OAuth provider endpoint, using a client ID and optional (not all OAuth flows require a secret) secret.

The Swagger UI also supports such a mode where the [OAuth provider configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/) is added.
As the Ditto UI saves the environments locally, the OAuth secret does not need to be persisted or configured statically which should suffice security wise.

thfries:
Interesting. Agree that would be great.
I was experimenting a while ago and concluded that the OAuth flows only work with a backend. So I stopped that at that time. I need to take another look and I would need some support here.
You remember that secrets are not persisted by the UI?

thjaeckle:
Yes, I remember.
However new OAuth2 flows like pkse do not require a secret, so for those this would be a really good fit.
For others, a backend proxy may add the secret without the need to leak it into the client.

thfries:
Hi @thjaeckle, I'm coming back to this ticket. PKCE looks very promising but it needs some infrastructure (Authorization Server and OAuth2 Proxy for ditto?) and I think it will be easier to test if this is available in the cloud. I started with Google as an authorization server and tried out the Google authentication to the sandbox HTTP API.
- My Google Oauth ClientID and Secret did not work, because I do not know the redirect URL for the ditto sandbox server. Is there a more detailed description, how this use of sandbox HTTP API with Google should work? May be a good start would be to use the same approach for the UI as for the Swagger UI. But for that I would need to being able to test it
- Do I understand right, that there is an OAuth2 proxy on the Ditto sandbox installed? If needed, can we extend that with PKCE flow? It would be nice to have Google SSO directly with the sandbox and the UI 😉

thjaeckle:
Hi @thfries 

So the Ditto sandbox does not yet run OAuth2 proxy, but it should probably :D 
Currently, there is only an nginx in front of Ditto which does some "magic" in inserting OpenId connect client secret for Google OAuth.
I however did not check that for a long time, no idea if this still works.

I will look into it once I find the time - that however could take a while :/

PR #2032, opened 2024-09-25
#1591 add SSO via OIDC to Ditto UI

thjaeckle:
* make it configurable which OIDC issuers to use
* re-did a lot of the existing configuration of the Ditto UI
* added typing for Environment

Resolves: #1591

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authentication/jwt/DefaultJwtValidator.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authentication/jwt/DefaultJwtValidator.java
@@ -80,7 +80,7 @@ public final class DefaultJwtValidator implements JwtValidator {
     }
 
     private BinaryValidationResult validateWithJwtParser(final JsonWebToken jsonWebToken, final JwtParser jwtParser) {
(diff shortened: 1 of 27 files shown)


thfries:
Highly appreciated! Thank you for introducing that lot of typing improvements.

thjaeckle:
@thfries sure, my pleasure :)

Could you have a look on the PR as a reviewer?

thfries:
Sure, I already looked at the code changes, but I can check it out and take a look.
I fear that I will not be able to set up the OICD infrastructure. We don't have any authorization server in place. Would this work with Google authentication (but I still need a ditto instance that can be reached in the internet, I assume)? I will do my best...

thjaeckle:
@thfries I added a subfolder in the `ui` folder containing a `docker-compose.yml` in order to start a mock SSO (OAuth2) server.  
I also developed the SSO feature testing this .. 

With the updated `environmentTemplates.json` there is already an environment named `oidc_example` - choosing that should connect to the started OAuth2 mock server.

thjaeckle:
I did now my last commit for this PR .. reflecting the searchFilter's current search also via a query param `filter` - so that this "survives" the SSO redirect as well.

Issue #1915, opened 2024-03-14
Cannot retrieve entity at revision


simleo:
Using Ditto 3.5.3, started via docker compose. I created the floor lamp Thing as explained in https://eclipse.dev/ditto/3.5/intro-hello-world.html, then I updated it twice to set two of the three spots to "on". Now if I try "Streaming historical events via SSE" it seems to work (I get a full record for revision 1, and two more records that show the changes made at revision 2 and 3), but "Retrieve entity at specific revision" gives an error:

```console
curl -u ditto:ditto 'http://localhost:8080/api/2/things/org.eclipse.ditto:67b0d1a5-df3a-48a4-883e-7fb7432a5390' --header 'at-historical-revision: 1' | jq
{
  "status": 404,
  "error": "things:thing.history.notfound",
  "message": "The Thing with ID 'org.eclipse.ditto:67b0d1a5-df3a-48a4-883e-7fb7432a5390' at revision '1' could not be found or requester had insufficient permissions to access it.",
  "description": "Check if the ID of your requested Thing was correct, you have sufficient permissions and ensure that the asked for revision/timestamp does not exceed the history-retention-duration or is from the future."
}
```

thjaeckle:
@simleo thanks for the detailed report.
I'll try to reproduce.

simleo:
I tried to reproduce myself today and I couldn't. Must have been something temporary the other day. Sorry about that. However, I cannot make `at-historical-timestamp` work. I always get the latest version of the thing, no matter which timestamp I try to select. I get the same record even if I set a timestamp in the future or one that's before the first revision existed.

PR #2033, opened 2024-09-26
#1915 fix retrieving the historical value of a thing with `at-historical-timestamp`

thjaeckle:
There were several issues with the logic:

* the first issue was that as fallback when no `at-historical-revision` was provided, the "Latest" sequence number was selected -> which resulted in returning the latest state
   * that was fixed by not falling back at that place to this "fallback value", but work with a Optional instead
* the second problem was that for a thing which was freshly created (and as a result of that "freshness" no Snapshot was yet written into the snapshot store), Ditto did not distinguish correctly between "has not yet a snapshot" and "is not existing at the provided timestamp"

Both issues are fixed by this PR and using `at-historical-timestamp` should now work.

Fixes: #1915


thjaeckle:

alstanchev (approved):
LGTM 👍

Issue #2029, opened 2024-09-24
Configure workflow for running system tests


alstanchev:
It is goot to run the system tests from ditto-testing repo against PRs and master branch.
Running from PRs should be manual after a review from a maintainer that running is safe.

alstanchev:
@thjaeckle i am almost ready with that. Any input for the ways we want to be able to run the tests is welcome. Currently i am going in the direction of manual run with providing the pranch/PR as input.

thjaeckle:
@alstanchev awesome, that sounds great :)

> Currently i am going in the direction of manual run with providing the pranch/PR as input.

I think this is perfect as first step.

Looking much forward to this 👍

PR #2034, opened 2024-09-26
system tests workflow

alstanchev:
A system tests workflow. 
Resolves: #2029

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -1,0 +1,407 @@
+# Copyright (c) 2024 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


alstanchev:
@thjaeckle it is posible that it may need fine tuning as i only ran it localy and havent tested the result s and logs publishing.

thjaeckle:
@alstanchev did you see?
I tried running the system-tests GitHub action - and it failed .. but the output seemed strange to me, because in "Setup JDK" step it already needed the Ditto pom.xml?!

alstanchev:
Yes it is strange to me too, but comparing it to the maven workflow where the checkout is first I suspect there is something underneath that expects the code to be present (maybe because of the maven caching) I made a change but saw some other things that maybe will be different compared to local run with act. So I went the documentation rabbit hole. Will commit the change later

thjaeckle:
Please go ahead an directly commit to `master` branch if that helps/speeds you up ..

PR #2042, opened 2024-10-14
downgrade Rhino version to latest stable

thjaeckle:
* with new Rhino version there seem to be a change in behavior of Java-type to JS-type mapping


thjaeckle:
Turns out there was no problem with Rhino version .. closing

PR #2043, opened 2024-10-14
fixed JS mapping header serialization to string for array headers

thjaeckle:
This fixes an issue about JavaScript mapping in Ditto connections which use Json arrays as value inside `headers`.

Right now, this leads to e.g. the following follow-up exception (trying to parse `requested-acks` header as array):
```
The value 'org.mozilla.javascript.NativeArray@e90b588' of the header 'requested-acks' is not a valid JSON array. Verify that the value of the header 'requested-acks' is a valid 'JSON array' and try again.
```

So arrays were not correctly serialized, but their `toString()` on `NativeArray` (from Rhino library) was invoked, which just displays the default `toString()` implementation. 

This fix will transform the Rhino `NativeObject` and `NativeArray` to ditto-json `JsonObject` and `JsonArray` and will use the buit-in `formatAsString()` for header values.


thjaeckle:

alstanchev (approved):
LGTM 👍

PR #2045, opened 2024-10-16
fix WoT tm validation "dynamic" config Helm template

thjaeckle:
* some errors were undetected
* also improve parsing of validationContext, making the 3 parts of it optional in code


thjaeckle:
@alstanchev could you have a quick look?
In the new configuration wot WoT TM validation there were some issues mainly with the Helm config .. which is quite "complex" :D

alstanchev (approved):
LGTM 👍

PR #2047, opened 2024-10-21
Add new endpoint checkpermissions

### LINKED ISSUE #1137: Introduce policy decision API
For our UI, we'd like an endpoint that we can call to check if we should enable/disable certain elements based on policies.
As briefly discussed on gitter with @thjaeckle, that endpoint could look something like:

`POST /api/2/policies/{policyId}/actions/checkPermissions` with a payload like
```json
{
  "resource": "thing:/features/X/properties/Y", 
  "id": "org.eclipse.ditto:mythingId",
  "permissions": [ "READ" ],
 }
```
The response could be something like

```json
[
   {
      "permission": "READ",
      "allowed": true/false
   }
]
```

That would require you to have the policyId at hand, potentially having to fetch the Thing prior to using this api. It does make it generic to use for any kind of resource (thing, policy, messages)

hu-ahmed:
fixes #1137 

This PR introduces the implementation of the /checkPermissions endpoint in Ditto’s API, enhancing the authorization model by allowing permission checks on thing, message, and policy resources without modifying them. This endpoint validates whether specified entities have the required permissions for resources, returning authorization statuses in the response. This feature is particularly beneficial for UI-driven applications that rely on dynamic permission-based display logic, where interface elements can be enabled or disabled based on user access rights. Documentation includes usage examples, request structure, and response details.

Example Request
A POST request to /checkPermissions to verify permissions for multiple entities:

```
{
  "lamp_reader": {
    "resource": "thing:/features/lamp/properties/on",
    "entityId": "org.eclipse.ditto:some-thing-1",
    "hasPermissions": ["READ"]
  },
  "lamp_toggler": {
    "resource": "message:/features/lamp/inbox/messages/toggle",
    "entityId": "org.eclipse.ditto:some-thing-1",
    "hasPermissions": ["WRITE"]
  },
  "policy_admin": {
    "resource": "policy:/",
    "entityId": "org.eclipse.ditto:some-policy-1",
    "hasPermissions": ["READ", "WRITE"]
  }
}
```
Example Response

```
{
  "lamp_reader": true,
  "lamp_toggler": false,
  "policy_admin": true
}
```


gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissionsResponse.java
```diff
@@ -0,0 +1,167 @@
...
+     * The type of the response.
+     */
+    public static final String TYPE = TYPE_PREFIX + CheckPermissions.NAME;
+
+    private static final JsonFieldDefinition<JsonObject> JSON_PERMISSION_RESULTS =
+            JsonFieldDefinition.ofJsonObject("permissionResults");
+
+    private static final HttpStatus HTTP_STATUS = HttpStatus.OK;
+
+    private static final CommandResponseJsonDeserializer<CheckPermissionsResponse> JSON_DESERIALIZER =
```
thjaeckle, line 62: This is ununsed - "normally" such a `JSON_DESERIALIZER` is used in the `fromJson` method - maybe you wanted to do that?

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,84 @@
...
+import org.eclipse.ditto.json.JsonObject;
+import org.eclipse.ditto.json.JsonValue;
+import org.eclipse.ditto.json.JsonFactory;
+
+import javax.annotation.concurrent.Immutable;
+
+import java.util.List;
+
+@Immutable
+public final class ImmutablePermissionCheck {
```
thjaeckle, line 25: Please add javadoc for public classes.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    private final String entityId;
+    private final List<String> hasPermissions;
+
+    private ImmutablePermissionCheck(final String resource, final String entityId, final List<String> hasPermissions) {
+        this.resource = resource;
+        this.entityId = entityId;
+        this.hasPermissions = List.copyOf(hasPermissions);
+    }
+
+    public static ImmutablePermissionCheck of(final String resource, final String entityId,
```
thjaeckle, line 37: Javadoc for public static methods.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Creates a PermissionCheck model from a JsonObject.
+     *
+     * @param jsonObject the JsonObject containing permission check data.
+     * @return ImmutablePermissionCheck object.
+     */
+    public static ImmutablePermissionCheck fromJson(final JsonObject jsonObject) {
+        final String resource = jsonObject.getValueOrThrow(JsonFactory.newStringFieldDefinition("resource"));
```
thjaeckle, line 74: Please define (class private) constants for the several `JsonFactory.newStringFieldDefinition(...)`

gateway/service/src/main/resources/gateway.conf
```diff
@@ -63,7 +63,9 @@ ditto {
     # The factory for the gateway authentication directive
     gateway-authentication-directive-factory = org.eclipse.ditto.gateway.service.endpoints.directives.auth.DittoGatewayAuthenticationDirectiveFactory
     # Creator of props of HTTP request actors. Must implement HttpRequestActorPropsFactory.
-    http-request-actor-props-factory = org.eclipse.ditto.gateway.service.endpoints.actors.DefaultHttpRequestActorPropsFactory
+    http-request-actor-props-factory = org.eclipse.ditto.gateway.service.endpoints.actors.DefaultHttpRequestActorPropsFactory    # Creator of props of HTTP request actors. Must implement HttpRequestActorPropsFactory.
+    # Creator of props of CheckPermissions actors. Must implement CheckPermissionsActorPropsFactory.
+    check-permissions-actor-props-factory = org.eclipse.ditto.gateway.service.endpoints.actors.DefaultCheckPermissionsActorPropsFactory
```
thjaeckle, line 68: I think that can be removed - we don't use it as an extension as far as I could see.

policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutableResourcePermissions.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public List<String> getPermissions() {
+        return permissions;
+    }
+
+    @Override
+    public JsonObject toJson() {
+        JsonObjectBuilder jsonBuilder = JsonFactory.newObjectBuilder();
+        jsonBuilder.set("resourceType", resourceType);
```
thjaeckle, line 107: Instead of re-defining the strings again, re-use the (class private) defined constants also to-be-used in `fromJson` method.

policies/model/src/main/java/org/eclipse/ditto/policies/model/ResourcePermissionFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+
+public class ResourcePermissionFactory {
+
+    /**
+     * Creates a {@code ResourcePermission} from the provided JSON object.
+     *
+     * @param jsonObject the JSON object representing the resource permission.
+     * @return a {@code ResourcePermission}.
```
thjaeckle, line 25: Javadoc is missing `@since 3.7.0`

policies/model/src/main/java/org/eclipse/ditto/policies/model/ResourcePermissionFactory.java
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public static ResourcePermissions fromJson(final JsonObject jsonObject) {
+        return ImmutableResourcePermissions.fromJson(jsonObject);
+    }
+
+    public static ResourcePermissions newInstance(String resourceType, String resourcePath, List<String> permissions) {
```
thjaeckle, line 31: Javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsCommand.java
```diff
@@ -0,0 +1,163 @@
...
+import org.eclipse.ditto.json.JsonObjectBuilder;
+import org.eclipse.ditto.json.JsonPointer;
+import org.eclipse.ditto.policies.model.PolicyId;
+import org.eclipse.ditto.policies.model.ResourcePermissionFactory;
+import org.eclipse.ditto.policies.model.ResourcePermissions;
+import org.eclipse.ditto.policies.model.signals.commands.PolicyCommand;
+
+@Immutable
+@JsonParsableCommand(typePrefix = PolicyCommand.TYPE_PREFIX, name = PolicyCheckPermissionsCommand.NAME)
+public final class PolicyCheckPermissionsCommand extends AbstractCommand<PolicyCheckPermissionsCommand>
```
thjaeckle, line 44: I would rename it to `CheckPolicyPermissions` - the commands normally don't have `*Command` suffix.
And `Policy` as prefix also seems strange to me.

Maybe even `CheckPermissions` would be sufficient. (Then we would have to rename the `CheckPermissions` in gateway .. but this is only gateway internal anyways).

Javadoc is missing;
Please add unit tests

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsCommand.java
```diff
@@ -0,0 +1,163 @@
...
+            final Map<String, ResourcePermissions> permissionsMap,
+            final DittoHeaders dittoHeaders) {
+        return new PolicyCheckPermissionsCommand(policyId, permissionsMap, dittoHeaders);
+    }
+
+    public static PolicyCheckPermissionsCommand fromJson(final String jsonString, final DittoHeaders dittoHeaders) {
+        return fromJson(JsonFactory.newObject(jsonString), dittoHeaders);
+    }
+
+    public static PolicyCheckPermissionsCommand fromJson(final JsonObject jsonObject, final DittoHeaders dittoHeaders) {
```
thjaeckle, line 74: Javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsCommand.java
```diff
@@ -0,0 +1,163 @@
...
+        return new CommandJsonDeserializer<PolicyCheckPermissionsCommand>(TYPE, jsonObject).deserialize(() -> {
+            final String extractedPolicyId = jsonObject.getValueOrThrow(PolicyCommand.JsonFields.JSON_POLICY_ID);
+            final PolicyId policyId = PolicyId.of(extractedPolicyId);
+            final JsonObject permissionsJsonObject =
+                    jsonObject.getValueOrThrow(JsonFieldDefinition.ofJsonObject(PERMISSIONS_MAP));
+            final Map<String, ResourcePermissions> permissionsMap = permissionsJsonObject.getKeys()
+                    .stream()
+                    .collect(Collectors.toMap(
```
thjaeckle, line 82: Please collect to `LinkedHashMap` in order to preserve the order.

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Response to an {@link PolicyCheckPermissionsCommand}.
+ * <p>
+ * This class holds the results of the permission checks for each resource. The result is
+ * a {@link Map} with the resource identifier as the key and {@code true}/{@code false} as
+ * the value, indicating whether the requested permission was granted or denied.
+ */
+@Immutable
+@JsonParsableCommandResponse(type = PolicyCheckPermissionsResponse.TYPE)
+public final class PolicyCheckPermissionsResponse extends AbstractCommandResponse<PolicyCheckPermissionsResponse>
```
thjaeckle, line 49: Missing `@since 3.7.0` from Javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+        implements PolicyQueryCommandResponse<PolicyCheckPermissionsResponse> {
+
+    public static final String TYPE = "policyCheckPermissionsResponse";
```
thjaeckle, line 52: `TYPE` of a a commandResponse must always link to the Command like this:
```
 public static final String TYPE = TYPE_PREFIX + PolicyCheckPermissionsResponse.NAME;
```

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+     * @return a new {@code EnforcePolicyResponse}.
+     */
+    public static PolicyCheckPermissionsResponse of(final PolicyId policyId,
+            final Map<String, Boolean> permissionResults,
+            final DittoHeaders dittoHeaders) {
+
+        return new PolicyCheckPermissionsResponse(policyId, fromMap(permissionResults), HTTP_STATUS, dittoHeaders);
+    }
+
+    public static PolicyCheckPermissionsResponse fromJson(final JsonObject jsonObject,
```
thjaeckle, line 83: Javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), policyId, permissionsResults);
+    }
+
+    @Override
+    public String toString() {
+        return "PolicyCheckPermissionsResponse{" +
```
thjaeckle, line 124: please use `[` braces in toString format

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+    public JsonValue getEntity(final JsonSchemaVersion schemaVersion) {
+        return permissionsResults;
+    }
+
+    @Override
+    public JsonPointer getResourcePath() {
+        return JsonPointer.of("/permissionResults");
+    }
+
+    public static Map<String, Boolean> toMap(JsonObject jsonObject) {
```
thjaeckle, line 156: Javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+        final JsonObject permissionsResultsJsonObject =
+                jsonObject.getValueOrThrow(JsonFieldDefinition.ofJsonObject(PERMISSIONS_RESULTS));
+
+        return permissionsResultsJsonObject.getKeys()
+                .stream()
+                .collect(Collectors.toMap(
```
thjaeckle, line 162: Collect to `LinkedHashMap`

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
```diff
@@ -0,0 +1,179 @@
...
+                        key -> key.toString(),
+                        key -> permissionsResultsJsonObject.getValue(key).get().asBoolean()
+                ));
+    }
+
+
+    public static JsonObject fromMap(Map<String, Boolean> map) {
```
thjaeckle, line 169: Javadoc

policies/service/src/main/java/org/eclipse/ditto/policies/service/persistence/actors/strategies/commands/PolicyCheckPermissionsStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+import org.eclipse.ditto.policies.model.signals.commands.query.PolicyCheckPermissionsCommand;
+
+/**
+ * Strategy that handles the {@link org.eclipse.ditto.policies.model.signals.commands.query.PolicyCheckPermissionsCommand}.
+ * <p>
+ * It checks the provided resource permissions against the specified {@link PolicyId} and
+ * returns a {@link Map} where the key is the resource and the value is {@code true} if
+ * the permission was granted, otherwise {@code false}.
+ */
+public final class PolicyCheckPermissionsStrategy extends AbstractPolicyCommandStrategy<PolicyCheckPermissionsCommand, PolicyEvent<?>> {
```
thjaeckle, line 44: `@since 3.7.0` is missing from javadoc

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
thjaeckle, line 169: param should be `final`

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/query/PolicyCheckPermissionsResponse.java
thjaeckle, line 156: param should be `final`

thjaeckle (changes requested):
Thanks a lot @hu-ahmed for the PR.
Functionality already looks good, I had a first look and did some inline comments.

We should definitely add some unit tests as well for this feature (model classes + for the actor + for the HTTP route).
Javadoc has to be added for every public class/method, etc. .. so choose wisely what to make "public" and what not ;)

edge/service/src/main/java/org/eclipse/ditto/edge/service/dispatching/EdgeCommandForwarderActor.java
```diff
@@ -144,6 +145,7 @@ public Receive createReceive() {
                 )
                 .match(RetrieveThings.class, this::forwardToThingsAggregatorProxy)
                 .match(SudoRetrieveThings.class, this::forwardToThingsAggregatorProxy)
+                .match(SudoRetrieveThing.class, this::handleSudoRetrieveThing)
```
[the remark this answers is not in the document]
hu-ahmed: no it wouldn't and in this case I have to adapt ThingsAggregatorProxyActor to handle both

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+     *
+     * @param edgeCommandForwarder the actor responsible for forwarding commands to edge services.
+     * @param sender the actor reference to the original sender.
+     * @param defaultTimeout the default timeout for async operations.
+     */
+    private CheckPermissionsActor(final ActorRef edgeCommandForwarder, final ActorRef sender,
+            final Duration defaultTimeout) {
+        this.edgeCommandForwarder = edgeCommandForwarder;
+        this.sender = sender;
+        this.defaultAskTimeout = defaultTimeout;
```
thjaeckle, line 87: We should add a "receive timeout" to the actor:
```
getContext().setReceiveTimeout(defaultTimeout);
```

If the actor does not receive any message within that timeout, it will get a `ReceiveTimeout` message, which you can handle in `createReceive` below:
```
.match(ReceiveTimeout.class, receiveTimeout -> this.handleReceiveTimeout())
```

That way, we can eventually shutdown the actor, in case something goes wrong.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+ * and Things API to retrieve policy data and permissions.
+ * It responds to the sender with a {@link CheckPermissionsResponse} after evaluating the
+ * permission checks.
+ * <p>
+ * The actor forwards requests to the edge command forwarder to retrieve policies and permission
+ * results, handling various entity and policy resources.
+ *
+ * @since 3.7.0
+ */
+final class CheckPermissionsActor extends AbstractActor {
```
thjaeckle, line 64: The actor is not shutting down itself.

We must add that, as one actor is created for each request .. this is a potential memory and resource leak, if we don't shut it down again after handling the request.

I will point out the place where to do best..

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+
+        CompletableFuture<Void> allPolicyRetrievals = CompletableFuture.allOf(
+                groupedByEntityId.keySet().stream()
+                        .map(entityId -> retrieveOrSetPolicyId(entityId, groupedByEntityId.get(entityId), command))
+                        .toArray(CompletableFuture[]::new)
+        );
+
+        allPolicyRetrievals.thenRun(
+                        () -> handlePolicyRetrievalCompletion(command, permissionResults, groupedByEntityId))
+                .exceptionally(ex -> handleFailure(ex, command));
```
thjaeckle, line 130: After `.exeptionally`, you can add `.whenComplete(...` in order to shutdown this actor.
For that, you can define a method which you would call:
 ```
private void stopSelf() {
            final var context = getContext();
            context.cancelReceiveTimeout();
            context.stop(getSelf());
        }
```

File: gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java [hunk 24]
```diff
@@ -0,0 +1,417 @@
...
+     * grouping permission checks by entity, and performing policy retrievals.
+     *
+     * @param command the {@link CheckPermissions} message to handle.
+     */
+    private void handleCheckPermissions(final CheckPermissions command) {
+        final Map<String, Boolean> permissionResults = initializePermissionResults(command);
+        final Map<String, Map<String, PermissionCheckWrapper>> groupedByEntityId =
+                groupByEntityAndPolicyResource(command);
+
+        CompletableFuture<Void> allPolicyRetrievals = CompletableFuture.allOf(
```
thjaeckle, line 122: can be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+                .exceptionally(ex -> handleFailure(ex, command));
+    }
+
+    /**
+     * Initializes the permission results map with 'false' for each permission check.
+     *
+     * @param command the {@link CheckPermissions} command containing the permission checks.
+     * @return a map with the permission check results, initialized to {@code false}.
+     */
+    private Map<String, Boolean> initializePermissionResults(final CheckPermissions command) {
```
thjaeckle, line 139: `command` can be `final`

thjaeckle, line 139: I would prefer Java simple types, so instead of `Boolean`, please use `boolean` in the map.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+    /**
+     * Handles the completion of policy retrieval by checking the permissions for each policy.
+     *
+     * @param command the {@link CheckPermissions} command.
+     * @param permissionResults the map to store the permission check results.
+     * @param groupedByEntityId the permission checks grouped by entity ID.
+     */
+    private void handlePolicyRetrievalCompletion(final CheckPermissions command,
+            final Map<String, Boolean> permissionResults,
+            Map<String, Map<String, PermissionCheckWrapper>> groupedByEntityId) {
```
thjaeckle, line 199: `groupedByEntityId` can be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+        // Aggregate permissions by PolicyId and Resource
+        final Map<PolicyId, Map<String, ResourcePermissions>> aggregatedPermissions =
+                groupPermissionsByPolicyId(groupedByEntityId);
+
+        List<CompletionStage<Void>> permissionCheckFutures = aggregatedPermissions.entrySet().stream()
```
thjaeckle, line 204: can be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    /**
+     * Handles the completion of policy retrieval by checking the permissions for each policy.
+     *
+     * @param command the {@link CheckPermissions} command.
+     * @param permissionResults the map to store the permission check results.
+     * @param groupedByEntityId the permission checks grouped by entity ID.
+     */
+    private void handlePolicyRetrievalCompletion(final CheckPermissions command,
```
thjaeckle, line 197: This method should return a `CompletionStage<Void>`, which the caller would invoke with `thenCompose` - in order to "wait" for its completion before e.g. completing itself.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+            Map<String, Map<String, PermissionCheckWrapper>> groupedByEntityId) {
+        // Aggregate permissions by PolicyId and Resource
+        final Map<PolicyId, Map<String, ResourcePermissions>> aggregatedPermissions =
+                groupPermissionsByPolicyId(groupedByEntityId);
+
+        List<CompletionStage<Void>> permissionCheckFutures = aggregatedPermissions.entrySet().stream()
+                .map(aggregateEntry -> checkPermissionsForPolicy(aggregateEntry, command, permissionResults))
+                .toList();
+
+        CompletableFuture.allOf(permissionCheckFutures.toArray(new CompletableFuture[0])).thenRun(() -> {
```
thjaeckle, line 208: return that CompletionStage as response of the method.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+            final CheckPermissionsResponse response =
+                    CheckPermissionsResponse.of(permissionResults, command.getDittoHeaders());
+            sender.tell(response, getSelf());
+        }).exceptionally(ex -> handleFailure(ex, command));
```
thjaeckle, line 212: This `.exceptionally` is not needed when "weaving in" the returned completion stage with `thenCompose` in the calling CompletionStage.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+     * @param aggregateEntry the entry containing the policy ID and resource permissions.
+     * @param command the {@link CheckPermissions} command.
+     * @param permissionResults the map to store the permission check results.
+     * @return a {@link CompletionStage} representing the result of the permission check.
+     */
+    private CompletionStage<Void> checkPermissionsForPolicy(
+            final Map.Entry<PolicyId, Map<String, ResourcePermissions>> aggregateEntry,
+            final CheckPermissions command,
+            final Map<String, Boolean> permissionResults) {
+        CheckPolicyPermissions policyCommand = CheckPolicyPermissions.of(
```
thjaeckle, line 246: can be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+
+    /**
+     * Handles a failure during the permission checking process by logging the error and sending a
+     * {@link DittoInternalErrorException} back to the sender.
+     *
+     * @param ex the exception that occurred during the permission check.
+     * @param command the {@link CheckPermissions} command containing permission checks.
+     * @return null, as the error is propagated via an exception and response is sent.
+     */
+    private Void handleFailure(Throwable ex, CheckPermissions command) {
```
thjaeckle, line 300: params should be `final`, response type should be primitive `void`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+        logError(ex, command.getPermissionChecks(), command.getDittoHeaders());
+
+        final DittoInternalErrorException errorResponse = DittoInternalErrorException.newBuilder()
+                .dittoHeaders(command.getDittoHeaders())
+                .cause(ex)
+                .message("Internal error while processing permission checks for command: " + command)
```
thjaeckle, line 306: The `toString` of `CheckPermissions` command is too long and technical for an exception message - this would e.g. be shown at the HTTP API as response.
I would not provide a message at all - the standard message should be sufficient for an internal server error.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+                .build();
+        sender.tell(errorResponse, getSelf());
+        throw new CompletionException(ex);
```
thjaeckle, line 309: No need to wrap that into `CompletionException` yourself - IMO you already handled it, so no need to throw at all ..

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+     * This method logs the failure of a permission check operation using a {@code logger} with correlation ID
+     * extracted from the provided {@code DittoHeaders}. The logged message includes the details of the failed
+     * permission check request and the associated exception message.
+     * </p>
+     *
+     * @param ex the exception that occurred during the permission check.
+     * @param permissionChecks the map containing the permission checks that failed.
+     * @param headers the {@code DittoHeaders} associated with the command, used for logging context.
+     */
+    private void logError(Throwable ex, LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks,
```
thjaeckle, line 325: params should be `final`
Instead of `LinkedHashMap` as param, you should use `Map` - `LinkedHashMap` is just the implementation we choose.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+            DittoHeaders headers) {
+        logger.withCorrelationId(headers)
+                .warn("Permission check failed for request: {}. Error: {}", permissionChecks, ex.getMessage());
```
thjaeckle, line 328: logError method logs a warning .. should we adjust the method name?

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+    /**
+     * Logs an error related to a failed permission check.
+     *
+     * @param ex the exception that occurred.
+     * @param entityId the ID of the entity being checked.
+     * @param resource the resource being checked.
+     */
+    private void logError(Throwable ex, String entityId, String resource, final DittoHeaders dittoHeaders) {
+        logger.withCorrelationId(dittoHeaders)
+                .warn("Permission check failed for entityId: {} and resource: {}. Error: {}", entityId, resource,
```
thjaeckle, line 340: logError method logs a warning .. should we adjust the method name?

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    /**
+     * Logs an error related to a failed permission check.
+     *
+     * @param ex the exception that occurred.
+     * @param entityId the ID of the entity being checked.
+     * @param resource the resource being checked.
+     */
+    private void logError(Throwable ex, String entityId, String resource, final DittoHeaders dittoHeaders) {
```
thjaeckle, line 338: params should be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+    /**
+     * Creates a {@link ResourcePermissions} object from the {@link ImmutablePermissionCheck}.
+     * <p>
+     * This method constructs a {@link ResourcePermissions} object for a specific resource by
+     * extracting the resource type, resource path, and the list of permissions.
+     *
+     * @param check the {@link ImmutablePermissionCheck} containing the permission data.
+     * @return the constructed {@link ResourcePermissions}.
+     */
+    private ResourcePermissions createResourcePermissions(ImmutablePermissionCheck check) {
```
thjaeckle, line 353: param `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+        ResourceKey resourceKey = check.getResourceKey();
+        List<String> permissions = check.getHasPermissions();
```
thjaeckle, line 355: Both `final` - or directly create `ResourcePermissionFactory.newInstance(check.getResourceKey(), check.getHasPermissions())`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+     * wrapped inside a {@link DittoInternalErrorException} is thrown, ensuring proper error handling.
+     *
+     * @param entityId the ID of the entity for which the {@link PolicyId} is being retrieved.
+     * @param isPolicyResource flag indicating whether the entity is a policy resource.
+     * @param headers the {@link DittoHeaders} associated with the request.
+     * @return a {@link CompletionStage} that completes with the {@link PolicyId} once the operation is finished.
+     * @throws DittoRuntimeException if an unexpected response is received, or the {@link PolicyId} is missing in the response.
+     * @throws DittoInternalErrorException if an unexpected response type is encountered, or if the {@link PolicyId} is missing.
+     */
+    private CompletionStage<PolicyId> retrievePolicyIdForEntity(String entityId, boolean isPolicyResource,
```
thjaeckle, line 380: params should be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+                        } else if (result instanceof DittoRuntimeException) {
+                            throw (DittoRuntimeException) result;
+                        } else {
+                            throw DittoInternalErrorException.newBuilder()
+                                    .dittoHeaders(headers)
+                                    .cause(new IllegalArgumentException(
+                                            "Unexpected response type: " + result.getClass().getSimpleName()))
+                                    .build();
+                        }
+                    }).handle((policyId, throwable) -> {
```
thjaeckle, line 403: You can combine the above `thenApply` and this `handle` to just a `handle` - then you don't have to create a `DittoInternalErrorException` twice and simplify the code..

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissions.java
```diff
@@ -0,0 +1,145 @@
...
+    private final LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks;
+
+    /**
+     * Constructs a new {@code CheckPermissionsCommand} object.
+     *
+     * @param dittoHeaders the headers of the command.
+     * @param permissionChecks a linked hash map of permission checks to be performed.
+     */
+    private CheckPermissions(final DittoHeaders dittoHeaders,
+            final LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks) {
```
thjaeckle, line 67: Please don't use `LinkedHashMap` in signatures - only in the implementation.
Signatures should always use just `Map`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissions.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * The type of the command.
+     */
+    public static final String TYPE = TYPE_PREFIX + CheckPermissions.NAME;
+
+    private static final JsonFieldDefinition<JsonObject> PERMISSION_CHECKS_FIELD = JsonFactory.newJsonObjectFieldDefinition(
+            "permissionChecks"
+    );
+
+    private final LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks;
```
thjaeckle, line 58: this should just be a `Map`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissions.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Constructs a new {@code CheckPermissionsCommand} object.
+     *
+     * @param dittoHeaders the headers of the command.
+     * @param permissionChecks a linked hash map of permission checks to be performed.
+     */
+    private CheckPermissions(final DittoHeaders dittoHeaders,
+            final LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks) {
+        super(TYPE, Category.QUERY, dittoHeaders);
+        this.permissionChecks = new LinkedHashMap<>(permissionChecks);
```
thjaeckle, line 69:
```suggestion
        this.permissionChecks = Collections.unmodifiableMap(new LinkedHashMap<>(permissionChecks));
```

Need to make this immutable, otherwise the class itself is not immutable.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissions.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Creates a new {@code CheckPermissionsCommand} with the provided headers and permission checks.
+     *
+     * @param headers the headers of the command.
+     * @param permissionChecks the permission checks to be included in the command.
+     * @return a new {@code CheckPermissionsCommand}.
+     */
+    public static CheckPermissions of(final LinkedHashMap<String, ImmutablePermissionCheck> permissionChecks,
```
thjaeckle, line 79: just `Map`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/CheckPermissions.java
```diff
@@ -0,0 +1,145 @@
...
+                ));
+        return new CheckPermissions(dittoHeaders, permissionChecks);
+    }
+
+    /**
+     * Returns the permission checks contained in this command.
+     *
+     * @return a linked hash map of permission checks.
+     */
+    public LinkedHashMap<String, ImmutablePermissionCheck> getPermissionChecks() {
```
thjaeckle, line 107: return just `Map`, the implementation is still "ordered"

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,140 @@
...
+ * The class is immutable and provides methods to convert from/to JSON, ensuring safe use across threads.
+ *
+ * @since 3.7.0
+ */
+@Immutable
+public final class ImmutablePermissionCheck {
+
+    private final ResourceKey resourceKey;
+    private final String entityId;
+    private final Boolean isPolicyResource;
```
thjaeckle, line 43: could be a `boolean` primitive type

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,140 @@
...
+    public ResourceKey getResourceKey() {
+        return resourceKey;
+    }
+
+    /**
+     * Returns whether the resource is a policy resource.
+     *
+     * @return {@code true} if the resource is a policy, {@code false} otherwise.
+     */
+    public Boolean isPolicyResource() {
```
thjaeckle, line 88: could return a `boolean` primitive type

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -0,0 +1,140 @@
...
+ * This class encapsulates a check to verify permissions for a specific resource (e.g., Thing or Policy)
+ * associated with an entityId. It holds information about the resource, entity, and the permissions that
+ * are being checked.
+ * <p>
+ * The class is immutable and provides methods to convert from/to JSON, ensuring safe use across threads.
+ *
+ * @since 3.7.0
+ */
+@Immutable
+public final class ImmutablePermissionCheck {
```
thjaeckle, line 39: This class could also implement `Jsonifiable<JsonObject>` in order to provide `JsonObject toJson()`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/PermissionCheckWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Wrapper class for handling permission checks along with the associated {@link PolicyId}.
+ * <p>
+ * This class wraps an {@link ImmutablePermissionCheck} and provides an additional field to hold
+ * the {@link PolicyId} that is associated with the permission check. The {@link PermissionCheckWrapper}
+ * allows for flexible management of both the permission check and its related policy.
+ *
+ * @since 3.7.0
+ */
+public class PermissionCheckWrapper {
```
thjaeckle, line 27: class needs to be `final` in order to be immutable.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/PermissionCheckWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+    private final ImmutablePermissionCheck permissionCheck;
+    private PolicyId policyId;
```
thjaeckle, line 29: this fields needs to be `final` so that the class is immutable

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/PermissionCheckWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Constructor to initialize the wrapper with a given {@link ImmutablePermissionCheck}.
+     *
+     * @param permissionCheck the permission check to wrap.
+     */
+    public PermissionCheckWrapper(ImmutablePermissionCheck permissionCheck) {
```
thjaeckle, line 36: needs the `policyId` as additional parameter, to initialize all immutable fields
also, params should be `final`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/PermissionCheckWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+    public ImmutablePermissionCheck getPermissionCheck() {
+        return permissionCheck;
+    }
+
+    /**
+     * Sets the {@link PolicyId} associated with this permission check.
+     *
+     * @param policyId the policy ID to set.
+     */
+    public void setPolicyId(PolicyId policyId) {
```
thjaeckle, line 54: Please remove setter - we cannot have mutable classes in Pekko to avoid concurrency issues.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,417 @@
...
+     * <p>
+     * If the entity is a policy resource, it directly sets the {@link PolicyId} for the corresponding wrappers.
+     * If the entity is not a policy resource, it retrieves the {@link PolicyId} from the entity using the
+     *
+     * @param entityId the ID of the entity for which the {@link PolicyId} is being retrieved or set.
+     * @param wrappers the wrappers containing the permission checks for the entity.
+     * @param command the {@link CheckPermissions} command containing the headers for the request.
+     * @return a {@link CompletableFuture} that completes when the {@link PolicyId} is set for all wrappers.
+     */
+    private CompletableFuture<Void> retrieveOrSetPolicyId(final String entityId,
```
thjaeckle, line 177: This is problematic as this relies on a mutable `PermissionCheckWrapper` class.
In order to avoid concurrency issues, we must get rid of the mutability of `PermissionCheckWrapper`.

gateway/service/src/main/resources/gateway.conf
thjaeckle, line 68: Those lines can be removed, as we do not create an "Extension" here ..

policies/api/src/main/java/org/eclipse/ditto/policies/api/commands/sudo/CheckPolicyPermissionsResponse.java
```diff
@@ -0,0 +1,214 @@
...
+ */
+@Immutable
+@JsonParsableCommandResponse(type = CheckPolicyPermissionsResponse.TYPE)
+public final class CheckPolicyPermissionsResponse extends AbstractCommandResponse<CheckPolicyPermissionsResponse>
+        implements PolicySudoQueryCommandResponse<CheckPolicyPermissionsResponse> {
+
+    /**
+     * The type of this response.
+     */
+    public static final String TYPE = "checkPolicyPermissionsResponse";
```
thjaeckle, line 60: Wrong "TYPE" string - responses must also use their `TYPE_PREFIX` + the command name, so here:
```
public static final String TYPE = TYPE_PREFIX + CheckPolicyPermissions.NAME;
```

thjaeckle (changes requested):
Great progress, @hu-ahmed - thanks a lot. 👍 
I did another round of review and provided feedback inline, please have a look :)

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
hu-ahmed: but Java Maps only support boxed types

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/checkpermissions/PermissionCheckWrapper.java
hu-ahmed: but the policyId meant to be mutable as I set it later when I retrieve the policyId from the things service or from the entityId in the retrieveOrSetPolicyId

gateway/service/src/main/resources/gateway.conf
hu-ahmed: it's already removed but I forgot to push this change

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
thjaeckle: Ooops, absolutely right 👍

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,398 @@
...
+     * @param enrichedGroupedByEntityId the permission checks grouped by entity ID and enriched with policyId.
+     */
+    private CompletionStage<Void> handlePolicyRetrievalCompletion(final CheckPermissions command,
+            final Map<String, Boolean> permissionResults,
+            final Map<String, Map<String, PermissionCheckWrapper>> enrichedGroupedByEntityId) {
+        final Map<PolicyId, Map<String, ResourcePermissions>> aggregatedPermissions =
+                groupPermissionsByPolicyId(enrichedGroupedByEntityId);
+
+        final List<CompletionStage<Void>> permissionCheckFutures = aggregatedPermissions.entrySet().stream()
+                .map(aggregateEntry -> checkPermissionsForPolicy(aggregateEntry, command, permissionResults))
```
thjaeckle, line 233: @hu-ahmed I had a closer look here .. this is a potential concurrency issue - as you modify the (non-concurrent) `permissionResults` Map as "side effect" in `checkPermissionsForPolicy`.
But we also should not "just" use `ConcurrentMap` - instead change it to a functional style.

To change that to a functional style, `checkPermissionsForPolicy` should instead return a CompletableFuture of an immutable map of calculated permission results.

And in this method we would "combine"/merge all of the calculated maps to a combined map.

Sorry I didn't see that before.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/AbstractHttpRequestActor.java
```diff
@@ -346,6 +348,15 @@ private void rememberResponseLocationUri(final CommandResponse<?> commandRespons
         }
     }
 
+    private void handleCheckPermissions(final CheckPermissions command) {
+        final ActorRef checkPermissionsActor = getContext().actorOf(
+                CheckPermissionsActor.props(proxyActor, getSelf(), getReceiveTimeout(command, commandConfig)),
+                CheckPermissionsActor.ACTOR_NAME
```
thjaeckle, line 354: @hu-ahmed another thing which I didn't see before ..
We cannot use the static name here to create the actor.
As we would get a "Actor name conflict" when handling 2 `checkPermissions` API calls at the same time.

Instead, just start the actor without providing a name - Pekko will then create a random actor name.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/actors/CheckPermissionsActor.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /**
+     * Handles the {@link CheckPermissions} message by initializing the permission results,
+     * grouping permission checks by entity, and performing policy retrievals.
+     *
+     * @param command the {@link CheckPermissions} message to handle.
+     */
+    private void handleCheckPermissions(final CheckPermissions command) {
+        final List<String> permissionOrder = new LinkedList<>(command.getPermissionChecks().keySet());
```
thjaeckle, line 121: @hu-ahmed an `ArrayList<>()` is completely sufficient here as it also maintains the order ..

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -10430,6 +10455,43 @@ components:
...
+    PermissionEntityCheck:
+      type: object
+      description: "Details for a specific permission check request."
+      properties:
+        resource:
+          type: string
+          description: "Resource path the permission check applies to."
+        entityId:
+          type: string
+          description: "ThingId of the entity performing the action."
```
thjaeckle, line 10481: `entityId` is not always a "ThingId" .. it can also be a PolicyId .. depdending on the `resource` ..
We should make that more clear here ..

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -10430,6 +10455,43 @@ components:
...
+          resource: "message:/features/lamp/inbox/toggle"
+          entityId: "org.eclipse.ditto:some-thing-1"
+          hasPermissions: [ "WRITE" ]
+    PermissionEntityCheck:
+      type: object
+      description: "Details for a specific permission check request."
+      properties:
+        resource:
+          type: string
+          description: "Resource path the permission check applies to."
```
thjaeckle, line 10478: Maybe we could be more specific here about the format of that `resource` .. Or even provide a regex `pattern`?

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -10430,6 +10455,43 @@ components:
...
+        entityId:
+          type: string
+          description: "ThingId of the entity performing the action."
+        hasPermissions:
+          type: array
+          items:
+            type: string
```
thjaeckle, line 10485: we could even define an `enum` here of `"READ"` and `"WRITE"`

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -10430,6 +10455,43 @@ components:
...
+      description: "Request to check permissions for various entities and resources."
+      additionalProperties:
+        $ref: '#/components/schemas/PermissionEntityCheck'
+      example:
+        lamp_reader:
+          resource: "thing:/features/lamp/properties/on"
+          entityId: "org.eclipse.ditto:some-thing-1"
+          hasPermissions: [ "READ" ]
+        lamp_toggler:
+          resource: "message:/features/lamp/inbox/toggle"
```
thjaeckle, line 10469: this example is wrong ..
the path should be instead:
```
"message:/features/lamp/inbox/messages/toggle"
```

documentation/src/main/resources/pages/ditto/baic-auth-checkpermissions.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Checking Permissions for Resources  
+keywords: permissions, authorization, resources, policy, checkPermissions  
+tags: [model]  
+permalink: check-permissions.html
```
thjaeckle, line 5: the permalink name should be the same as the filename
And the filename contains a typo: `baic-auth-checkpermissions.md`

documentation/src/main/resources/pages/ditto/httpapi-concepts.md
```diff
@@ -138,6 +139,29 @@ This maps to the following HTTP API endpoints:
...
+{
+    "lamp_reader": {
+        "resource": "thing:/features/lamp/properties/on",
+        "entityId": "org.eclipse.ditto:some-thing-1",
+        "hasPermissions": [
+            "READ"
+        ]
+    },
+    "lamp_toggler": {
+        "resource": "message:/features/lamp/inbox/toggle",
```
thjaeckle, line 155: same mistake as in OpenAPI - wrong path.
Should be:
```
"message:/features/lamp/inbox/messages/toggle"
```

documentation/src/main/resources/pages/ditto/httpapi-concepts.md
```diff
@@ -138,6 +139,29 @@ This maps to the following HTTP API endpoints:
 * `/policies/{policyId}/entries/{entryLabel-1}/resources`: accessing the resources of a single `Policy` entry with the
   label `{entryLabel-1}`
 
+#### `/checkPermissions` in API 2
+The checkPermissions endpoint verifies permissions for specified entities on various resources, ensuring they have the necessary access rights as defined in the policies.
```
thjaeckle, line 143: I think we should move this to the new `check-permissions.html` page.  
Especially the example.

And then maybe only link to that page inside this section.

thjaeckle (changes requested):
Added some remarks for the added documentation.

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -10466,7 +10466,7 @@ components:
           entityId: "org.eclipse.ditto:some-thing-1"
           hasPermissions: [ "READ" ]
         lamp_toggler:
-          resource: "message:/features/lamp/inbox/toggle"
+          resource: "message:/features/lamp/inbox/message/toggle"
```
thjaeckle, line 10469: "message:/features/lamp/inbox/message**s**/toggle"

the plural "s" is missing ;)

documentation/src/main/resources/pages/ditto/baic-auth-checkpermissions.md
```diff
@@ -2,10 +2,10 @@
 title: Checking Permissions for Resources  
 keywords: permissions, authorization, resources, policy, checkPermissions  
 tags: [model]  
-permalink: check-permissions.html  
+permalink: basic-auth-checkpermissions.html
```
thjaeckle, line 5: the filename is still wrong:
`baic-auth-checkpermissions.md` (missing the "s" in basic)

thjaeckle (changes requested):
Some few minor fixes needed ..

thjaeckle (approved):

thjaeckle:
@hu-ahmed changes look good to me 👍 Thanks a lot for your patience.

Looking forward to a system test: https://github.com/eclipse-ditto/ditto-testing

PR #2050, opened 2024-10-24
fix that only aggregated metrics for filters which actually matched >0 things are reported

thjaeckle:
* also "forbid" to have multiple filters matching at the same time


alstanchev (approved):
LGTM 👍

PR #2051, opened 2024-10-25
enhance Helm chart to configure Ditto services via ConfigMap

thjaeckle:
* instead of configuring via System Properties

Right now, configuration of very dynamic settings (like e.g. custom metrics or entity creation) is done via SystemProperties in the Helm chart.
This can lead to some issues - e.g. if the command line arguments get too long, `trivy`, the wrapper process for starting the JVM can no longer handle those.
Or another problem is that the Kubernetes "Pods" HTTP API will return all environment variables and all system-properties as part of its response when querying for other Ditto pods during bootstrapping a cluster.

Moving the very dynamic configuration to config files (mounted via ConfigMaps) is an elegant solution for that.

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.6.1  # chart version is effectively set by release-job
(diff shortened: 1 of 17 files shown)


thjaeckle:
@kalinkostashki could you have a look, please?
@alstanchev you might as well, of course :)

thjaeckle:
@kalinkostashki ping

kalinkostashki:
> @kalinkostashki ping

on it :)

kalinkostashki (approved):

kalinkostashki (approved):
LGTM :+1:

PR #2060, opened 2024-11-14
fixed tracing and moving traces to their correct parent in many places

thjaeckle:
After enabling tracing in our Ditto installation, we found several places where spans are not correctly assigned to their parent spans and also where e.g. HTTP endpoints are reported falsely as `/other`.

This PR fixes that in a lot of encountered places.


alstanchev (approved):
LGTM 👍

PR #2063, opened 2024-11-15
Enhance Helm chart by startupProbe

thjaeckle:
* and more configuration options, e.g. several topologySpreadConstraints
* also configure "ephemeral-storage" requests and limits


thjaeckle:
@kalinkostashki @alstanchev whoever wants to, a quick look at this would be nice.

By using `startupProbe`, the `initialDelaySeconds` of the readinessProbe can be removed - as the readinessProbe is only started after the startup probe succeeded.

We also had the requirement to configure more than a single `topologySpreadConstraint`, so I changed that to be an array.

PR #2069, opened 2024-11-26
Fix resolving WoT submodels from extended model

thjaeckle:
* right now Ditto `tm:extends` resolving does not preserve the "extended" submodels from a parent model because merging of arrays will replace array content
* this provides a fix to add special handling for merging the "links" array


thjaeckle:
@alstanchev I would like to also add this fix for the upcoming 3.6.3 release.
We noticed that when applying some more complex extension hierarchies in WoT models.

alstanchev (approved):
LGTM 👍

PR #2073, opened 2024-12-03
fix aggregation metrics config unnecessary force of tags

alstanchev:
1.  Tags resolving relies on the fields configured in the group-by field of config but group-by does not rely on tags. 
Some times we may want to group by a field that we don't want as a tag so this fix removes the unwanted forcing af adding tags.

2.  Also renamed the variable of AggregateThingsMetricsActor to comply with the overall code pattern and to easily find the actor behind the ActorRef


thjaeckle (approved):
Looks good to me.
Thanks for the fix 👍

PR #2075, opened 2024-12-16
Fix WoT model validation calculating wrong path

thjaeckle:
* for enforcing feature properties in all submodels


kalinkostashki (approved):
LGTM :+1:

PR #2078, opened 2024-12-17
fix missing headers from message after js mapper

### LINKED ISSUE #2077: Missing headers added from published messeage added by js payload mapper
Headers added to an outbound message by a js mapper can be missing in the published message.

alstanchev:
fixes #2077


thjaeckle (approved):
Looks good to me 👍 
Nice fix

PR #2085, opened 2025-01-07
fix configuration bug in WoT validation-context and log-warning-instead-of-failing-api-calls setting

thjaeckle:


thjaeckle:
@alstanchev could you have a brief look?

PR #2086, opened 2025-01-07
Fix that generated default for WoT array properties contained one generated "neutral element"

thjaeckle:
… even if no "minItems" was defined


thjaeckle:
@alstanchev could you have a brief look?

alstanchev (approved):
LGTM 👍

PR #2087, opened 2025-01-08
add missing thingId to log of failed WoT validations

thjaeckle:
* without it, it is not easy to detect which thing was not valid
* also log in case API call is rejected, but on INFO level


alstanchev (approved):
LGTM 👍

alstanchev (approved):

PR #2088, opened 2025-01-09
fix Gateway trace not keeping correct parent "hierarchy" when receiving existing "traceparent" header

thjaeckle:


thjaeckle:

alstanchev:
Already did but i was waiting to see if there will be any more commits.

alstanchev (approved):
LGTM 👍

PR #2091, opened 2025-01-13
fix wrongly calculated path in WoT validation errors

thjaeckle:
* also fix traceparent-trace-id logging


thjaeckle:
@alstanchev or @kalinkostashki could you have a look?

PR #2093, opened 2025-01-14
fix wrongly done validation of removal of desired feature properties

thjaeckle:
* it was enforced that "required" properties could not be removed
* this does not make sense for desired properties, as they are all optional by convention


kalinkostashki (approved):
LGTM :+1:

PR #2094, opened 2025-01-14
fix spans for consuming messages to not be the "parent" span for following spans

thjaeckle:


thjaeckle:
@alstanchev or @kalinkostashki could you have a look?

kalinkostashki (approved):
LGTM :+1:

PR #2100, opened 2025-01-20
fix NPE in ScriptedOutgoingMapping when value from script resolved to null

thjaeckle:
Encountered through failing SystemTests on `master` branch:
https://github.com/eclipse-ditto/ditto/runs/35877513588#r25


thjaeckle:
It seems like message headers for AMQP 1.0 and RabbitMQ are somehow lowercased.

Before the integration test `ensureSourceAuthenticationPlaceholderSubstitution()` did send a header named `headerId` - and used that header in a placeholder substitution.  
Running the test locally I found out that this header is received at Ditto as `headerid` so the capital `I` gets lowercases to `i`.  
I found the same for `messageId` header - lowercased to `messageid`.

For MQTT5 e.g. this test passes, so case is preserved there.

@alstanchev do you have any idea why this could happen?  
Maybe a dependency update?

alstanchev:
No idea, doesn't ring any bells.

alstanchev (approved):
LGTM 👍

Issue #1843, opened 2023-12-18
Support updating referenced WoT ThingModel minor version for Ditto Things


thjaeckle:
Currently, the WoT ThingModel is only taken into account when creating new Things in Ditto. In this case, a JSON skeleton of the Thing based on the WoT TM is created.  
Over time, a WoT ThingModel might however evolve, e.g. new `properties`, new `actions` or new `submodels`, etc. are added.

Those evolvements can be categorised in:
* non-breaking evolvements (according to [Semantic Versioning](https://semver.org) changes to the **MINOR** version):
   * only additions to existing functionality without removing existing properties/actions/events
* breaking evolvements (according to [Semantic Versioning](https://semver.org) changes to the **MAJOR** version)
   * also containing breaking functionality with renaming/removing existing properties/actions/events
   * **out of scope for this issue** - as this would require e.g. also deleting stuff from things or renaming

In my experience, non-breaking evolvements can happen very regularly, as the (exposed) functionality of IoT devices evolves quickly over time.  
For such non-breaking evolvements, Ditto shall provide means to "migrate" existing things to a newer **MINOR** version of referenced WoT ThingModels.

Concretely, the following steps would have be done when updating the **minor** version of a WoT ThingModel in a Ditto thing's `definition`:
* `definition` of a **Thing** is modified (either via `ModifyDefinition` or via `MergeThing` for the `"definition"`)
* Ditto checks if the "minor" version increased - and only if it did - performs a "migration":
   * resolving all "submodel" versions of the new "minor" thing model
   * update all feature `definition` fields with the updated "submodel" models
   * potentially adding previously non-yet-existing features in the process
   * for existing features / "submodels" which received a **minor** update, for each pair of `<old model version>,<new model version>`:
      * generate JSON skeleton for `<old model version>`
      * generate JSON skeleton for `<new model version>`
      * calculate the "diff" as Json Merge Patch from new to old JSON skeleton
      * for all "new added" (non-optional) properties:
         * create the property with the defined `default` value
         * but only if the Thing does not yet contain a "real" value for the property yet (which could already be sent by the real device - we don't want to overwrite "real" values with default values)

The algorithm might still not be "complete" - but in a nutshell, we have to handle:
* updating Thing `attributes` with default values for new TM "properties"
* updating Feature `properties`  with default values for new TM "properties"
* updating Feature `definition`s
* updating the Thing's `definition` if all was successful
   * TODO: what to do in case of an error?
[...]

PR #2103, opened 2025-01-27
add new wot-models definition update endpoint #1843

hu-ahmed:
Resolves: #1843 

The migrateThingDefinition endpoint introduces a comprehensive mechanism for migrating a Thing's definition while ensuring data consistency and applying necessary migrations. This endpoint is designed to facilitate seamless updates by incorporating a structured approach to definition changes and migration payloads.

Key Features:
- Definition Update with Skeleton Generation:
The provided Thing definition URL is used to generate a new skeleton Thing, which serves as the baseline for the update.
The skeleton includes a hierarchical representation of the Thing, ensuring consistency with the WoT model.

- Patch Condition Evaluation:
The migration payload can include multiple patch conditions, evaluated using RQL-based expressions to determine whether specific updates should be applied.
This allows for conditional updates based on existing Thing attributes or feature values.

- Merging Strategy:
The update process merges the newly generated skeleton with the existing Thing while preserving essential attributes and properties.
An optional initialization step applies default values from the skeleton to fill in missing or incomplete properties.

- Migration Payload Application:
The provided migration payload is selectively merged into the updated Thing, allowing for targeted modifications without overriding all properties.
Ensures data integrity by applying only necessary changes based on predefined rules.

- Dry-Run Mode:
The API supports a dry-run option (dryRun: true), allowing users to simulate the migration without persisting changes.

- Validation and Persistence:
Before finalizing the update, the modified Thing undergoes WoT model validation based on the current configuration, so either it will be rejected or logged as warning in case of inconsistency.

- Ditto Protocols supported: 
this command is also available for other ditto protocols

- API Usage:
Endpoint:
POST /api/2/things/{thingId}/migrateDefinition
Request Body:

```json
{
  "thingDefinitionUrl": "https://models.example.com/thing-definition-1.0.0.tm.jsonld",
  "migrationPayload": {
    "attributes": {
      "manufacturer": "New Corp",
      "location": "Berlin, main floor"
    },
    "features": {
      "thermostat": {
        "properties": {
          "status": {
            "temperature": {
              "value": 23.5,
              "unit": "DEGREE_CELSIUS"
            }
          }
        }
      }
    }
  },
  "patchConditions": {
    "thing:/features/thermostat": "not(exists(/features/thermostat))"
  },
  "initializeMissingPropertiesFromDefaults": true
}
```

### DIFF
--- a/things/model/src/main/java/org/eclipse/ditto/things/model/signals/commands/modify/MigrateThingDefinition.java
+++ b/things/model/src/main/java/org/eclipse/ditto/things/model/signals/commands/modify/MigrateThingDefinition.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.ditto.things.model.signals.commands.modify;
+
+import static org.eclipse.ditto.base.model.common.ConditionChecker.checkNotNull;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import javax.annotation.concurrent.Immutable;
+
+import org.eclipse.ditto.base.model.headers.DittoHeaders;
+import org.eclipse.ditto.base.model.json.FieldType;
+import org.eclipse.ditto.base.model.json.JsonParsableCommand;
+import org.eclipse.ditto.base.model.json.JsonSchemaVersion;
+import org.eclipse.ditto.base.model.signals.UnsupportedSchemaVersionException;
+import org.eclipse.ditto.base.model.signals.commands.AbstractCommand;
+import org.eclipse.ditto.json.JsonFactory;
+import org.eclipse.ditto.json.JsonField;
+import org.eclipse.ditto.json.JsonFieldDefinition;
+import org.eclipse.ditto.json.JsonObject;
+import org.eclipse.ditto.json.JsonObjectBuilder;
+import org.eclipse.ditto.json.JsonPointer;
+import org.eclipse.ditto.json.JsonValue;
+import org.eclipse.ditto.policies.model.ResourceKey;
+import org.eclipse.ditto.things.model.ThingId;
+import org.eclipse.ditto.things.model.signals.commands.ThingCommand;
+import org.eclipse.ditto.things.model.signals.commands.ThingCommandSizeValidator;
+
+/**
+ * Command to update a Thing's definition and apply a migration payload.
+ *
+ * @since 3.7.0
+ */
+@Immutable
+@JsonParsableCommand(typePrefix = ThingCommand.TYPE_PREFIX, name = MigrateThingDefinition.NAME)
+public final class MigrateThingDefinition extends AbstractCommand<MigrateThingDefinition>
+        implements ThingModifyCommand<MigrateThingDefinition> {
+
+    /**
+     * Name of the "Migrate Thing Definition" command.
+     */
+    public static final String NAME = "migrateThingDefinition";
+
+    /**
+     * Type of this command.
+     */
+    public static final String TYPE = TYPE_PREFIX + NAME;
+
+    private final ThingId thingId;
+    private final String thingDefinitionUrl;
+    private final JsonObject migrationPayload;
+    private final Map<ResourceKey, String> patchConditions;
+    private final boolean initializeMissingPropertiesFromDefaults;
+
+
+    private MigrateThingDefinition(final ThingId thingId,
+            final String thingDefinitionUrl,
+            final JsonObject migrationPayload,
+            final Map<ResourceKey, String> patchConditions,
+            final boolean initializeMissingPropertiesFromDefaults,
+            final DittoHeaders dittoHeaders) {
+        super(TYPE, dittoHeaders);
+        this.thingId = checkNotNull(thingId, "thingId");
+        this.thingDefinitionUrl = checkNotNull(thingDefinitionUrl, "thingDefinitionUrl");
+        this.migrationPayload = checkJsonSize(checkNotNull(migrationPayload, "migrationPayload"), dittoHeaders);
+        this.patchConditions = Collections.unmodifiableMap(checkNotNull(patchConditions, "patchConditions"));
+        this.initializeMissingPropertiesFromDefaults = initializeMissingPropertiesFromDefaults;
+        checkSchemaVersion();
+    }
+
+    /**
+     * Factory method to create a new {@code MigrateThingDefinition} command.
+     *
+     * @param thingId the Thing ID.
+     * @param thingDefinitionUrl the URL of the new Thing definition.
+     * @param migrationPayload the migration payload.
+     * @param patchConditions the patch conditions.
+     * @param initializeMissingPropertiesFromDefaults whether to initialize properties.
+     * @param dittoHeaders the Ditto headers.
+     * @return the created {@link MigrateThingDefinition} command.
+     */
+    public static MigrateThingDefinition of(final ThingId thingId,
+            final String thingDefinitionUrl,
+            final JsonObject migrationPayload,
+            final Map<ResourceKey, String> patchConditions,
+            final boolean initializeMissingPropertiesFromDefaults,
+            final DittoHeaders dittoHeaders) {
+        return new MigrateThingDefinition(thingId, thingDefinitionUrl, migrationPayload,
+                patchConditions, initializeMissingPropertiesFromDefaults, dittoHeaders);
+    }
+
+    /**
+     * Creates a new {@code MigrateThingDefinition} from a JSON object.
+     *
+     * @param jsonObject the JSON object from which to create the command.
+     * @param dittoHeaders the Ditto headers.
+     * @return the created {@code MigrateThingDefinition} command.
+     */
+    public static MigrateThingDefinition fromJson(final JsonObject jsonObject, final DittoHeaders dittoHeaders) {
+        final String thingIdStr = jsonObject.getValueOrThrow(ThingCommand.JsonFields.JSON_THING_ID);
+        final String thingDefinitionUrl = jsonObject.getValueOrThrow(JsonFields.JSON_THING_DEFINITION_URL);
+        final JsonObject migrationPayload = jsonObject.getValue(JsonFields.JSON_MIGRATION_PAYLOAD)
+                .map(JsonValue::asObject).orElse(JsonFactory.newObject());
+        final JsonObject patchConditionsJson = jsonObject.getValue(JsonFields.JSON_PATCH_CONDITIONS)
+                .map(JsonValue::asObject).orElse(JsonFactory.newObject());
+        final Map<ResourceKey, String> patchConditions = patchConditionsJson.stream()
+                .collect(Collectors.toMap(
+                        field -> ResourceKey.newInstance(field.getKey()),
+                        field -> field.getValue().asString()
+                ));
+
+        final boolean initializeMissingPropertiesFromDefaults =
+                jsonObject.getValue(JsonFields.JSON_INITIALIZE_MISSING_PROPERTIES_FROM_DEFAULTS)
+                        .orElse(false);
+
+        return new MigrateThingDefinition(
+                ThingId.of(thingIdStr),
+                thingDefinitionUrl,
+                migrationPayload,
+                patchConditions,
+                initializeMissingPropertiesFromDefaults,
+                dittoHeaders);
+    }
+
+    @Override
+    public ThingId getEntityId() {
+        return thingId;
+    }
+
+    public String getThingDefinitionUrl() {
+        return thingDefinitionUrl;
+    }
+
+    public JsonObject getMigrationPayload() {
+        return migrationPayload;
+    }
+
+    public Map<ResourceKey, String> getPatchConditions() {
+        return patchConditions;
+    }
+
+    public boolean isInitializeMissingPropertiesFromDefaults() {
+        return initializeMissingPropertiesFromDefaults;
+    }
+
+    @Override
+    public Optional<JsonValue> getEntity() {
+        // This command doesn't represent an entity directly.
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<JsonValue> getEntity(final JsonSchemaVersion schemaVersion) {
+        return getEntity();
+    }
+
+    @Override
+    public MigrateThingDefinition setEntity(final JsonValue entity) {
+        return this;
+    }
+
+    @Override
+    public JsonPointer getResourcePath() {
+        return JsonPointer.empty();
+    }
+
+    @Override
+    public boolean changesAuthorization() {
+        return false;
+    }
+
+    @Override
+    public Category getCategory() {
+        return Category.MIGRATE;
+    }
+
+    @Override
+    public MigrateThingDefinition setDittoHeaders(final DittoHeaders dittoHeaders) {
+        return new MigrateThingDefinition(
+                thingId, thingDefinitionUrl, migrationPayload, patchConditions, initializeMissingPropertiesFromDefaults,
+                dittoHeaders);
+    }
+
+    @Override
+    protected void appendPayload(final JsonObjectBuilder jsonObjectBuilder,
+            final JsonSchemaVersion schemaVersion,
+            final Predicate<JsonField> predicateParameter) {
+        final Predicate<JsonField> predicate = schemaVersion.and(predicateParameter);
+        jsonObjectBuilder.set(ThingCommand.JsonFields.JSON_THING_ID, thingId.toString(), predicate);
+        jsonObjectBuilder.set(JsonFields.JSON_THING_DEFINITION_URL, thingDefinitionUrl, predicate);
+        jsonObjectBuilder.set(JsonFields.JSON_MIGRATION_PAYLOAD, migrationPayload, predicate);
+        jsonObjectBuilder.set(JsonFields.JSON_INITIALIZE_MISSING_PROPERTIES_FROM_DEFAULTS,
+                initializeMissingPropertiesFromDefaults, predicate);
+
+        if (!patchConditions.isEmpty()) {
+            final JsonObjectBuilder conditionsBuilder = JsonFactory.newObjectBuilder();
+            patchConditions.forEach(conditionsBuilder::set);
+            jsonObjectBuilder.set(JsonFields.JSON_PATCH_CONDITIONS, conditionsBuilder.build(), predicate);
+        }
+    }
+
+    private void checkSchemaVersion() {
+        final JsonSchemaVersion implementedSchemaVersion = getImplementedSchemaVersion();
+        if (!implementsSchemaVersion(implementedSchemaVersion)) {
+            throw UnsupportedSchemaVersionException.newBuilder(implementedSchemaVersion).build();
+        }
+    }
+
+    private JsonObject checkJsonSize(final JsonObject value, final DittoHeaders dittoHeaders) {
+        ThingCommandSizeValidator.getInstance().ensureValidSize(
+                value::getUpperBoundForStringSize,
+                () -> value.toString().length(),
+                () -> dittoHeaders);
+        return value;
+    }
+
+    /**
+     * An enumeration of the JSON fields of an {@code MigrateThingDefinition} command.
+     */
+    @Immutable
+    public static final class JsonFields {
+
+        public static final JsonFieldDefinition<String> JSON_THING_DEFINITION_URL =
+                JsonFactory.newStringFieldDefinition("thingDefinitionUrl", FieldType.REGULAR, JsonSchemaVersion.V_2);
+
+        public static final JsonFieldDefinition<JsonObject> JSON_MIGRATION_PAYLOAD =
+                JsonFactory.newJsonObjectFieldDefinition("migrationPayload", FieldType.REGULAR, JsonSchemaVersion.V_2);
+
+        public static final JsonFieldDefinition<JsonObject> JSON_PATCH_CONDITIONS =
+                JsonFactory.newJsonObjectFieldDefinition("patchConditions", FieldType.REGULAR, JsonSchemaVersion.V_2);
+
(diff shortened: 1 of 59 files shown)


thjaeckle (changes requested):
Hi @hu-ahmed and thanks a lot for providing this PR.

Looks already really promising .. I did a quick check and left a few comments before going deeper :)

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+            final long nextRevision,
+            final MigrateThingDefinition command,
+            @Nullable final Metadata metadata) {
+
+        final Thing existingThing = getEntityOrThrow(thing);
+        final Instant eventTs = getEventTimestamp();
+
+        return handleUpdateDefinition(context, existingThing, eventTs, nextRevision, command, metadata)
+                .toCompletableFuture()
+                .join();
```
thjaeckle, line 95: No joining completableFutures like this - this will block the main actor thread and could cause death of the whole actor system.
There are APIs at `ResultFactory.newMutationResult` which support completionStages instead of sync API.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+@Immutable
+public final class MigrateThingDefinitionStrategy extends AbstractThingModifyCommandStrategy<MigrateThingDefinition> {
+
+    private static final ThingResourceMapper<Thing, Optional<EntityTag>> ENTITY_TAG_MAPPER =
+            ThingResourceMapper.from(EntityTagCalculator.getInstance());
+
+    private static final TimePlaceholder TIME_PLACEHOLDER = TimePlaceholder.getInstance();
+
+
+    public MigrateThingDefinitionStrategy(final ActorSystem actorSystem) {
```
thjaeckle, line 79: Constructor should be package private - and have javadoc

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+        );
+    }
+
+    private JsonObject evaluatePatchConditions(final Thing existingThing,
+            final JsonObject migrationPayload,
+            final Map<ResourceKey, String> patchConditions,
+            final DittoHeaders dittoHeaders) {
+        final JsonObjectBuilder adjustedPayloadBuilder = migrationPayload.toBuilder();
+
+        for (Map.Entry<ResourceKey, String> entry : patchConditions.entrySet()) {
```
thjaeckle, line 154: `entry` should be final

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+
+            final JsonPointer resourcePointer = JsonFactory.newPointer(resourceKey.getResourcePath());
+            if (!conditionMatches && doesMigrationPayloadContainResourceKey(migrationPayload, resourcePointer)) {
+                adjustedPayloadBuilder.remove(resourcePointer);
+            }
+        }
+
+        return adjustedPayloadBuilder.build();
+    }
+    public boolean doesMigrationPayloadContainResourceKey(JsonObject migrationPayload, JsonPointer pointer) {
```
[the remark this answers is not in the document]
thjaeckle: and why is this method `public`? it could be even `private static`

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+            final var predicate = ThingPredicateVisitor.apply(criteria,
+                    PlaceholderFactory.newPlaceholderResolver(TIME_PLACEHOLDER, new Object()));
+
+            return predicate.test(existingThing);
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    private CompletionStage<Thing> generateSkeleton(final Context<ThingId> context,
```
thjaeckle, line 188: `context` param is not used

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+            final MigrateThingDefinition command,
+            final DittoHeaders dittoHeaders) {
+        return wotThingSkeletonGenerator.provideThingSkeletonForCreation(
+                        command.getEntityId(),
+                        ThingsModelFactory.newDefinition(command.getThingDefinitionUrl()),
+                        dittoHeaders
+                )
+                .thenApply(optionalSkeleton -> optionalSkeleton.orElseThrow());
```
thjaeckle, line 196: `orElseThrow` is not good here .. it will throw an `IllegalArgumentException` which will not contain any useful information about why this failed.
I would suggest to throw a more meaningful subclass of `WotException`, also maintaining `dittoHeaders` to have proper propagation of correlation-id, trace, etc.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+
+    private Thing extractDefinitions(final Thing thing, final String thingDefinitionUrl) {
+        var thingBuilder = ThingsModelFactory.newThingBuilder();
+        thingBuilder.setDefinition(ThingsModelFactory.newDefinition(thingDefinitionUrl));
+        thing.getFeatures().orElseGet(null).forEach(feature-> {
```
thjaeckle, line 203: you cannot pass `null` in `orElseGet` - this will lead to a NPE, as a (non-nullable) `Supplier` is needed

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+            thingBuilder.setFeature(feature.getId(), feature.getDefinition().get(), null);
+        });
+        return thingBuilder.build();
+    }
+
+
+    private Thing extractDefaultValues(Thing thing) {
+        var thingBuilder = ThingsModelFactory.newThingBuilder();
+        thingBuilder.setAttributes(thing.getAttributes().get());
```
thjaeckle, line 212: what if the thing does not have attributes?
then this `thing.getAttributes().get()` will fail with a NPE

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+        var thingBuilder = ThingsModelFactory.newThingBuilder();
+        thingBuilder.setDefinition(ThingsModelFactory.newDefinition(thingDefinitionUrl));
+        thing.getFeatures().orElseGet(null).forEach(feature-> {
+            thingBuilder.setFeature(feature.getId(), feature.getDefinition().get(), null);
+        });
+        return thingBuilder.build();
+    }
+
+
+    private Thing extractDefaultValues(Thing thing) {
```
thjaeckle, line 210: param should be `final`

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+        var thingBuilder = ThingsModelFactory.newThingBuilder();
+        thingBuilder.setAttributes(thing.getAttributes().get());
+        thing.getFeatures().orElseGet(null).forEach(feature-> {
```
thjaeckle, line 213: same as before .. we cannot pass `null` as `Supplier` to `orElseGet()`

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+ * - Skeleton generation extracts and merges Thing definitions and default values separately,
+ *   ensuring a clear distinction between structural updates and default settings.
+ * - After applying skeleton-based modifications and migration payloads, the changes are merged
+ * - Property initialization can be optionally enabled via the command, applying default values
+ *   to the updated Thing when set to true.
+ * - The resulting Thing undergoes validation to ensure compliance with WoT model constraints
+ *   before persisting changes.
+ */
+@Immutable
+public final class MigrateThingDefinitionStrategy extends AbstractThingModifyCommandStrategy<MigrateThingDefinition> {
```
thjaeckle, line 71: Unit test is missing for this strategy class.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+ *   conditions for fine-grained updates.
+ * - Patch conditions are evaluated using RQL-based expressions to determine which migration
+ *   payload entries should be applied.
+ * - Skeleton generation extracts and merges Thing definitions and default values separately,
+ *   ensuring a clear distinction between structural updates and default settings.
+ * - After applying skeleton-based modifications and migration payloads, the changes are merged
+ * - Property initialization can be optionally enabled via the command, applying default values
+ *   to the updated Thing when set to true.
+ * - The resulting Thing undergoes validation to ensure compliance with WoT model constraints
+ *   before persisting changes.
```
thjaeckle, line 68: when doing a list in JavaDoc, you should use:
```
<ul>
<li>...</li>
</ul>
```
This will be properly rendered as list in e.g. IDE popups.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/commands/modify/MigrateThingDefinition.java
```diff
@@ -0,0 +1,312 @@
...
+            final JsonObject migrationPayload,
+            final Map<ResourceKey, String> patchConditions,
+            final Boolean initializeMissingPropertiesFromDefaults,
+            final DittoHeaders dittoHeaders) {
+        super(TYPE, FeatureToggle.checkMergeFeatureEnabled(TYPE, dittoHeaders));
+        this.thingId = checkNotNull(thingId, "thingId");
+        this.thingDefinitionUrl = checkNotNull(thingDefinitionUrl, "thingDefinitionUrl");
+        this.migrationPayload = checkJsonSize(checkNotNull(migrationPayload, "migrationPayload"), dittoHeaders);
+        this.patchConditions =
+                Collections.unmodifiableMap(patchConditions != null ? patchConditions : Collections.emptyMap());
```
thjaeckle, line 81: this parameter `patchConditions` is not marked as `@Nullable` and (I think) cannot be `null` from the code flow.  
So no need to check for null ..

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,279 @@
...
+                                    command.getEntityId(),
+                                    JsonPointer.empty(),
+                                    updatedThing.toJson(),
+                                    dittoHeaders
+                            );
+
+                            final MergeThingStrategy mergeStrategy = new MergeThingStrategy(context.getActorSystem());
+                            final Result<ThingEvent<?>> mergeResult = mergeStrategy.doApply(
+                                    context,
+                                    existingThing,
```
thjaeckle, line 137: Why apply the mergeStrategy for the `existingThing` here?
I would expect the `updatedThing`.

So currently this e.g. validates the old, existingThing (as part of the MergeThingStrategy) against the WoT model instead of the new one.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Override
+    public ThingMigrated setDittoHeaders(final DittoHeaders dittoHeaders) {
+        return of(thingId, path, value, getRevision(), getTimestamp().orElse(null), dittoHeaders,
+                getMetadata().orElse(null));
+    }
+
+    @Override
+    public Command.Category getCommandCategory() {
+        return Command.Category.MODIFY;
```
thjaeckle, line 136: I am unsure about the commandCategory here ..
Modify has "replace" semantics .. Merge has "merge" semantics - however looking at how this method is called from elsewhere, I fear that both would have the wrong outcome.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,222 @@
...
+                getMetadata().orElse(null));
+    }
+
+    @Override
+    public JsonPointer getResourcePath() {
+        return path;
+    }
+
+    public JsonValue getValue() {
+        return value;
```
thjaeckle, line 151: What does this `value` represent?
This is too technical for me - I don't understand only looking at this class what "value" does or should persist.

Part of the problem is that public getters are not documented ;)

protocol/src/main/java/org/eclipse/ditto/protocol/mappingstrategies/ThingMigratedEventMappingStrategies.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public JsonifiableMapper<ThingMigrated> find(final String type) {
+        return ThingMigratedEventMappingStrategies::thingMigrated;
+    }
+
+    private static ThingMigrated thingMigrated(final Adaptable adaptable) {
+        return ThingMigrated.of(
+                thingIdFrom(adaptable),
+                JsonPointer.of(adaptable.getPayload().getPath().toString()),
+                adaptable.getPayload().getValue().orElse(null),
```
thjaeckle, line 52: the passed in parameter is now allowed to be `null` - this will throw a NullPointer down the execution chain.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,222 @@
...
+     * Type of this event.
+     */
+    public static final String TYPE = TYPE_PREFIX + NAME;
+
+    private final ThingId thingId;
+    private final JsonPointer path;
+    private final JsonValue value;
+
+    private ThingMigrated(final ThingId thingId,
+            final JsonPointer path,
```
thjaeckle, line 67: As far as I can see, this is always `JsonPointer.empty()`.
Why should we have a dynamic parameter and a field (and a JSON representation) for that if it is always the same?

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,222 @@
...
+            final JsonValue value,
```
thjaeckle, line 68: As far as I can see, this is always a `Thing`, right?
So why not use `Thing` type and only in the JSON representation convert to `JsonObject` to store it in JSON.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingEventToThingConverter.java
```diff
@@ -121,6 +121,17 @@ private static Map<Class<?>, BiFunction<ThingEvent<?>, ThingBuilder.FromScratch,
                     return ThingsModelFactory.newThing(mergedJson.asObject());
                 }
         );
+        mappers.put(ThingMigrated.class,
+                (te, tb) -> {
```
thjaeckle, line 125: This shouldn't just be copy&paste from `ThingMerged` - ThingMigrated e.g. always contains the whole migrated thing - so using a `resourcePath` does not make sense.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,191 @@
...
+    public static final String TYPE = TYPE_PREFIX + NAME;
+
+    private final Thing thing;
+
+    private ThingMigrated(final Thing thing,
+                          final long revision,
+                          @Nullable final Instant timestamp,
+                          final DittoHeaders dittoHeaders,
+                          @Nullable final Metadata metadata) {
+        super(TYPE, thing.getEntityId().orElseThrow(() -> new NullPointerException("Thing has no ID!")), revision, timestamp, dittoHeaders, metadata);
```
thjaeckle, line 66: I would not throw `NullPointerException` - but rather `IllegalArgumentException`.
NPE is something really technical.

things/model/src/main/java/org/eclipse/ditto/things/model/signals/events/ThingMigrated.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Override
+    public ThingMigrated setEntity(final JsonValue entity) {
+        return of(ThingsModelFactory.newThing(entity.asObject()), getRevision(), getTimestamp().orElse(null),
+                getDittoHeaders(), getMetadata().orElse(null));
+    }
+
+    @Override
+    public JsonSchemaVersion[] getSupportedSchemaVersions() {
```
thjaeckle, line 149: This overwritten method can be removed .. same code as in parent class.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,365 @@
...
+            );
+        }
+
+        // 4. Apply migration and generate event
+        final CompletionStage<ThingEvent<?>> eventStage = validatedStage.thenApply(pair -> ThingMigrated.of(
+                pair.first(), nextRevision, eventTs, dittoHeaders,
+                metadata));
+
+        final CompletionStage<WithDittoHeaders> responseStage = validatedStage.thenApply(pair ->
+                appendETagHeaderIfProvided(command, MigrateThingDefinitionResponse.applied(existingThing.getEntityId().get(),
```
thjaeckle, line 159: `existingThing.getEntityId().get()` could be replaced with `context.getState()` - as `existingThing.getEntityId()` is optional, the `.get()` might throw an exception

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,365 @@
...
+                            .build();
+                });
+    }
+
+
+
+    private Thing extractDefinitions(final Thing thing) {
+        var thingBuilder = ThingsModelFactory.newThingBuilder();
+        thing.getFeatures().orElseGet(ThingsModelFactory::emptyFeatures).forEach(feature ->
+                thingBuilder.setFeature(feature.getId(), feature.getDefinition().get(), null));
```
thjaeckle, line 239: `feature.getDefinition().get()` might throw exception when the feature has no definition

also, passing `null` as last param is now allowed (at least in contract of ThingBuilder.setFeature(String featureId, FeatureDefinition featureDefinition, FeatureProperties featureProperties)`

The implementation however supports nullable `FeatureDefinition` and `FeatureProperties` .. 
So could you:
* change the signature of ThingBuilder.FromScratch.setFeature(...)` and make `FeatureDefinition` and `FeatureProperties` `@Nullable`?
* secondly, pass in `null` if the feature has no definition instead of `.get()` (which will throw an exception if optional was empty)

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -0,0 +1,365 @@
...
+     * Resolves conflicting JsonValue objects by recursively comparing them.
+     * If both values are JsonObjects, it calls {@link #removeConflicts(JsonObject, JsonObject)}
+     * to recursively filter out conflicting values. Otherwise, it returns null,
+     * indicating that the value should not be retained.
+     *
+     * @param defaultValue  The JsonValue from the default values object.
+     * @param existingValue The JsonValue from the existing values object.
+     * @return A filtered JsonObject if both values are objects; otherwise, null.
+     */
+    private static JsonValue resolveConflictingValues(final JsonValue defaultValue, final JsonValue existingValue) {
```
thjaeckle, line 318: Please change signature to `Optional<JsonValue>` - Ditto codebase always uses `Optional` for methods which might not return a value .. and never return `null` - as this will only cause NPEs.

protocol/src/main/java/org/eclipse/ditto/protocol/adapter/MigrateCommandResponseAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.ditto.protocol.TopicPath;
+import org.eclipse.ditto.things.model.signals.commands.modify.MigrateThingDefinitionResponse;
+
+import java.util.EnumSet;
+import java.util.Set;
+
+/**
+ * An {@code Adapter} mixin for merge command responses.
+ *
+ * @since 2.0.0
```
thjaeckle, line 24: wrong since ;)

thjaeckle (approved):
All review remarks were addressed.
Thanks for squashing into a single commit 👍 

And thanks a lot for this great addition to the Ditto WoT integration @hu-ahmed

PR #2105, opened 2025-01-29
fix tracing propagation in general

thjaeckle:
* remove many manual "traceparent" header removals which mitigated the root problem


thjaeckle:
in the past, I often fixed some trace propagation issues .. this one should fix basically all of the remaining issues :D

Issue #2108, opened 2025-01-31
MQTT connection fails to reconnect in case of DDNS record IP adress change


vvasilevbosch:
hivemq-mqtt-client uses java.net.InetSocketAddress for server address and caches it. Currently, ditto provides [resolved InetSocketAddress](https://github.com/eclipse-ditto/ditto/blob/9322980d2c13936831d599316ad231daca00986b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/mqtt/hivemq/client/HiveMqttClientFactory.java#L138), which is never re-resolved and not handled by the hivemq mqtt client. This causes mqtt connections to reconnect without success infinitely.

PR #2109, opened 2025-02-03
#2108 MQTT Connection - add posibility to configure if server address should be resolved by ditto or on demand by mqtt client

vvasilevbosch:
* add possibility to configure whether server address should be resolved by ditto in mqtt connection
* fix retry timeout strategy unable increase retry count and apply effective backoff

Fixes: #2108


thjaeckle (approved):
Thanks a lot for this fix @vvasilevbosch .. looks overall very good 👍

PR #2116, opened 2025-02-10
fix "If-Equal" header predicate for new different from predicate for existing

thjaeckle:
* causing that e.g. the existing Json did not contain "thingId" whereas the new one did


thjaeckle:

alstanchev (approved):
LGTM 👍

PR #2118, opened 2025-02-11
fixed regression in message response WoT validation done at wrong place

thjaeckle:
* leading to message responses not being of type "application/json" being rejected, even if WoT validation was deactivated


alstanchev (approved):
LGTM 👍

Issue #2115, opened 2025-02-10
Ditto connections metrics are lacking type "other" - as connection logs report to


thjaeckle:
Currently, whenever an Exception occurred e.g. in a Ditto connection's payload mapper, Ditto will produce a "connection log entry" with category `other` showing the exception/error.  
There is however no "metric" for category "other" which would count the very same error.

This is problematic as the Connection loggers are only meant to be activated temporarily when debugging or developing connections or payload mappers.  
It would be good to see on the metric level (as metrics are always gathered) if a connection contains errors/failures in the "other" category.

PR #2119, opened 2025-02-11
#2115: fix exceptions during outbound mapping not being recorded via "metrics"

thjaeckle:
Fixes: #2115


thjaeckle:

PR #2123, opened 2025-02-17
fix ExistsThingPredicateVisitor did not flatten arrays correctly

thjaeckle:
* so that "exists()" RQL containing arrays was not evaluated as expected


alstanchev (approved):
LGTM 👍

PR #2125, opened 2025-02-17
add throttling option to the thingSearch persistence updater

### LINKED ISSUE #2122: Provide option to throttle stream of things to update in search index after policy update
Whenever a Ditto managed policy is changed, the Ditto `things-search` service reacts on that change in order to update all search index documents to reflect the changes made to the Policy in the search index.

This mechanism works well if only few things are affected by a policy change.  
If however many thousand of things are affected by the change of a single policy (e.g. when "policy imports" are used and a policy is imported in many other policies), then the Ditto search service is overwhelmed with load and will eventually crash or be killed by k8s "liveness" probe not being able to pass.

This "selects" all affected things in the search index which shall be updated:
https://github.com/eclipse-ditto/ditto/blob/602abb417c7ab4d5382cbf5a2a0dfe71307b08af/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/impl/MongoThingsSearchUpdaterPersistence.java#L101-L109

This current implementation will however currently not throttle in any way - so a Stream of thousand of elements could be published and almost "at once" are sent to be updated in the search index.  

This issue suggests to provide a configuration to apply throttling at this place to be able to e.g. limit the output of the stream to "x search index updates per timeinterval", e.g. 100 per second.

hu-ahmed:
fix #2122 by adding throttling config

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -1377,7 +1377,7 @@ thingsSearch:
       # updaterPersistenceReadPreference configures the MongoDB read preference for the "ThingUpdater"
       updaterPersistenceReadPreference: "primaryPreferred"
       # PolicyModificationCausedSearchIndexUpdateThrottling contains throttling configuration for the search Index update after a policy update
-      PolicyModificationCausedSearchIndexUpdateThrottling:
+      policyModificationCausedSearchIndexUpdateThrottling:
         # enabled defines whether throttling should be applied for search Index update after a policy update.
         enabled: false
         # The time window within which the throttling limit applies.
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -1376,6 +1376,14 @@ thingsSearch:
       updaterPersistenceReadConcern: "local"
       # updaterPersistenceReadPreference configures the MongoDB read preference for the "ThingUpdater"
       updaterPersistenceReadPreference: "primaryPreferred"
+      # PolicyModificationCausedSearchIndexUpdateThrottling contains throttling configuration for the search Index update after a policy update
+      PolicyModificationCausedSearchIndexUpdateThrottling:
+        # enabled defines whether throttling should be applied for search Index update after a policy update.
+        enabled: false
+        # The time window within which the throttling limit applies.
+        interval: 1s
+        # The maximum number of updates allowed within each throttling interval.
+        limit: 100
     # updater contains configuration for the "Things Updater" of things-search service
     updater:
       # activityCheckInterval configures to keep thing updaters for that amount of time in memory when no update did happen:
--- a/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultSearchPersistenceConfig.java
+++ b/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultSearchPersistenceConfig.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 
 import javax.annotation.concurrent.Immutable;
 
+import org.eclipse.ditto.base.service.config.ThrottlingConfig;
 import org.eclipse.ditto.internal.utils.config.ConfigWithFallback;
 import org.eclipse.ditto.internal.utils.config.DittoConfigError;
 import org.eclipse.ditto.internal.utils.persistence.mongo.config.MongoDbConfig;
@@ -35,6 +36,7 @@ public final class DefaultSearchPersistenceConfig implements SearchPersistenceCo
 
     private final ReadPreference readPreference;
     private final ReadConcern readConcern;
+    private final ThrottlingConfig policyModificationCausedSearchIndexUpdateThrottling;
 
 
     private DefaultSearchPersistenceConfig(final ConfigWithFallback config) {
@@ -56,6 +58,7 @@ public final class DefaultSearchPersistenceConfig implements SearchPersistenceCo
                                     readConcernString);
                     return new DittoConfigError(msg);
                 });
+        policyModificationCausedSearchIndexUpdateThrottling = ThrottlingConfig.of(config);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultSearchPersistenceConfig.java
```diff
@@ -89,19 +97,20 @@ public boolean equals(final Object o) {
             return false;
         }
         final DefaultSearchPersistenceConfig that = (DefaultSearchPersistenceConfig) o;
-        return readPreference == that.readPreference && readConcern == that.readConcern;
+        return readPreference == that.readPreference && readConcern == that.readConcern && updaterThrottlingConfig == that.updaterThrottlingConfig;
```
thjaeckle, line 100: `updaterThrottlingConfig` cannot be compared with `==`

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/SearchPersistenceConfig.java
```diff
@@ -38,6 +39,14 @@ public interface SearchPersistenceConfig {
      */
     ReadConcern readConcern();
 
+    /**
+     * Returns the {@code ThrottlingConfig} for the search Index update.
+     *
+     * @return the Search Index ThrottlingConfig.
+     * @since 3.7.0
+     */
+    ThrottlingConfig getThrottlingConfig();
```
thjaeckle, line 48: I would name it more specific.  
Reasoning: We should make clear what is throttled.

It is about throttling after a policy update - so e.g. name it (also everywhere else, including the config) maybe: `getPolicyModificationCausedSearchIndexUpdateThrottling()`

thjaeckle (changes requested):
Awesome work @hu-ahmed  
Thanks a lot.

I would name the "throttling" config more specific.
Reasoning: We should make clear what is throttled. And we maybe need to add another throttling config for something else at the same config in the future.

It is about throttling after a policy update - so e.g. name it (also everywhere else, including the config) maybe: `getPolicyModificationCausedSearchIndexUpdateThrottling()`

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultSearchPersistenceConfig.java
```diff
@@ -35,6 +36,7 @@ public final class DefaultSearchPersistenceConfig implements SearchPersistenceCo
 
     private final ReadPreference readPreference;
     private final ReadConcern readConcern;
+    private final ThrottlingConfig updaterThrottlingConfig;
```
thjaeckle, line 39: I would also rename the field accordingly ..

deployment/helm/ditto/values.yaml
```diff
@@ -1376,6 +1376,13 @@ thingsSearch:
       updaterPersistenceReadConcern: "local"
       # updaterPersistenceReadPreference configures the MongoDB read preference for the "ThingUpdater"
       updaterPersistenceReadPreference: "primaryPreferred"
+      updaterPersistenceThrottling:
```
thjaeckle, line 1379: I would also rename the config accordingly to be more specific what is throttled.

And - the `updaterPersistenceThrottling` does not yet have a documentation line.

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/SearchPersistenceConfig.java
```diff
@@ -38,6 +39,14 @@ public interface SearchPersistenceConfig {
      */
     ReadConcern readConcern();
 
+    /**
+     * Returns the {@code ThrottlingConfig} for the policy update after search Index update.
```
thjaeckle, line 43:
```suggestion
     * Returns the {@code ThrottlingConfig} for the search Index update after a policy update.
```

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/impl/MongoThingsSearchUpdaterPersistence.java
```diff
@@ -58,8 +59,11 @@ public final class MongoThingsSearchUpdaterPersistence implements ThingsSearchUp
 
     private final MongoCollection<Document> collection;
 
+    private final ThrottlingConfig throttlingConfig;
```
thjaeckle, line 62: Please also rename field to be more specific

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/impl/MongoThingsSearchUpdaterPersistence.java
```diff
@@ -106,18 +110,26 @@ public Source<PolicyReferenceTag, NotUsed> getPolicyReferenceTags(final Map<Poli
...
-                    final Collection<PolicyId> referencedPolicyIds = referencedPolicyIds(doc);
-                    return referencedPolicyIds.stream()
-                            .map(referencedPolicyId -> Optional.ofNullable(policyRevisions.get(referencedPolicyId))
-                                    .map(revision -> PolicyTag.of(referencedPolicyId, revision))
-                                    .map(policyTag -> PolicyReferenceTag.of(thingId, policyTag))
-                                    .orElse(null))
-                            .filter(Objects::nonNull)
-                            .toList();
-                });
+        var source = Source.fromPublisher(publisher);
```
thjaeckle, line 113: I would rather do this not as mutable source - but make it final instead.
Meaning adding e.g. an `else` where you would only initialize `source = Source.fromPublisher(publisher)` - and when throttling enabled, do in addition the `.throttle()`

thingsearch/service/src/main/resources/search.conf
```diff
@@ -299,6 +299,20 @@ ditto {
         # read concern is one of: default, local, majority, linearizable, snapshot, available
         readConcern = ${ditto.mongodb.options.readConcern}
         readConcern = ${?UPDATER_PERSISTENCE_MONGO_DB_READ_CONCERN}
+
+        throttling {
+          enabled = false
+          enabled = ${?UPDATER_PERSISTENCE_THROTTLING_ENABLED}
```
thjaeckle, line 305: Same for the environment variables.. more specific what is throttled

thjaeckle (approved):
LGTM, nice 👍

PR #2126, opened 2025-02-18
update dependencies

thjaeckle:


alstanchev (approved):
👍

thjaeckle:
System tests passed (apart from known 12 `RestIntegrationIT` tests which fail in GH Actions but run locally currently):
https://github.com/eclipse-ditto/ditto/runs/37385135225

Issue #2084, opened 2025-01-03
Update to Rhino 1.8.0 for Ditto connectivity JavaScript payload mapper


thjaeckle:
See release notes: https://github.com/mozilla/rhino/releases/tag/Rhino1_8_0_Release

PR #2127, opened 2025-02-18
#2084: update to Rhino 1.8.0

thjaeckle:


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/runs/37451849508

PR #2130, opened 2025-02-24
add namespace to all Helm managed k8s resources

thjaeckle:


thjaeckle:
@kalinkostashki or @alstanchev could you have a look?
One of our toolings linting Helm charts reported that we did not use `namespace` - so I added them to all k8s resources in the chart.

deployment/helm/ditto/templates/connectivity-configmap.yaml
```diff
@@ -17,6 +17,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ $releaseName }}-service-config-connectivity-extension-conf
+  namespace: {{ $root.Release.Namespace }}
```
kalinkostashki, line 20: what is the difference between configmaps and other resources that requires the use of $root?
I guess this question is valid for all occurences :)

kalinkostashki (changes requested):
not clear on the use of $root instead of .Release.Namespace

deployment/helm/ditto/templates/connectivity-configmap.yaml
thjaeckle: I just used everywhere where a variable `$root` was defined this variable instead of `.`

It is only necessary for places where we e.g. iterate over something - in that case `.` is no longer the "root", but is scoped to the current element being iterated over.

kalinkostashki (approved):
LGTM :+1:

PR #2135, opened 2025-03-03
fix migrateExistingThing test using mock

hu-ahmed:


thjaeckle (approved):
LGTM, thanks a lot for the quick fix 👍

PR #2136, opened 2025-03-03
fix MigrateThingDefinition wot validation issue

hu-ahmed:


thjaeckle:
@hu-ahmed nice catch .. this is definetly a problem
I however wonder if we should just remove the overwritten `onStagedQuery` method, as its parent already does the job correctly:
https://github.com/eclipse-ditto/ditto/blob/0c80cf336e3347791fccdfba013c7b8435a2e725/internal/utils/persistent-actors/src/main/java/org/eclipse/ditto/internal/utils/persistentactors/AbstractPersistenceActor.java#L815-L834

thjaeckle (approved):
LGTM 👍

thjaeckle (approved):

PR #2137, opened 2025-03-04
fix pre-defined extraFields were causing missing extra from Thing created events

thjaeckle:
With the in Ditto 3.7.0 added "pre-defined" extraFields configuration `ThingCreated` events did not contain the configured "extra" fields if pre-defined extra-fields were configured.

The reason was that the Policy is not yet available during creation of the Thing and thus, no fields were added to `extra` as authorization was missing.

This PR provides a fix which solves this by doing a roundtrip and extra-fields retrieval in such (rare) cases of thing creation.  
It also improves how it is calculated if all asked for "extra" fields were available as pre-configured fields.


thjaeckle:
Should only affect users of the new functionality, configuring pre-defined extraFields

alstanchev (approved):
LGTM 👍

PR #2139, opened 2025-03-04
fix http-push-connection-dispatcher config overrides

### LINKED ISSUE #2138: http push connection issues with slow target endpoints
Http connection scaling against slow responding endpoints is failing and leading to dropped messages.

alstanchev:
properties were not in the right place and the config was defaulting to https://github.com/apache/pekko/blob/2469f729f7503acf814bcbd042b4bb0863103c9d/actor/src/main/resources/reference.conf#L506

Also setting the defaults to some more sensible values so that the threadpool is created with a reasonable size and only grow on demand.

fixes #2138

### DIFF
--- a/connectivity/service/src/main/resources/connectivity.conf
+++ b/connectivity/service/src/main/resources/connectivity.conf
@@ -1331,15 +1331,28 @@ http-push-connection-dispatcher {
   # The core-pool-size-min remains unchanged so its quite small (8). Unused threads will expire after 60s.
 
   thread-pool-executor {
+    # The core-pool-size-factor is used to determine corePoolSize of the
+    # ThreadPoolExecutor using the following formula:
+    # ceil(available processors * factor).
+    # Resulting size is then bounded by the core-pool-size-min and
+    # core-pool-size-max values.
+    # Increasing can help aleviate errors of type "Dropped message as result of backpressure strategy!" during high load.
     core-pool-size-factor = 4
     core-pool-size-factor = ${?HTTP_PUSH_CORE_POOL_SIZE_FACTOR}
-    core-pool-size-max = 64,
+    # Max number of threads to cap factor-based corePoolSize number to
+    core-pool-size-max = 64
     core-pool-size-max = ${?HTTP_PUSH_CORE_POOL_SIZE_MAX}
-    max-pool-size-min= 64,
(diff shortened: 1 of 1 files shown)


connectivity/service/src/main/resources/connectivity.conf
```diff
@@ -1329,11 +1329,18 @@ http-push-connection-dispatcher {
...
-  core-pool-size-factor = 2147483647
-  core-pool-size-factor = ${?HTTP_PUSH_CORE_POOL_SIZE_FACTOR}
 
-  core-pool-size-max = 2147483647
-  core-pool-size-max = ${?HTTP_PUSH_CORE_POOL_SIZE_MAX}
+  thread-pool-executor {
+    core-pool-size-factor = 4
+    core-pool-size-factor = ${?HTTP_PUSH_CORE_POOL_SIZE_FACTOR}
+    core-pool-size-max = 64,
+    core-pool-size-max = ${?HTTP_PUSH_CORE_POOL_SIZE_MAX}
```
thjaeckle, line 1337: You adjusted the defaults.
The prior defaults seemed not logical to me .. but did you test what effect those configs have with e.g. different load?

Might be worth to add a little documentation for those configurations.

alstanchev: Prior defaults were not working at all as those properties should be inside "thread-pool-executor" object. If putted in such a threadpool with int.max number of threads was being created. 
I did tested with different load and was able to publish more than ~200 messages to an endpoint that responds after 1.5s only by changing the dispatcher settings and parallelism of the connection. Before changing the dispatcher config there were "There are too many in-flight requests." (handled by parallelism) and "Dropped message as result of backpressure strategy!" (handled by tweaking the dispatcher). Maybe the default factor can be left at 3 but i don't see a drawback if set to 4. Nevertheless i don't have a strong opinion on the default value as i will be overriding those later.

thjaeckle: Ok. 
If you already know you want/need to overwrite them, you could give a hint in the config documentation about how  and when to tweak them (e.g. when inspecting certain dispatcher metrics).

But I assume this is highly dependent on the use-case :D

thjaeckle (approved):

PR #2140, opened 2025-03-04
fix kamon.conf to be loaded in config

alstanchev:
Seems like kamon.conf was never used.


internal/utils/config/src/main/resources/ditto-kamon.conf
```diff
@@ -48,7 +48,12 @@ kamon {
         0.5,
         0.75,
         1,
+        1.25,
```
thjaeckle, line 51: This will give more exact histrogram metrics, right?
And the "cost" is probably more Prometheus resources needed to store those?

alstanchev: Yes i expect so but it shouldn't have much of an impact as time buckets in the context of ditto metrics are mostly used in the ms range and those would just give some more visibility if something goes in the low seconds. At least that is my assumption.

thjaeckle (approved):
LGTM - strange that we did not notice earlier :D

PR #2150, opened 2025-03-28
CVE-2025-1974: add allow-snippet-annotations: true in nginx-ingress c…

vvasilevbosch:
…onfig, since it's disabled by default, and we are using it to configure CORS on ingress

https://github.com/kubernetes/kubernetes/issues/131009


kalinkostashki (changes requested):
This would be valid if nginx ingress controller version is upgraded.
Shouldn't you do this as well?

kalinkostashki (approved):
LGTM :+1:

thjaeckle (approved):

Issue #2148, opened 2025-03-25
Wot Validation fails when updating primitive values due to full-object validation context


hu-ahmed:
When updating a primitive property (e.g., number, string, boolean) in a feature property structure — using the Ditto HTTP API — a valid primitive payload is incorrectly rejected with a validation error, even though the value matches the expected type in the WoT model.

Example:

`PUT /api/2/things/${thingId}/features/temperatureControl/properties/configuration/targetTemperature/value` 
Payload: 22.5
Model:
In the WoT model, targetTemperature.value is of type number.

Response:

```
{
  "status": 400,
  "error": "wot:payload.validation.error",
  "message": "The provided payload did not conform to the specified WoT (Web of Things) model.",
  "validationDetails": {
    "/features/temperatureControl/properties/configuration/targetTemperature/value": []
  }
}
```

**Expected Behavior:**

Primitive values (like 22.5 or 20) that match the expected type in the WoT model should be accepted and pass validation.

**Observed Behavior:**

The primitive value is internally wrapped into a synthetic JSON structure (e.g., { "value": 22.5 }) for validation.

This synthetic object may miss required sibling fields defined at the parent level in the WoT model.

Validation fails because the entire object is validated instead of just the targeted property.

Later, when validationDetails are built, the failed path does not map back to the updated leaf node, resulting in an empty array, which is misleading and unhelpful for debugging.

**Impact:**

Even valid primitive value updates are incorrectly rejected.

The error message is misleading and lacks proper detail.

**Suggestion:**

Validation logic should consider the context of the update path and validate only the relevant part of the structure.

If synthetic wrapping is used for validation, Ditto should either include the necessary required fields or limit the validation scope to the updated node.

validationDetails should accurately reflect the affected paths, even in partial updates or primitive leaf-value updates.

thjaeckle:
Thanks @hu-ahmed - very detailed and comprehensible bug description.
We should definitely tackle this for the next bugfix release 👍

PR #2152, opened 2025-03-29
#2148: fix primitive values validation issue

hu-ahmed:
Fixes: #2148


thjaeckle (approved):
LGTM, thanks for the report and the quick fix 👍

PR #2157, opened 2025-04-02
fix WoT based validation rejecting message API calls if content-type was not application/json

thjaeckle:
* and in case `log-warning-instead-of-failing-api-calls` is configured to be `true


alstanchev (commented):
LGTM 👍

alstanchev (approved):

PR #2158, opened 2025-04-08
stabilize pre-defined extraFields enrichment, treating occurring exceptions by not pre-enriching

thjaeckle:
* also use Patterns.pipe on completionstage for result


alstanchev (approved):
👍

PR #2165, opened 2025-05-05
Fix several completion stages to use async execution and other performance improvements

thjaeckle:
* For WoT validation, specifying the "wot-dispatcher" Executor to use
* For policy enforcement, specifying the "enforcement-dispatcher" Executor to use

Classifying as "bug" as we already have dedicated dispatchers / executors configured which shall be used to perform those async operations, but most of the time they are not used.
Instead, they use the "commonPool" JoinForkPool - which has a limited amount of threads and is used for "everything else" not explicitly defining an executor to run on.

This can be problematic, as under load e.g. WoT validation might influence other async tasks of the things service. Or the other way around as well - without having a proper isolation or options to "tweak".

### DIFF
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
@@ -143,7 +143,7 @@ public final class ThingEnforcerActor
 
         final DittoWotIntegration wotIntegration = DittoWotIntegration.get(system);
         thingModelValidator = wotIntegration.getWotThingModelValidator();
-        wotValidationExecutor = getContext().getSystem().dispatchers().lookup("wot-dispatcher");
+        wotValidationExecutor = getContext().getSystem().dispatchers().lookup(DittoWotIntegration.WOT_DISPATCHER);
     }
 
     /**
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
@@ -19,6 +19,7 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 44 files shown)


thjaeckle:
@alstanchev @kalinkostashki could you have a look please?
I did some profiling on out prod instance and found out that e.g. the configured dispatchers for WoT validation, but also for policy enforcement were not used as I would have expected ..  
A lot of async CompletionStage executions were done on the "commonPool" (which e.g. in its size is dependent on the amount of CPU cores of the OS).

I also fixed some other performance issues found during profiling, e.g. in `ditto-json` or in the DittoHeader parsing.

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
```diff
@@ -143,6 +143,7 @@ private ThingEnforcerActor(final ThingId thingId,
 
         final DittoWotIntegration wotIntegration = DittoWotIntegration.get(system);
         thingModelValidator = wotIntegration.getWotThingModelValidator();
+        wotValidationExecutor = getContext().getSystem().dispatchers().lookup("wot-dispatcher");
```
alstanchev, line 146: The dispatcher name here is duplicated also in the DefaultDittoWotIntegration. Maybe it should be a constant like the "enforcement-dispatcher".I don't have a strong opinion on that just pointing it if it is a miss. The DefaultDittoWotIntegration is not public and maybe you left it this way on purpose.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/AbstractThingModifyCommandStrategy.java
```diff
@@ -35,8 +36,11 @@
 abstract class AbstractThingModifyCommandStrategy<C extends ThingModifyCommand<C>>
         extends AbstractThingCommandStrategy<C> {
 
+    protected final Executor wotValidationExecutor;
+
     protected AbstractThingModifyCommandStrategy(final Class<C> theMatchingClass, final ActorSystem actorSystem) {
         super(theMatchingClass, actorSystem);
+        wotValidationExecutor = actorSystem.dispatchers().lookup("wot-dispatcher");
```
alstanchev, line 43: Also here the dispatcher name a string.

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingEnforcerActor.java
thjaeckle: @alstanchev agreed .. I now extracted the 2 wot dispatchers as constants to the public `DittoWotIntegration` interface.

PR #2167, opened 2025-05-13
added configurations ingress-nginx

kalinkostashki:
- workerProcesses, workerConnections, workerOpenFiles have been added.
 - this helps to avoid to use the default "auto" value of workerProcesses which may cause very high RAM usage if you have a lot of CPUs in place

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -302,6 +302,18 @@ ingress:
     enabled: false
     # replicaCount configuration for the ingress controller
     replicaCount: 1
+    # Exposes the ingress controller settings in order to configure its behavior.
+    config:
+      # Sets the number of worker processes that should be started by the ingress controller.
+      # Depending on the load of the ingress controller, this value can be increased or decreased.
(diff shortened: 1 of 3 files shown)


deployment/helm/ditto/values.yaml
```diff
@@ -302,6 +302,14 @@ ingress:
     enabled: false
     # replicaCount configuration for the ingress controller
     replicaCount: 1
+    config:
```
thjaeckle, line 305: could you add a documentation line before, please?

deployment/helm/ditto/values.yaml
```diff
@@ -302,6 +302,14 @@ ingress:
...
+      workerProcesses: 2
```
thjaeckle, line 306: could you add a documentation line before, please?

thjaeckle (approved):
LGTM in general 👍 
Could you however add also a doc line for the 2 mentioned places?

deployment/helm/ditto/values.yaml
kalinkostashki: added necessary explanation

PR #2169, opened 2025-05-13
Fixed setting tag of hook image in Helm chart

kyberpunk:
Hi. In the previous PR I did a mistake and mixed up order of parameters. So, the tag is always "latest" and cannot be overwritten. Fixing it in this PR, thanks.


kalinkostashki (approved):
Looks good :)

thjaeckle (approved):
Thx 👍

PR #2178, opened 2025-05-30
fix negative from/to for historical streaming confusion

thjaeckle:
* the current behavior was instead index/offset semantics, which provides the wrong semantics for the documented options


alstanchev:
LGTM 👍

PR #2181, opened 2025-06-04
fix "thing definition migrated" event signal enrichment caching

thjaeckle:
There is a bug which leads to enriched / `extra` fields no being merged correctly by the `DittoCachingSignalEnrichmentFacade`.  
This lead to missing fields from "extra" in emitted events (e.g. via Kafka).


thjaeckle:

alstanchev (approved):
LGTM 👍

PR #2187, opened 2025-06-12
increased pekko version to 1.2.0-M1

kalinkostashki:
- this is done so a 3.8.0-M1 release can be built
 - added propagate-harmless-quarantine-events=off as a potential fix to https://github.com/apache/pekko/issues/578

### DIFF
--- a/internal/utils/config/src/main/resources/ditto-pekko-config.conf
+++ b/internal/utils/config/src/main/resources/ditto-pekko-config.conf
@@ -177,6 +177,14 @@ pekko {
         port = ${?BIND_REMOTE_PORT}
       }
 
+      # If set to "on", InboundQuarantineCheck will propagate harmless quarantine events.
+      # This is the legacy behavior. Users who see these harmless quarantine events lead
+      # to problems can set this to "off" to suppress them (https://github.com/apache/pekko/pull/1555).
+      # This issue should fix https://github.com/apache/pekko/issues/578 where a cluster under load undergoing rollout restart
+      # or helm upgrade procedure interprets harmless quarantine events as DownSelfQuarantinedByRemote which drags down
+      # and downs the neighbours of that particular node.
+      propagate-harmless-quarantine-events = off
+
       advanced {
         # Maximum serialized message size, including header data. # default: 256 KiB
         maximum-frame-size = 256 KiB
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -44,11 +44,11 @@
         <lz4-java.version>1.8.0</lz4-java.version>
 
         <pekko-bom.version>1.2.0-M1</pekko-bom.version>
-        <pekko-http-bom.version>1.1.0</pekko-http-bom.version>
+        <pekko-http-bom.version>1.2.0</pekko-http-bom.version>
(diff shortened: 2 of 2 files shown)


thjaeckle:
@kalinkostashki the tests fail with
```
CborJsonValueSerializerTest.beforeClass:48 » NoSuchMethod 'com.fasterxml.jackson.core.StreamReadConstraints$Builder com.fasterxml.jackson.core.StreamReadConstraints$Builder.maxTokenCount(long)'
```

I assume some inconsistencies about Jackson - which Pekko 1.2.0 probably updated.

internal/utils/config/src/main/resources/ditto-pekko-config.conf
```diff
@@ -222,6 +222,14 @@ pekko {
...
         idle-cpu-level = ${?REMOTING_IDLE_CPU_LEVEL}
       }
+
+      # If set to "on", InboundQuarantineCheck will propagate harmless quarantine events.
+      # This is the legacy behavior. Users who see these harmless quarantine events lead
+      # to problems can set this to "off" to suppress them (https://github.com/apache/pekko/pull/1555).
+      # This issue should fix https://github.com/apache/pekko/issues/578 where a cluster under load undergoing rollout restart
+      # or helm upgrade procedure interprets harmless quarantine events as DownSelfQuarantinedByRemote which drags down
+      # and downs the neighbours of that particular node.
+      propagate-harmless-quarantine-events = off
```
thjaeckle: @kalinkostashki maybe we should move that above the `advanced {` block - as it also is that way in the default pekko config.

kalinkostashki: I've actually looked at the reference.conf file and it seems to be under artery directly:
image

thjaeckle: Yes, true ...
However it is "before" the `advanced {` block .. I just want it to be at the same place as in the reference config.   
It work also like it is now, of course .. only cosmetic change :D

kalinkostashki: sorry and problem.
I just spaced out and didn't read the **above** part of your comment :laughing: 
fixed ;)

kalinkostashki:
@thjaeckle does it make sense to push out all available and used pekko dependencies to their latest version if we're going to be building a milestone release anyway?

thjaeckle:
@kalinkostashki if other pekko dependencies also already are available in a newer milestone, I would say yes ..

Then we can test and even give feedback before a final release

kalinkostashki:
> @kalinkostashki let me just run the SystemTests on this branch before merge ..

I already asked @alstanchev to do this :)

thjaeckle:
Oh, you're right .. here are the results of that first run:
https://github.com/eclipse-ditto/ditto/runs/44030601528

10 failed in `RestConnectionsIT` - I think this test sometimes was very unstable in the past as well .. let's see what run 2 does ..

thjaeckle:
Cool - next run was green: https://github.com/eclipse-ditto/ditto/actions/runs/15629645828

PR #2190, opened 2025-06-19
dynamic response diversion between connections

alstanchev:
Adds response diversion functionality to route command responses to different connections at runtime. Addresses use cases where responses need to go to systems other than the original sender.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
@@ -222,6 +222,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
     REPLY_TARGET("ditto-reply-target", Integer.class, false, false, HeaderValueValidators.getIntValidator()),
 
     /**
+     * Header mapping key for the target connection ID to divert responses to.
+     * When configured in a source's headerMapping, responses from that source will be diverted
+     * to the specified connection instead of being sent to the configured replyTarget.
+     * The value should be a valid connection ID string or header placeholder.
+     *
+     * @since 3.8.0
+     */
+    DITTO_DIVERT_RESPONSE_TO("ditto-divert-response-to",
+            String.class,
+            true,
+            false,
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
+     * Header mapping key for specifying which response types should be diverted.
+     * Value should be a comma-separated list of response types: "response,error,nack".
+     * If not specified, all response types will be diverted.
+     *
+     * @since 3.8.0
+     */
+    DITTO_DIVERT_EXPECTED_RESPONSE_TYPES("ditto-divert-expected-response-types",
+            String.class,
+            true,
+            false,
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
+     * Header key for tracking the source connection of a diverted response.
+     *
+     * @since 3.8.0
+     */
+
+    DITTO_DIVERTED_RESPONSE_FROM("ditto-diverted-response-from",
+            String.class,
+            true,
+            true,
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
      * Header definition for the internal header "ditto-expected-response-types".
      * This header is evaluated to distinguish if a response should published or not.
      *
@@ -564,7 +603,7 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
 
     /**
      * Internal header containing the pre-defined configured {@code extraFields} as keys and the allowed "read subjects"
-     * as array of stings - defining which "auth subjects" are allowed to read which pre-defined extra field.
+     * as array of strings - defining which "auth subjects" are allowed to read which pre-defined extra field.
      *
      * @since 3.7.0
      */
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
@@ -77,6 +77,7 @@ import org.eclipse.ditto.base.model.acks.PubSubTerminatedException;
 import org.eclipse.ditto.base.model.entity.id.EntityId;
 import org.eclipse.ditto.base.model.entity.id.WithEntityId;
 import org.eclipse.ditto.base.model.exceptions.DittoRuntimeException;
+import org.eclipse.ditto.base.model.headers.DittoHeaderDefinition;
 import org.eclipse.ditto.base.model.headers.DittoHeaders;
 import org.eclipse.ditto.base.model.headers.WithDittoHeaders;
 import org.eclipse.ditto.base.model.signals.Signal;
@@ -86,6 +87,7 @@ import org.eclipse.ditto.base.model.signals.commands.streaming.StreamingSubscrip
 import org.eclipse.ditto.connectivity.api.BaseClientState;
 import org.eclipse.ditto.connectivity.api.InboundSignal;
 import org.eclipse.ditto.connectivity.api.OutboundSignal;
+import org.eclipse.ditto.connectivity.api.OutboundSignalFactory;
 import org.eclipse.ditto.connectivity.api.messaging.monitoring.logs.LogEntryFactory;
 import org.eclipse.ditto.connectivity.model.Connection;
 import org.eclipse.ditto.connectivity.model.ConnectionId;
@@ -312,7 +314,13 @@ public abstract class BaseClientActor extends AbstractFSMWithStash<BaseClientSta
         registerConnectionPubSub();
         addCoordinatedShutdownTasks();
 
+        registerDiversionPubSub();
+
         init();
+    }
+
+    private boolean isDiversionTarget() {
+        return Boolean.parseBoolean(connection.getSpecificConfig().getOrDefault("is-diversion-target", "false"));
     }
 
     protected ConnectivityConfig connectivityConfig() {
@@ -356,6 +364,25 @@ public abstract class BaseClientActor extends AbstractFSMWithStash<BaseClientSta
             }
         });
 
+    }
+
+    private void registerDiversionPubSub() {
+        if (dryRun || !isDiversionTarget()) {
+            // no need to modify distributed data if it is a dry run
+            return;
+        }
+        final ConnectionId connectionId = connectionId();
+        // Subscribe to receive diverted responses if this connection is configured as a target
+        // Note: Subscription happens in preStart but the actor might not be in CONNECTED state yet
+        logger.info("Connection {} is configured as diversion target, subscribing to diverted responses",
+                connectionId);
+        connectionPubSub.subscribeForDivertedResponses(connectionId, getSelf(), false).whenComplete((consistent, error) -> {
+            if (error != null) {
+                logger.error(error, "Failed to register client actor of connection <{}> for diverted responses", connectionId);
+            } else {
+                logger.info("Client actor registered for diverted responses: <{}>", connectionId);
+            }
+        });
     }
 
     private void addCoordinatedShutdownTasks() {
@@ -524,6 +551,7 @@ public abstract class BaseClientActor extends AbstractFSMWithStash<BaseClientSta
                 .event(InboundSignal.class, this::handleInboundSignal)
                 .event(PublishMappedMessage.class, this::publishMappedMessage)
                 .event(ConnectivityCommand.class, this::onUnknownEvent) // relevant connectivity commands were handled
+                .event(Signal.class, this::isDivertedResponse, this::handleDivertedResponse)
                 .event(Signal.class, this::handleSignal)
                 .event(FatalPubSubException.class, this::failConnectionDueToPubSubException)
                 .eventEquals(Control.ACKREGATOR_STARTED, this::ackregatorStarted)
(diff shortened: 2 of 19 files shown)


alstanchev:
TODO:

- [ ] System tests (in progress)
- [ ] Few TODOs to be discused

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -221,6 +221,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
 
+    /**
+     * Header mapping key for the target connection ID to divert responses to.
+     * When configured in a source's headerMapping, responses from that source will be diverted
+     * to the specified connection instead of being sent to the configured replyTarget.
+     * The value should be a valid connection ID string or header placeholder.
+     *
+     * @since 3.8.0
+     */
+    DITTO_DIVERT_RESPONSE_TO("divert-response-to",
```
thjaeckle: Should we maybe add `-connection` - so `"divert-response-to-connection"`?  
And rename the enum to `DIVERT_RESPONSE_TO_CONNECTION`?  
This only becomes obvious from the javadoc

thjaeckle: Only Ditto internal enum values start with `DITTO_` - so this one should not, because it is marked as `readFromExternalHeaders`

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -221,6 +221,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
+     * Header mapping key for specifying which response types should be diverted.
+     * Value should be a comma-separated list of response types: "response,error,nack".
+     * If not specified, all response types will be diverted.
+     *
+     * @since 3.8.0
+     */
+    DITTO_DIVERT_EXPECTED_RESPONSE_TYPES("divert-expected-response-types",
```
thjaeckle: Only Ditto internal enum values start with `DITTO_` - so this one should not, because it is marked as `readFromExternalHeaders`

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -221,6 +221,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+            false,
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
+     * Header key for tracking the source connection of a diverted response.
+     *
+     * @since 3.8.0
+     */
+
+    DITTO_DIVERTED_RESPONSE_FROM("diverted-response-from",
```
thjaeckle: Only Ditto internal enum values start with `DITTO_` - so this one should not, because it is marked as `readFromExternalHeaders`

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -221,6 +221,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+            String.class,
+            true,
```
thjaeckle, line 258: I'm not sure if this makes sense to let this be read from external headers.  
I don't think so .. what was the reasoning of marking it as such?

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -221,6 +221,45 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+            false,
+            HeaderValueValidators.getNonEmptyValidator()),
+
+    /**
+     * Header key for tracking the source connection of a diverted response.
+     *
+     * @since 3.8.0
+     */
+
```
thjaeckle: Empty line should be removed

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
```diff
@@ -312,9 +314,15 @@ public void preStart() throws Exception {
         registerConnectionPubSub();
         addCoordinatedShutdownTasks();
 
+        registerDiversionPubSub();
+
         init();
     }
 
+    private boolean isDiversionTarget() {
+        return Boolean.parseBoolean(connection.getSpecificConfig().getOrDefault("is-diversion-target", "false"));
```
thjaeckle: "is-diversion-target" and also the default "false" should both be (private) constants to this class

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
```diff
@@ -550,6 +578,54 @@ protected static String escapeActorName(final String name) {
         return URLEncoder.encode(name, StandardCharsets.US_ASCII);
     }
 
+    private boolean isDivertedResponse(final Signal<?> signal, final BaseClientData data) {
+        return !signal.getDittoHeaders().get(DittoHeaderDefinition.ORIGIN.getKey()).equals(connectionId().toString());
```
thjaeckle: IMO this would only make sense for responses, right?  
I fear that now also Events which e.g. "originated" from commands (e.g. PUT to a thing) via HTTP API would also match this condition.

So I would only do that for `CommandResponse` signals. Maybe even already in the `.event(...` matching calling this `isDivertedResponse` as predicate

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
thjaeckle: Maybe the string should also end with `-connection`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ResponseDiversionValidator.java
```diff
@@ -0,0 +1,124 @@
...
+@Immutable
+public final class ResponseDiversionValidator {
+
+    private static Map<String, String> mappings;
+
+    private ResponseDiversionValidator() {
+        // no instantiation
+    }
+
+    public static void validate(final Connection connection, final DittoHeaders dittoHeaders) {
```
thjaeckle: Missing javadoc

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/util/ConnectionPubSub.java
```diff
@@ -69,6 +69,25 @@ public void publishSignal(final Signal<?> signal, final ConnectionId connectionI
...
+     *
+     * @param signal the signal.
+     * @param connectionId the connection ID.
+     * @param groupIndexKey the group index key to choose the responsible client actor, either the entity ID or the
+     * subscription ID prefix.
+     * @param sender the sender of the signal.
+     */
+    public void publishSignalForDiversion(final Signal<?> signal, final ConnectionId connectionId, final CharSequence groupIndexKey,
+            @Nullable final ActorRef sender) {
+        final String topic = CONNECTION_RESPONSE_DIVERSION_PREFIX + connectionId.toString();
```
thjaeckle: Kind of misleading .. this is not really a `topic` in the sense of pub/sub, or is it?

It is "just" used as `connection-id` header .. maybe this is a copy&paste error?

documentation/src/main/resources/_posts/2025-10-09-response-diversion.md
```diff
@@ -0,0 +1,187 @@
...
+With response diversion, Eclipse Ditto becomes even more versatile in bridging different IoT protocols and systems, 
+enabling complex routing scenarios that were previously challenging or impossible to achieve.
+
+## The challenge: Multi-protocol IoT landscapes
+
+Modern IoT deployments often involve multiple protocols and systems working together. Consider these common scenarios:
+
+- **Cloud integration**: Your devices use MQTT to communicate with AWS IoT Core, but your analytics pipeline consumes data via Kafka
+- **Protocol translation**: Legacy systems expect HTTP webhooks, but your devices communicate via AMQP
+- **Response aggregation**: You want to collect all device responses in a central monitoring system regardless of the original protocol
```
thjaeckle, line 25: @alstanchev so would it also support "diverting" responses to e.g. a "monitoring" connection **additionally**? Leaving the "normal" response flow in-tact via the source connection?

I kind of understand this point that way ..

That would be as well a nice feature

thjaeckle (changes requested):
@alstanchev I had a first brief look and left some feedback .. but did not yet "dive into" the pub/sub stuff too much :D

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
alstanchev: The idea of adding the headers to the enum was to utilize the existing mechanisms  of preventing reading from external. But then i was surprised by the fact that the headers were removed after being added by the header mapper. So i changed to read from external in order to be at least working and left it like that with the idea for discussion during the review.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ResponseDiversionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+        validateResponseDiversionTypes(connection, dittoHeaders);
+        validateResponseDiversionConnectionId(connection, dittoHeaders);
+        validateAuthorizedDiversionSources(connection, dittoHeaders);
+    }
+
+    private static void validateAuthorizedDiversionSources(final Connection connection,
+            final DittoHeaders dittoHeaders) {
+        boolean isDiversionTarget = check(connection, IS_DIVERSION_TARGET, IS_DIVERSION_TARGET_DEFAULT, dittoHeaders);
+        if (isDiversionTarget) {
+            final Set<ConnectionId> authorizedSources = connection.getSpecificConfig()
```
thjaeckle: `authorizedSources` is not used - can this be removed?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ResponseDiversionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+            final DittoHeaders dittoHeaders) {
+        final String isSourseString = connection.getSpecificConfig().getOrDefault(configKey, defaultValue);
+        return Boolean.parseBoolean(isSourseString);
+    }
+
+    private static void validateResponseDiversionTypes(final Connection connection, final DittoHeaders dittoHeaders) {
+        final boolean isDiversionSource =
+                check(connection, IS_DIVERSION_SOURCE, IS_DIVERSION_SOURCE_DEFAULT, dittoHeaders);
+        if (isDiversionSource) {
+            final List<Source> list = connection.getSources().stream()
```
thjaeckle: `list` is not used .. can this be removed or does it serve a purpose?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/ResponseDiversionInterceptor.java
```diff
@@ -0,0 +1,240 @@
...
+ * 3. Publishing signals to the target connection via PubSub
+ * 4. Preventing circular diversion chains
+ */
+@Immutable
+public class ResponseDiversionInterceptor {
+
+    private static final DittoLogger LOGGER = DittoLoggerFactory.getLogger(ResponseDiversionInterceptor.class);
+    private static final Set<String> DEFAULT_RESPONSE_TYPES =
+            Set.of(ResponseType.RESPONSE.getName(), ResponseType.ERROR.getName());
+    public static final String AUTHORIZED_CONNECTIONS_AS_SOURCES = "authorized-connections-as-sources";
```
thjaeckle: This `authorized-connections-as-sources` is not documented .. it only is listed in the blogpost.

It would be good to:
* add Javadoc to the public variable
* add documentation in the markdown docs for the website .. it is mentioned in the blogpost, but this is not really documentation

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ResponseDiversionValidator.java
alstanchev: Indeed the result is not needed, will refactor to be more clear that only calling of ConnectionId.of is the intention to validate the connectionId strings

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ResponseDiversionValidator.java
alstanchev: Here the same as the previous the list was used as terminal operation to remove IDE warnings.

documentation/src/main/resources/_posts/2025-10-09-response-diversion.md
alstanchev: This was not my initial intention but i have added that functionality with the latest commit.

alstanchev:
Hi @alstanchev, yes i saw them but was dealing with other issues. I will see why, fix them and wright back as soon as possible

alstanchev:
@thjaeckle fixed the tests. Currently there are failing RestConnectionsIT but they also fail on master. There seems to be some issue with thr JWT configuration.

documentation/src/main/resources/pages/ditto/connectivity-response-diversion.md
```diff
@@ -1,55 +1,80 @@
...
+Response diversion is configured at the connection source and target level
+using header mapping and specific config.
 The following special headers control the diversion behavior:
 
-### Header mapping keys
+### Source connection configuration
+
+#### Header mapping keys
 
 #### divert-response-to-connection
```
thjaeckle: If "Header mapping keys" is level 4 heading, this (and the other header keys as well) must be a level 5

PR #2191, opened 2025-06-23
enhance Helm chart with signal enrichment cache config and feature flags config

thjaeckle:


thjaeckle:
@alstanchev @kalinkostashki could you please check + approve this PR?

deployment/helm/ditto/values.yaml
```diff
@@ -131,6 +131,19 @@ global:
...
+    mergeThingsEnabled: true
+    # wotIntegrationEnabled controls whether the Web of Things (WoT) integration is enabled or not
+    wotIntegrationEnabled: true
+    # historicalApisEnabled controls whether the historical APIs (using the event journal) are enabled or not
+    historicalApisEnabled: true
+    # preserveKnownMqttHeadersEnabled controls whether known MQTT headers (e.g., mqtt.topic) are preserved in outgoing messages
+    preserveKnownMqttHeadersEnabled: true
+    # jsonKeyValidationEnabled controls whether for each json object, each JSON key should be validated against a regex pattern
+    #  excluding control characters to be used as JSON keys. This can have quite a performance impact.
+    jsonKeyValidationEnabled: true
```
kalinkostashki, line 146: shouldn't be added to the things-deployment.yaml as well since jsonKeyValidation is used in ThingsModelFactory?

deployment/helm/ditto/values.yaml
```diff
@@ -131,6 +131,19 @@ global:
...
     # port where prometheus metrics will be provided
     port: 9095
+  # featureFlags contains feature flags enabling/disabling certain features
+  featureFlags:
+    # mergeThingsEnabled controls whether MergeThing commands (patching of things) are enabled or not
+    mergeThingsEnabled: true
+    # wotIntegrationEnabled controls whether the Web of Things (WoT) integration is enabled or not
+    wotIntegrationEnabled: true
+    # historicalApisEnabled controls whether the historical APIs (using the event journal) are enabled or not
+    historicalApisEnabled: true
```
kalinkostashki, line 141: this should be enabled for thins-deployment.yaml as well I believe.

deployment/helm/ditto/values.yaml
```diff
@@ -1674,6 +1687,16 @@ connectivity:
...
         # expireAfterAccess prolonged on each cache access by that duration
         expireAfterAccess: 2h
+    # signalEnrichment contains configuration regarding signal enrichment (retrieval of "extra" fields to e.g. enrich events)
+    signalEnrichment:
+      # cache holds the configuration of signal enrichment caching
+      cache:
+        # enabled whether signal enrichment caching (incl. smart updates updating the cache based on event data) should be enabled
+        enabled: true
+        # maximumSize how many things to cache in total on a single cluster node
+        maximumSize: 2000
```
kalinkostashki, line 1702: since this is the number of things and not a buffer shouldn't we increase this to match the 20 000 as configured in the connectivity.conf:
maximum-size = 20000
Or do we want it 2000... since 20 000 * 100kb would be 2 GB worth of things.
So maybe it makes more sense for a barebones default install to be 2000 instead of 20 000 :)

deployment/helm/ditto/values.yaml
```diff
@@ -2045,6 +2068,16 @@ gateway:
...
         #  throttling interval e.g. 100 msgs/s
         limit: 100
+    # signalEnrichment contains configuration regarding signal enrichment (retrieval of "extra" fields to e.g. enrich events)
+    signalEnrichment:
+      # cache holds the configuration of signal enrichment caching
+      cache:
+        # enabled whether signal enrichment caching (incl. smart updates updating the cache based on event data) should be enabled
+        enabled: true
+        # maximumSize how many things to cache in total on a single cluster node
+        maximumSize: 2000
```
kalinkostashki, line 2083: same comment as for the maximum size since gateway.conf has it on 20 000.

kalinkostashki (changes requested):
Some small comments :)

deployment/helm/ditto/values.yaml
thjaeckle: @kalinkostashki good catch .. that was not on purpose .. I seem to have forgotten it :/

deployment/helm/ditto/values.yaml
thjaeckle: yes 👍

deployment/helm/ditto/values.yaml
thjaeckle: yeah, those are connectivity and gateway caches ..
I think you need to tune that anyways depending on use-case .. and 20.000 seems a little much for e.g. a minimal setup.

I could also revert to 20k - no hard feelings about that.

kalinkostashki: let's just keep it 2000 since that would be in line with a base deployment.
20k is a more production grade number :)

kalinkostashki (approved):
LGTM :+1:

PR #2192, opened 2025-06-24
fix dynamic WoT validation config to use camel cased setting name

thjaeckle:
instead of the dashed name static config variant of "log-warning-instead-of-failing-api-calls"


thjaeckle:
@hu-ahmed as creator of this feature, would you please review this fix?

Issue #2176, opened 2025-05-27
Add support for "password" OAuth grant-type for HTTP connections


thjaeckle:
At the moment, Ditto only supports [Client Credentials flow](https://eclipse.dev/ditto/connectivity-protocol-bindings-http.html#oauth2-client-credentials-flow) prior to HTTP calls to foreign HTTP APIs managed within a HTTP connection.

However, with more OAuth flows adopting PKCE (without defining a `clientSecret`), the `password` grant type is also an alternative which Ditto should support.

This issue suggests to provide this additional grant type to HTTP push connections.

PR #2195, opened 2025-06-26
#2176: adding support for OAuth2 grant type "password" for HTTP-push connections

thjaeckle:
Resolves: #2176

### DIFF
--- a/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/OAuthPassword.java
+++ b/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/OAuthPassword.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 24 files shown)


connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/OAuthPassword.java
```diff
@@ -0,0 +1,300 @@
...
+            @Nullable final String audience,
+            @Nullable final String username,
+            @Nullable final String password
+    ) {
+        this.tokenEndpoint = checkNotNull(tokenEndpoint, "tokenEndpoint");
+        this.clientId = checkNotNull(clientId, "clientId");
+        this.clientSecret = clientSecret; // clientSecret might be null in case of a public client
+        this.requestedScopes = checkNotNull(requestedScopes, "requestedScopes");
+        this.audience = audience;
+        this.username = checkNotNull(username, "username");
```
kalinkostashki, line 63: you have 3 instances of "username" here so you could do a constant.

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/OAuthPassword.java
```diff
@@ -0,0 +1,300 @@
...
+        this.password = checkNotNull(password, "password");
```
kalinkostashki, line 64: you have 3 instances of "password" here so you could do a constant.

kalinkostashki (changes requested):
LGTM in general.
Since the password field is sensitive would it make sense to use the FieldEncryptionConfig or the simpler approach with the ConnectionUri.createUriStringWithMaskedPassword()

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/OAuthPassword.java
thjaeckle: I would not do that .. I also have that warnings in my IDE, but tend to ignore them - it does not really make the code more readable :)

Issue #2198, opened 2025-07-13
No easy entry to ditto possible. Configurations to start local ditto are not working or are not documented


thfries:
I'm quite frustrated because I lost a lot of time to "just" start a local ditto instance for local development. I assume new people that try to get into ditto will fail too.

This is what I tried:

### Checked out latest release tag 3.7.6 and followed the instructions with docker compose. After six minutes the containers ended up unhealthy:

```
CONTAINER ID   IMAGE                                COMMAND                  CREATED         STATUS                     PORTS                      NAMES
b13091319a83   nginx:1.27-alpine                    "/docker-entrypoint.…"   6 minutes ago   Up 6 minutes               0.0.0.0:8080->80/tcp       docker-nginx-1
cf8ca489fa1e   eclipse/ditto-connectivity:latest    "/sbin/tini -- java …"   6 minutes ago   Up 6 minutes (unhealthy)   8080/tcp                   docker-connectivity-1
9e3ffab9cdc2   eclipse/ditto-gateway:latest         "/sbin/tini -- java …"   6 minutes ago   Up 6 minutes (unhealthy)   0.0.0.0:8081->8080/tcp     docker-gateway-1
ad25cc999840   eclipse/ditto-things:latest          "/sbin/tini -- java …"   6 minutes ago   Up 6 minutes (unhealthy)   8080/tcp                   docker-things-1
d395ea23cfd4   eclipse/ditto-things-search:latest   "/sbin/tini -- java …"   6 minutes ago   Up 6 minutes (unhealthy)   8080/tcp                   docker-things-search-1
42fc9455a1ee   swaggerapi/swagger-ui:v5.9.1         "/docker-entrypoint.…"   6 minutes ago   Up 6 minutes               80/tcp, 8080/tcp           docker-swagger-ui-1
5485f8f28a1e   mongo:7.0                            "docker-entrypoint.s…"   6 minutes ago   Up 6 minutes               0.0.0.0:27017->27017/tcp   docker-mongodb-1
644a5455fbca   eclipse/ditto-ui:latest              "/docker-entrypoint.…"   6 minutes ago   Up 6 minutes               8080/tcp                   docker-ditto-ui-1
29559b74dc75   eclipse/ditto-policies:latest        "/sbin/tini -- java …"   6 minutes ago   Up 6 minutes (unhealthy)   8080/tcp                   docker-policies-1
```

### Installed the latest package 3.7.6 from dockerhub

[...]

PR #2199, opened 2025-07-15
#2198: update docker compose and helm doc

kalinkostashki:
- small corrections to make starting with ditto easier.


thjaeckle (approved):
LGTM 👍

PR #2201, opened 2025-07-24
enhance distributed tracing to report HTTP calls with 4xx or 5xx status code as "failed" and include the response body

thjaeckle:


kalinkostashki (approved):
LGTM :+1:

PR #2210, opened 2025-08-28
fixed CORS behavior for missing "Origin" header

thjaeckle:
fixing Safari CORS issues with Ditto


thjaeckle:
@kalinkostashki could you have a look, please?

PR #2211, opened 2025-08-29
Improve WoT validation performance for "MergeThing" commands

thjaeckle:
* previously, the complete thing (with all of its features) were validated against all submodels 
* this could be potentially very CPU intensive
* even if not much was changed within a `PATCH` or `MergeThing` command, this wasted a lot of resources
* the new approach only validates parts which were changed in the "Merge" command

E.g.:
* if an attribute was modified, all of the attributes are validated
* if part of a feature was modified, all properties of that feature are validated
* if all features were modified, all properties of all features are validated (but then PATCH is not really an optimisation compared to using PUT)
* special case: if the thing's `definition` was updated, the complete thing is validated as all parts have to be valid according to the updated definition

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,8 +16,8 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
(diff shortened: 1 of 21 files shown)


thjaeckle:
@alstanchev @kalinkostashki could you please have a look at this small improvement?  
We have lots of CPU usage in our Ditto installation and hope to get it down by this improvement - only validating features which were contained in a `MergeThing` command instead of all features of the thing.

PR #2212, opened 2025-09-01
improve Thing enforcement performance by caching a resolved policyId in ThingEnforcerActor

thjaeckle:
* and invalidate it only after processing commands which altered the policyId

I encountered via tracing that the `sudo_retrieve_thing` done for each "ThingCommand" can take quite long, e.g. few milliseconds (even if it **should** be a local method invocation eventually).

I added a simple field for storing the `policyId` in `ThingEnforcerActor` and invalidate it when processing commands which alter the policyId.


thjaeckle:
@alstanchev review here would also be appreciated :)

This should also be very beneficial for your setups .. I don't know why we did not think about this before :)

alstanchev (approved):
LGTM 👍 
Great optimisation

Issue #2145, opened 2025-03-11
Extract configurable JWT claims to be added to custom headers


thjaeckle:
Right now, claims from a JWT can be mapped to "auth subject" which Ditto can make use of in its Policies to authorise HTTP calls based on e.g. a `sub` or based on `groups` in a JWT.

There are situations where it would come handy to extract another claim like e.g. `email` and provide this as custom (configurable) header to Ditto.

This would be part of the `openid-connect-issuers` configuration:
```hocon
ditto.gateway.authentication {
    oauth {
      openid-connect-issuers = {
        myprovider = {
          issuer = "localhost:9000"
          auth-subjects = [
            "{{ jwt:sub }}",
            "{{ jwt:groups }}"
          ]
          inject-into-headers = {
            the-users-email = "{{ jwt:email }}"
          }
        }
      }
    }
}
```

This would be helpful to e.g. identify the user without the need of mapping the identifying factor to "auth-subjects" (which is only needed for authorisation, not for identification).

PR #2216, opened 2025-09-02
#2145: support configuration of JWT claim injection into headers

thjaeckle:
Resolves: #2145


thjaeckle:
@alstanchev if you find even some more time 😁 .. this is a new feature which can add arbitrary JWT claims as custom headers..

PR #2220, opened 2025-09-08
fix array sort values in GetSortBsonVisitor

hu-ahmed:
Sorting on array fields caused a JsonParseException because GetSortBsonVisitor#toJsonValue did not handle array values correctly.

**Fix**

Added support in toJsonValue for converting Java collections into proper JsonArray instead of treating them as plain JSON strings.


thjaeckle (approved):
LGTM, thanks a lot for the fix @hu-ahmed

PR #2222, opened 2025-09-10
fix: Parse full certificate chain for client certificates

JCapucho:
Previously the code for parsing the certificate chain for use in connection's client certificate based authentication only parsed the very first certificate of what the user submitted (the client certificate itself), and disregarded all other certificates.

This works when the validation path doesn't include any intermediate certificates. However, as soon as an intermediate certificate is introduced everything stops working, because when the server requests a certificate for it's configured client certificate authority, no path can be made and client certificate authentication fails.

This PR fixes this issue by properly parsing the full certificate chain.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
@@ -26,6 +26,7 @@ import java.security.spec.KeySpec;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Base64;
+import java.util.Collection;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
@@ -109,19 +110,25 @@ final class Keys {
     }
 
     /**
-     * Loads a {@link java.security.cert.Certificate} from the given string. The private key must be given in PKCS#8 format.
+     * Loads a {@link java.security.cert.Certificate} chain from the given string.
      *
-     * @param certificatePem certificate in X.509 format
+     * @param certificatePem certificate chain in X.509 format
      * @param exceptionMapper maps ssl exceptions to ditto exceptions providing context information
-     * @return the certificate
+     * @return the certificate chain
      */
-    static Certificate getCertificate(final String certificatePem, final ExceptionMapper exceptionMapper) {
+    static Certificate[] getCertificateChain(final String certificatePem, final ExceptionMapper exceptionMapper) {
         final byte[] asciiBytes = certificatePem.getBytes(StandardCharsets.US_ASCII);
+        final Collection<? extends Certificate> chain;
         try {
-            return X509_CERTIFICATE_FACTORY.generateCertificate(new ByteArrayInputStream(asciiBytes));
+            chain = X509_CERTIFICATE_FACTORY.generateCertificates(new ByteArrayInputStream(asciiBytes));
         } catch (final CertificateException e) {
             throw exceptionMapper.badCertificateFormat("CERTIFICATE", "DER").build();
         }
+
+        if (chain.isEmpty())
+            throw exceptionMapper.badCertificateFormat("CERTIFICATE", "DER").build();
+
+        return chain.toArray(new Certificate[0]);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
```diff
@@ -115,10 +115,11 @@ private static <T extends KeySpec> T matchKey(final Matcher matcher, final Funct
      * @param exceptionMapper maps ssl exceptions to ditto exceptions providing context information
      * @return the certificate
```
thjaeckle: please adjust javadoc - both of the top description and of the `@return`

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
```diff
@@ -115,10 +115,11 @@ private static <T extends KeySpec> T matchKey(final Matcher matcher, final Funct
...
      */
-    static Certificate getCertificate(final String certificatePem, final ExceptionMapper exceptionMapper) {
+    static Certificate[] getCertificateChain(final String certificatePem, final ExceptionMapper exceptionMapper) {
         final byte[] asciiBytes = certificatePem.getBytes(StandardCharsets.US_ASCII);
         try {
-            return X509_CERTIFICATE_FACTORY.generateCertificate(new ByteArrayInputStream(asciiBytes));
+            return X509_CERTIFICATE_FACTORY.generateCertificates(new ByteArrayInputStream(asciiBytes))
```
thjaeckle: This `X509_CERTIFICATE_FACTORY.generateCertificates` does the "parsing of the full certificate chain", right?

Because returned and used is also only the client certificate, right?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
JCapucho: Sorry forgot to update it. Should be fixed now

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
JCapucho: > This X509_CERTIFICATE_FACTORY.generateCertificates does the "parsing of the full certificate chain", right?

Yes, `generateCertificates` reads a sequence of certificates from the stream as opposed to `generateCertificate` which only reads one certificate. The certificate chain is just various certificates concatenated together, starting from the client certificate and then all intermediates certificates until the CA (which is not included in the chain).

> Because returned and used is also only the client certificate, right?

I'm sorry I don't quite understand the question. This code returns the client certificate and all intermediate certificates included in the chain that the user submitted. All these need to be used when performing client certificate authentication to guarantee a validation path to the root CA.

thjaeckle:
Started a systemTests run at: https://github.com/eclipse-ditto/ditto/actions/runs/17616067613

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/internal/ssl/Keys.java
thjaeckle: @JCapucho ok .. but you only select the first element of the returned certificates chain to be used:
```
keyStoreFactory.setCertificate(keystore, certificate[0]);
```

That's what I meant with that still only the client certificate - and not the chain - is used.

JCapucho: I see what you mean now, as far as i understand `setCertificateEntry` sets a trusted certificate for the keystore, which in the context of client certificate authentication is not relevant. What's actually relevant is the `setPrivateKey` line above it, which sets the private key and the **certificate chain** used for authentication, and as you can see uses the full certificate chain returned from the method.

I'm going to be honest I don't quite understand the purpose of the `setCertificate` line so I just left it with it's original behavior.

JCapucho:
> One relevant System-Test failed: [org.eclipse.ditto.testing.system.connectivity.amqp10.Amqp10SaslIT](https://github.com/eclipse-ditto/ditto/actions/runs/17616067613#user-content-r29)
> 
> We must look into this before being able to accept the PR. Test code is located here: https://github.com/eclipse-ditto/ditto-testing/blob/main/system/src/test/java/org/eclipse/ditto/testing/system/connectivity/amqp10/Amqp10SaslIT.java

I've had a quick look and this test seems to be using hmac credentials and there doesn't seem to be anything involving client certificates (or even amqps) and the timeout error seems a bit suspicious.

I'll try to look deeper as soon as I can find the time.

thjaeckle:
Ran another test-run: https://github.com/eclipse-ditto/ditto/actions/runs/17738343683

The test failure/error is still there - and it is not on a run against the "master" branch .. I'll look into it once I find some time.

PR #2223, opened 2025-09-10
Add support for client certificate authentication for AMQP 1.0

JCapucho:
Adds support for using client certificate authentication in AMQP 1.0 connections by properly configuring the SSL context to use the connection certificate and keys and by setting the EXTERNAL authentication mechanism.


JCapucho:
Hi @thjaeckle, I tested this against an [Apache Qpid Dispatch router](https://qpid.apache.org/components/dispatch-router/index.html), and could confirm that the client certificate authentication was being used (without certificates it didn't work and even with certificates as I having problems because of the problem fixed in #2222) and that with these changes client certificates were exchanged and the handshake succeeded.

thjaeckle (approved):
System tests look good 👍 

LGTM - thanks for the contribution :)

Issue #2207, opened 2025-08-21
Openshift deploy nginx issue


dguarascio:
Hi,

I'm trying to deploy ditto with an umbrella chart on openshift, but I'm having troubles with nginx, by default it tries to open the pidfile on `/var/run/nginx.pid` and on openshift it's not permitted, so I tried to switch to the [unprivileged image](https://hub.docker.com/r/nginxinc/nginx-unprivileged) that should use `/tmp/nginx.pid` but when it start seems to open `/run/nginx.pid`:
```
nginx: [emerg] open() "/run/nginx.pid" failed (13: Permission denied)
```

It's possible to change the path in the default config inside the [template](https://github.com/eclipse-ditto/ditto/blob/master/deployment/helm/ditto/templates/nginx-configmap.yaml) as you already do for [ditto-ui](https://github.com/eclipse-ditto/ditto/blob/master/deployment/helm/ditto/dittoui-config/nginx.conf#L4)?

PR #2225, opened 2025-09-16
Fix deploy on openshift

dguarascio:
This should address #2207, I manually tested the path for the pidfile with `oc apply -f nginx-configmap-patched.yaml -n <my_namespace>`, nginx pod correctly started:
```
+ curl -sL -w '%{http_code}\n' http://<my_namespace>:8080/health -o /dev/null
+ '[[' 200 '!=' 200 ]]
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
```

EDIT: I spotted another permission problem with swaggerui image, in the initScript a new temporary folder is created: `mkdir -p ~/openapi`, on openshift this result in: `mkdir: can't create directory '//openapi': Permission denied`, the reason seems that the random user used to spawn the container doesn't have a home folder, so this tries to create the folder in the root, resulting in a permission denied, in commit https://github.com/eclipse-ditto/ditto/pull/2225/commits/71befe6e42025441e2ca82bd33875119b98b9edf I changed the `initScript` to directly work into `/init-config` folder that is a volume mounted and writable, if you have better alternatives feel free to modify my fix

Resolves: #2207

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.8.0-M91 # chart version is effectively set by release-job
+version: 3.8.0-M91  # chart version is effectively set by release-job
 appVersion: 3.8.0-M6
 keywords:
   - iot-chart


thjaeckle:
Hm, the ticketing system at the Eclipse Foundation also requires to have an Eclipse account.

I would suggest to write to the webmaster: 
https://wiki.eclipse.org/WebMaster

Guara92:
Let me know if I need to bump the chart version too.
Any chance that this can be backported to 3.5?

deployment/helm/ditto/Chart.yaml
```diff
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.8.0-M9  # chart version is effectively set by release-job
+version: 3.8.0-M10 # chart version is effectively set by release-job
```
thjaeckle: @dguarascio could you try to increase to `3.8.0-M91` instead?

Seems the used linting tool does not thing M10 is greater than M9 .. M91 should be greater when it compares alphabetically

thjaeckle: @dguarascio wow, this one is picky .. you need to have 2 spaces before the comment 😄

Guara92: I had to use a different editor to do it 😅

Issue #1790, opened 2023-10-31
Using wildcard in "put-metadata" header for PATCH / MergeThing command adds the metadata also for non-affected fields


thjaeckle:
When e.g. using the following `put-metadata` header:
```json
[{"key":"*/createdAt","value": "now!"}]
```

On a HTTP request:
```
PATCH /api/2/things/my:thing-1
```

E.g. sending the payload:
```json
{
  "attributes": {
    "only-one": 2
  }
}
```

Will also lead to all existing fields in the Thing getting the metadata, e.g. as in:
```json
{
  "_metadata": {
    "thingId": {
        "createdAt": "now!"
    },
    "policyId": {
        "createdAt": "now!"
    },
    "attributes": {
        "only-one": {
            "createdAt": "now!"
        }
    }
  }
}
```

That is unexpected and should instead only modify the metadata of the provided fields to "PATCH".

sejal-bansal:
@thjaeckle I would like to work on this issue. Could you please assign it to me?

thjaeckle:
@sejal-bansal cool, thanks
This however is probably not trivial to fix, I would assume.

thjaeckle:
@sejal-bansal any update here? Do you work on this or shall I unassign the issue again?

PR #2226, opened 2025-09-16
fix several issues around "put-metadata" in combination with MergeThing commands

thjaeckle:
* wildcards were not correctly placed (fixes: #1790)
* using non root "/" path was not supported for MergeThing command
* rules about at which places metadata can be placed were overly complex and faulty
* validation of wildcards was restricting use of wildcard at many places


thjaeckle:
@alstanchev system-tests are green except one flaky one .. which locally is also green

PR #2229, opened 2025-09-18
support silent token refresh

hu-ahmed:
This PR adds support for silent token refresh to the Ditto UI when using OIDC/SSO authentication.
Users no longer need to manually re-login once their token expires; the session will be refreshed in the background.


thjaeckle (approved):
LGTM - thanks a lot for the addition @hu-ahmed

PR #2230, opened 2025-09-19
enhance Helm config with "maxUnavailable" for pdb and G1 GC option

thjaeckle:
Also fix the bitnami mongodb dependency in this PR .. Helm build is broken currently because of the bitnami changes.


thjaeckle:
@alstanchev or @kalinkostashki could you have a look, please?

PR #2232, opened 2025-10-01
add multiple patch conditions

hu-ahmed:


documentation/src/main/resources/pages/ditto/basic-conditional-requests.md
```diff
@@ -160,6 +160,154 @@ was introduced.
 Upon activation, the digital twin stored in Eclipse Ditto will implicitly be updated with the latest data from the 
 _live response_ sent by the device.
 
+## Path-specific conditions (Since 3.8.0)
+
+In addition to global conditions, Ditto supports path-specific conditions for merge operations using the `patch-conditions` header. This feature allows you to apply different RQL conditions to different parts of a merge patch, enabling fine-grained control over which parts of the patch are applied based on the current state of the Thing.
```
thjaeckle: Hm, maybe we should name the header `merge-thing-patch-conditions` to already make clear in the header that only the `MergeThing` command supports that ..

What do you think @hu-ahmed ?

documentation/src/main/resources/pages/ditto/basic-conditional-requests.md
```diff
@@ -160,6 +160,154 @@ was introduced.
...
+
+The third option to use path-specific conditions is the ditto-client.
+The following code snippet demonstrates how to achieve this.
+
+```java
+final Map<String, String> patchConditions = new HashMap<>();
+patchConditions.put("features/temperature/properties/value", "gt(features/temperature/properties/value,20)");
+patchConditions.put("features/humidity/properties/value", "lt(features/humidity/properties/value,80)");
+
+final Option<Map<String, String>> patchConditionsOption = Options.patchConditions(patchConditions);
```
thjaeckle: In order to make this work, you would need to also provide a PR to `ditto-clients` to add this `Options.patchConditions` method @hu-ahmed 

I like it .. we should do it 👍

things/model/src/main/java/org/eclipse/ditto/things/model/signals/commands/modify/MergeThing.java
```diff
@@ -599,6 +600,21 @@ public JsonValue getValue() {
...
     }
 
+    /**
+     * Returns the patch conditions from the patch-conditions header as an Optional JsonObject.
+     * The patch conditions map JSON pointer paths to RQL condition expressions.
+     * 
+     * @return Optional containing the patch conditions JsonObject, or empty if the header is not present
+     * @since 3.8.0
+     */
+    public Optional<JsonObject> getPatchConditions() {
```
thjaeckle: Hm, we could also move this to `DittoHeaders` and the method could already return a `Map<JsonPointer, String>` .. 
And we could the "parsing" of the patch conditions as part of the implementation in `DittoHeaders` - so using it would be quite easy with `dittoHeaders.getThingMergePatchConditions()`

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingCommandEnforcement.java
```diff
@@ -461,6 +468,43 @@ private static void enforceReadPermissionOnCondition(final String condition,
...
+     * on all of them.
+     * </p>
+     * 
+     * @param patchConditions the JSON object containing path-to-condition mappings
+     * @param enforcer the policy enforcer to check permissions against
+     * @param dittoHeaders the Ditto headers for error reporting
+     * @throws ThingConditionFailedException if the user lacks READ permission on any referenced resource
+     * @since 3.8.0
+     */
+    private static void enforceReadPermissionOnPatchConditions(final JsonObject patchConditions,
```
thjaeckle: This should probably also be moved to the Util, right?

documentation/src/main/resources/pages/ditto/basic-conditional-requests.md
hu-ahmed: yeah make sense

documentation/src/main/resources/pages/ditto/basic-conditional-requests.md
```diff
@@ -160,6 +160,154 @@ was introduced.
...
+  }
+}
+```
+
+### HTTP API
+
+Using the HTTP API it is possible to specify path-specific conditions via HTTP header
+
+```
+curl -X PATCH -H 'Content-Type: application/merge-patch+json' -H 'patch-conditions: {"features/temperature/properties/value": "gt(features/temperature/properties/value,20)", "features/humidity/properties/value": "lt(features/humidity/properties/value,80)"}' /api/2/things/org.eclipse.ditto:fancy-thing -d '{"features": {"temperature": {"properties": {"value": 25}}, "humidity": {"properties": {"value": 60}}, "status": {"properties": {"state": "updated"}}}}'
```
thjaeckle: merge-thing-patch-conditions

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -583,6 +583,21 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+    /**
+     * Header definition for patch conditions in merge operations.
+     * <p>
+     * Key: {@code "merge-thing-patch-conditions"}, Java type: {@link JsonObject}.
+     * Contains a JSON object mapping JsonPointer paths to RQL condition expressions.
+     * </p>
+     *
+     * @since 3.8.0
+     */
+    PATCH_CONDITIONS("merge-thing-patch-conditions",
```
thjaeckle: The constant could also be renamed accordingly

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PatchConditionsEvaluator.java
```diff
@@ -0,0 +1,180 @@
...
+        } catch (final Exception e) {
+            throw InvalidRqlExpressionException.newBuilder()
+                    .message(e.getMessage())
+                    .cause(e)
+                    .dittoHeaders(dittoHeaders)
+                    .build();
+        }
+    }
+
+    private static boolean doesMigrationPayloadContainResourceKey(final JsonObject migrationPayload,
```
thjaeckle: "wrong" method name now that this not only is used for "migration payload"

things/service/src/main/java/org/eclipse/ditto/things/service/enforcement/ThingCommandEnforcement.java
hu-ahmed: but we have in the same class the other enforcement

thjaeckle: ah, true .. it is already central at one place 👍

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PatchConditionsEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+        final JsonObjectBuilder adjustedPayloadBuilder = mergeObject.toBuilder();
+
+        for (final Map.Entry<JsonPointer, String> entry : patchConditions.entrySet()) {
+            final JsonPointer conditionPath = entry.getKey();
+            final String conditionExpression = entry.getValue();
+
+            final boolean conditionMatches = evaluateCondition(existingThing, conditionExpression, command.getDittoHeaders());
+            final boolean containsResource = mergeObject.getValue(conditionPath).isPresent();
+
+            if (!conditionMatches && containsResource) {
```
thjaeckle: @hu-ahmed small optimisation: there is no need to evaluate the condition (which is a more expensive operation than checking for presence of the resource) if `containsResource` is already `false` ..

PR #2246, opened 2025-10-20
Fixes for reducing ERROR logs during restart of "ditto-things" service

thjaeckle:
* made local ask timeout for sending messages to "Enforcer" actor differently configurable than the local ask-timeout for interacting with persistence actor
* retry things aggregation after searches on RemoteStreamRefActorTerminatedException
   * which are thrown during rolling updates / restarts of things
* introduce "askWithRetry" pattern to signal enrichment in connectivity in order to stabilise retrieving "extra" fields when timeouts occur


alstanchev:
Sure, already skimmed over it but didn't give it much time since the draft status. I will go over it again now.

PR #2247, opened 2025-10-23
fix feature toggles were not applied correctly

thjaeckle:
I noticed that those were not correctly read out - and always resolved to `true` as consequence.


thjaeckle:

PR #2248, opened 2025-10-23
add configuration for string escaping buffer size in order to be able to tune a very often invoked operation

thjaeckle:
During profiling we identified that `org.eclipse.ditto.json.JavaStringToEscapedJsonString.apply(Object)` uses a lot of CPU resources when invoking `StringBuilder.append(char)`, invoking `AbstractStringBuilder.ensureCapacityInternal(int)`.  
We assume that this is because the chosen buffer-size of the string builder is too small:
https://github.com/eclipse-ditto/ditto/blob/7567896502d5f698a0573abed65a58b212ec699e/json/src/main/java/org/eclipse/ditto/json/JavaStringToEscapedJsonString.java#L54

If there are any characters which need to be escaped in a string, the buffer size is too small. This will cause `ensureCapacityInternal(int)` to start copying arrays - which results in a lot of CPU heavy work.

This PR proposes a config flag to configure a factor with which this capacity can be increased. The default is `1.0` - which is the current behavior.  
I did not want to always increase this, as this will also have an impact of memory-demands .. the higher the factor, the more memory will be reserved for each operation.

I assume we will try out a factor of `2.0` - as only few chars in json strings should need escaping. Most commonly probably double quotes `"` or newlines.

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.8.3  # chart version is effectively set by release-job
+version: 3.8.4  # chart version is effectively set by release-job
 appVersion: 3.8.2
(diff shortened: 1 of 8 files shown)


alstanchev (approved):
LGTM 👍

PR #2260, opened 2025-11-07
fix wrongly configured healthcheck name for "search" shard region

thjaeckle:
* leading to startup failures of thingssearch


thjaeckle:
This is rather critical as things-search will not start up in a cluster 🙄

PR #2262, opened 2025-11-10
fix wrong CBOR upper bound calculation

thjaeckle:
* in case CBOR representation is bigger than jsonString representation
* made JacksonCborFactory.toByteBuffer more error-proof by handling BufferOverflowException, doubling the buffer

The latest improvements in Ditto 3.8.4 / 3.8.5 caused potential `BufferOverflowException` (depending on the validated payload) when using WoT based validation.

Reason was a wrongly determined "upper bound" for a ByteBuffer allocation.  
This fix also introduces an additional safeguard, increasing the allocated buffer on an exception instead of failing with `BufferOverflowException`.


thjaeckle:
should only affect WoT validation - nonetheless I would want to build another bugfix release containing this

PR #2265, opened 2025-11-12
improve stability of ThingsAggregatorProxy in "things" service during restart

thjaeckle:
* right now it fails when encountering ask timeouts (e.g. because shard proxy is not available)
* improve this by using "ask-with-retry" instead


alstanchev (approved):
LGTM 👍

Issue #96, opened 2018-01-15
Support emitting partial change notifications based on Policy READ permission


thjaeckle:
Currently we only deliver change notifications via the WS or SSE for subjects which have unrestricted "READ" permission for changed Things (via the so-called "read-subjects").

For ACL that was and is sufficient.

But when using a Policy for a Thing in which I only have "READ" permission for a single Feature property I would expect to get a WS message if this property is changed (or also if the complete Feature/Thing is changed and the property was part of that change).

That is however a little difficult as we would have to build "views" of each emitted Event for each WS session.

As we do not want to instantiate the PolicyEnforcer for all used policies again in the Websocket sessions I suggest serializing this kind of information (which subject is allowed to read which parts of the Thing affected by an emitted ThingEvent) as header field.

The format of the header could look like:
```json
{
  "sub1": [
    "/features/lamp1/properties/illuminance",
    "/attributes"
  ],
  "sub2": [
    "/"
  ]
}
```

This format could be calculated by the `PolicyEnforcerActor` for an incoming `ModifyCommand` - and fill the header based on the actually changed Json fields inside.

The Websocket/SSE would "only" have to use a small interpreter for this format in order to return only a partial Event.

thjaeckle:
Some more background:

## Example

### Policy

```json
{
  "policyId":"my.namespace:policy-a",
  "entries":{
    "owner":{
      "subjects":{
        "nginx:owner":{
          "type":"owner user A"
        }
      },
      "resources":{
        "thing:/":{
          "grant":[ "READ", "WRITE" ],
          "revoke":[ ]
        },
        "policy:/":{
          "grant":[ "READ", "WRITE" ],
          "revoke":[ ]
        }
      }
    },
    "attr-observer":{
      "subjects":{
        "nginx:attr-observer":{
          "type":"attributes observer user"
        }
      },
      "resources":{
        "thing:/attributes":{
          "grant":[ "READ" ],
          "revoke":[ ]
        },
        "thing:/attributes/secret":{
          "grant":[ ],
          "revoke":[ "READ" ]
        }
      }
    },
    "feature-a-status-observer":{
      "subjects":{
        "nginx:feature-a-status-observer":{
          "type":"Feature A status observer user"
        },
        "resources":{
          "thing:/features/A/properties/status":{
            "grant":[ "READ" ],
            "revoke":[ ]
          }
        }
      }
    }
  }
}
```

### Thing

```json
{
  "thingId":"my.namespace:thing-a",
[...]

PR #2266, opened 2025-11-12
Implement partial read access events emitting

hu-ahmed:
#96

### DIFF
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
@@ -310,6 +310,9 @@ public final class ThingEventEnricher {
                     final Map<String, List<JsonPointer>> partialAccessPaths =
                             PartialAccessPathCalculator.calculatePartialAccessPaths(
                                     thingEvent, thing, policyEnforcerOpt.get());
+                    if (partialAccessPaths.isEmpty()) {
+                        return JsonFactory.newObject();
+                    }
                     final JsonObject result = PartialAccessPathCalculator.toIndexedJsonObject(partialAccessPaths);
                     LOGGER.debug("Calculated partial access paths for event '{}' (thingId: {}): {} subjects with partial access, result: {}",
                             thingEvent.getType(), thingEvent.getEntityId(), partialAccessPaths.size(), result);
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
@@ -78,10 +78,16 @@ public final class PartialAccessPathCalculator {
         }
 
         final Map<String, List<JsonPointer>> result = new LinkedHashMap<>();
-
         final JsonObject thingJson = thing.toJson();
 
-        for (final AuthorizationSubject subject : subjectsWithPartialPermission) {
+        final var effectedSubjects = enforcer.getSubjectsWithPermission(rootResourceKey,
+                Permissions.newInstance(Permission.READ));
+        final Set<AuthorizationSubject> subjectsWithPermissionOnRoot = effectedSubjects.getGranted();
+        
+        final Set<AuthorizationSubject> subjectsWithRestrictedAccess = new LinkedHashSet<>(subjectsWithPartialPermission);
+        subjectsWithRestrictedAccess.removeAll(subjectsWithPermissionOnRoot);
+
+        for (final AuthorizationSubject subject : subjectsWithRestrictedAccess) {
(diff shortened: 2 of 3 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
```diff
@@ -246,35 +253,109 @@ private List<MappingOutcome<OutboundSignal.Mapped>> processMappableSignals(
...
-        final var adaptableWithInternalCorrelationId = mappableSignals.stream()
-                .findFirst()
-                .map(signal -> setInternalCorrelationIdToAdaptable(adaptable, signal.getSource()))
-                .orElse(adaptable);
+        final boolean hasPartialAccessPaths = dittoHeaders.containsKey(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
 
         return mappingTimer.overall(() -> mappableSignals.stream()
                 .flatMap(mappableSignal -> {
-                    final Signal<?> source = mappableSignal.getSource();
+                    final Signal<?> source = signalToMap;
```
thjaeckle: @hu-ahmed I don't think this change is right .. 
This should be `mappableSignal.getSource();` - because in a payload mapping, a single "signal" can cause multiple signals to be published.

E.g. a JavaScript payload mapping can create 10 messages to be published for a single event.  
All of those would need to be checked/filtered for permissions.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
```diff
@@ -246,35 +253,109 @@ private List<MappingOutcome<OutboundSignal.Mapped>> processMappableSignals(
...
-                    return mappers.stream()
-                            .flatMap(mapper -> runMapper(
-                                    mappableSignal,
-                                    adaptableWithInternalCorrelationId,
-                                    mapper,
-                                    mappingTimer
-                            ));
+                    
+                    if (targets.isEmpty()) {
+                        return processTargetsEmpty(mappableSignal, source, outboundSignal, mappingTimer, mappers);
```
thjaeckle: Why would we need to even run the mapper if there are no targets to publish to?

Maybe this optimisation was also not done before - but it seems to make not much sense.

base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
```diff
@@ -612,6 +612,56 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
...
+     * Used together with {@link #PRE_DEFINED_EXTRA_FIELDS_READ_GRANT_OBJECT} to reduce header size
+     * by using indices instead of repeating subject IDs.
+     * <p>
+     * Key: {@code "ditto-pre-defined-extra-fields-read-grant-subjects"}, Java type: {@link JsonObject}.
+     * Format: {@code {"0": "subject1", "1": "subject2", ...}}
+     * </p>
+     *
+     * @since 3.9.0
+     */
+    PRE_DEFINED_EXTRA_FIELDS_READ_GRANT_SUBJECTS("ditto-pre-defined-extra-fields-read-grant-subjects",
```
thjaeckle: I see this constant only used in `ThingEventEnrichter` to set the header - but it never is read.  
Is this still missing? Should it reuse/replace the subjects from the `PARTIAL_ACCESS_PATHS` header?

thjaeckle: I guess it would make sense to define just one header
* being a JsonArray
* with a list of "subjects" - used in both "pre-defined-extra-read-subjects" and in the partial access paths

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/PartialAccessPathResolver.java
```diff
@@ -0,0 +1,258 @@
...
+ */
+public final class PartialAccessPathResolver {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PartialAccessPathResolver.class);
+
+    // TODO check if needed to be configurable
+    private static final int MAX_CACHE_SIZE = 5000;
+    private static final Duration CACHE_EXPIRE_AFTER_ACCESS = Duration.ofMinutes(15);
+    private static final Duration CACHE_EXPIRE_AFTER_WRITE = Duration.ofMinutes(15);
+    private static final com.github.benmanes.caffeine.cache.Cache<String, IndexedPartialAccessPaths> PARSED_HEADER_CACHE =
```
thjaeckle: Are you sure that a cache is really needed here?

I would not add it pre-maturely - but maybe instead add a metric to get insights how it performs.

PR #2268, opened 2025-11-14
fix search timeout error handling

thjaeckle:
* not logging an "error" log
* returning a 408 (client error, timeout) status instead of 504 (server timeout) as this indicates a too complex query
* preserving DittoHeaders for logs and exception


alstanchev (approved):
LGTM 👍

PR #2270, opened 2025-11-18
add possibility to configure tracing tick-interval and queue-size

thjaeckle:
This is e.g. needed if not all spans are reaching the OTEL tracing backend.

Kamon documents for the `reporterQueueSize`:
> If the queue becomes full then sampled finished spans will be dropped in order to avoid consuming excessive
> amounts of memory. Each configured reporter has a separate queue.


alstanchev (approved):
LGTM 👍

PR #2275, opened 2025-11-20
add feature toggle for configuring trace span metric reporting

thjaeckle:
* by default Kamon enables reporting trace spans as histogram "span_processing_time_seconds"
* make this configurable for custom created Ditto spans

### DIFF
--- a/deployment/helm/ditto/templates/connectivity-deployment.yaml
+++ b/deployment/helm/ditto/templates/connectivity-deployment.yaml
@@ -216,6 +216,8 @@ spec:
               value: "{{ .Values.global.featureFlags.preserveKnownMqttHeadersEnabled }}"
             - name: DITTO_DEVOPS_FEATURE_JSON_KEY_VALIDATION_ENABLED
               value: "{{ .Values.global.featureFlags.jsonKeyValidationEnabled }}"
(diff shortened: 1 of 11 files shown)


deployment/helm/ditto/templates/connectivity-deployment.yaml
```diff
@@ -216,6 +216,8 @@ spec:
               value: "{{ .Values.global.featureFlags.preserveKnownMqttHeadersEnabled }}"
             - name: DITTO_DEVOPS_FEATURE_JSON_KEY_VALIDATION_ENABLED
               value: "{{ .Values.global.featureFlags.jsonKeyValidationEnabled }}"
+            - name: DITTO_DEVOPS_FEATURE_TRACING_SPAN_METRICS_ENABLED
```
kalinkostashki, line 219: as this won't take any effect if tracing is disabled shouldn't this be conditional based on that value. Or maybe just give a tip in the feature flags that you should enable that one first.

thjaeckle: @kalinkostashki agreed .. I had it documented in the code, but not in the `.conf` of the feature flags ..
I added the documentation also there

PR #2279, opened 2025-11-24
fix retrieving thing for a message with a condition

thjaeckle:
* by removing conditional headers
* by removing "response-required:False" and "timeout" which would fail the "retrieve"
* by switching from SudoRetrieveThing to RetrieveThing to not bypass policy READ check


alstanchev (approved):
LGTM 👍

alstanchev (approved):

PR #2286, opened 2025-12-05
introduce feature toggle to choose implementation of policy evaluator

thjaeckle:
Ditto has 2 implementations - however it is hardcoded that the "throughput optimized" is always used. 

This PR adds a feature toggle which lets the Ditto operator choose the memory optimized variant instead.


policies/model/README.md
```diff
@@ -10,7 +10,7 @@ http://java-performance.info/jmh/
 
 Simply execute:
 ```bash
-mvn clean package
+mvn clean package -Pbuild-benchmark-assembly
 
 java -jar target/ditto-model-policies-enforcers-<version>-benchmark.jar
```
alstanchev: the jar name is wrong. Should be:
ditto-model-policies-enforcers-<version>-benchmark.jar
Guess it's been like that for some time.

PR #2287, opened 2025-12-06
#96 implement partial read access events emitting

hu-ahmed:
Resolves: #96

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
@@ -241,25 +241,29 @@ public final class OutboundMappingProcessor extends AbstractMappingProcessor<Out
     ) {
         final var outboundSignalSource = outboundSignal.getSource();
         final var dittoHeaders = outboundSignalSource.getDittoHeaders();
+        final String partialAccessPathsHeader =
+                dittoHeaders.get(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
+        final Map<String, List<JsonPointer>> partialAccessPaths =
+                PartialAccessPathResolver.parsePartialAccessPathsAsMap(partialAccessPathsHeader);
 
         final var mappingTimer = MappingTimer.outbound(connectionId, connectionType, dittoHeaders);
+        final boolean hasPartialAccessPaths = partialAccessPathsHeader != null && !partialAccessPathsHeader.isEmpty();
+        final Map<PayloadMapping, List<MessageMapper>> mappersByPayloadMapping = new LinkedHashMap<>();
+        for (final OutboundSignal.Mappable mappableSignal : mappableSignals) {
+            mappersByPayloadMapping.computeIfAbsent(mappableSignal.getPayloadMapping(), this::getMappers);
+        }
 
         return mappingTimer.overall(() -> mappableSignals.stream()
                 .flatMap(mappableSignal -> {
                     final Signal<?> source = mappableSignal.getSource();
                     final List<Target> targets = mappableSignal.getTargets();
-                    final List<MessageMapper> mappers = getMappers(mappableSignal.getPayloadMapping());
+                    final List<MessageMapper> mappers =
+                            mappersByPayloadMapping.get(mappableSignal.getPayloadMapping());
                     logger.withCorrelationId(source)
                             .debug("Resolved mappers for message {} to targets {}: {}", source, targets, mappers);
-                    
-                    final boolean hasPartialAccessPaths = dittoHeaders.containsKey(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
-                    
+
                     final TargetMappingStrategy mappingStrategy;
                     if (hasPartialAccessPaths && !targets.isEmpty()) {
-                        final String partialAccessPathsHeader =
-                                dittoHeaders.get(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
-                        final Map<String, List<JsonPointer>> partialAccessPaths =
-                                PartialAccessPathResolver.parsePartialAccessPathsAsMap(partialAccessPathsHeader);
                         mappingStrategy = new PartialAccessTargetMapping(mappableSignal, source, targets, outboundSignal,
                                 mappingTimer, mappers, partialAccessPathsHeader, partialAccessPaths);
                     } else {
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
@@ -93,19 +93,19 @@ public final class ThingEventEnricher {
      * Enriches the passed in {@code withDittoHeaders} with pre-defined extraFields based on the provided {@code thing}
      * and the global configuration (based on namespace and optional RQL condition).
      *
+     * @param preDefinedExtraFieldsConfigs the list of config entries for pre-defined extraFields enrichment
      * @param thingId the Thing ID to enrich for
      * @param thing the Thing entity to use for getting extra fields from
      * @param policyId the Policy ID to use for looking up permissions
-     * @param preDefinedExtraFieldsConfigs the list of config entries for pre-defined extraFields enrichment
      * @param withDittoHeaders the object to enrich with pre-defined extraFields (e.g. a Signal)
      * @param <T> the type of the signal to enrich
      * @return an enriched version of the passed in {@code withDittoHeaders} with pre-defined extraFields
      */
     public <T extends DittoHeadersSettable<? extends T>> CompletionStage<T> enrichWithPredefinedExtraFields(
+            final List<PreDefinedExtraFieldsConfig> preDefinedExtraFieldsConfigs,
             final ThingId thingId,
             @Nullable final Thing thing,
             @Nullable final PolicyId policyId,
-            final List<PreDefinedExtraFieldsConfig> preDefinedExtraFieldsConfigs,
             final T withDittoHeaders
     ) {
         final CompletionStage<T> partialAccessPathsStage =
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
@@ -292,11 +292,6 @@ public final class PartialAccessPathCalculator {
         for (final Map.Entry<String, List<JsonPointer>> entry : partialAccessPaths.entrySet()) {
             final String subjectId = entry.getKey();
             final Integer subjectIndex = subjectToIndex.get(subjectId);
-            
-            if (subjectIndex == null) {
-                // Subject not in index (should not happen, but defensive check)
-                continue;
-            }
 
             for (final JsonPointer path : entry.getValue()) {
                 pointerToSubjectIndexes.computeIfAbsent(path, k -> new LinkedHashSet<>())
@@ -339,4 +334,3 @@ public final class PartialAccessPathCalculator {
                 .build();
     }
 }
-
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/ThingPersistenceActor.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/ThingPersistenceActor.java
@@ -275,6 +275,7 @@ public final class ThingPersistenceActor
     @Override
     protected void publishEvent(@Nullable final Thing previousEntity, final ThingEvent<?> event) {
         final CompletionStage<ThingEvent<?>> stage = thingEventEnricher.enrichWithPredefinedExtraFields(
+                thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                 entityId,
                 entity,
                 Optional.ofNullable(entity).flatMap(Thing::getPolicyId)
@@ -282,7 +283,6 @@ public final class ThingPersistenceActor
                                 .flatMap(Thing::getPolicyId)
                                 .orElse(null)
                         ),
-                thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                 event
         );
         stage.whenComplete((modifiedEvent, ex) -> {
@@ -335,18 +335,18 @@ public final class ThingPersistenceActor
         switch (signal) {
             case MessageCommand<?, ?> messageCommand ->
                 stage = thingEventEnricher.enrichWithPredefinedExtraFields(
+                        thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                         entityId,
                         entity,
                         Optional.ofNullable(entity).flatMap(Thing::getPolicyId).orElse(null),
-                        thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                         messageCommand
                 );
             case ThingEvent<?> thingEvent ->
                 stage = thingEventEnricher.enrichWithPredefinedExtraFields(
+                        thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                         entityId,
                         entity,
                         Optional.ofNullable(entity).flatMap(Thing::getPolicyId).orElse(null),
-                        thingConfig.getEventConfig().getPredefinedExtraFieldsConfigs(),
                         thingEvent
                 );
             default ->
--- a/things/service/src/test/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricherTest.java
+++ b/things/service/src/test/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricherTest.java
@@ -308,7 +308,7 @@ public final class ThingEventEnricherTest {
 
         final CompletionStage<AttributeModified> resultStage =
                 sut.enrichWithPredefinedExtraFields(
-                        KNOWN_THING_ID, KNOWN_THING, KNOWN_POLICY_ID, preDefinedExtraFieldsConfigs, event);
+                        preDefinedExtraFieldsConfigs, KNOWN_THING_ID, KNOWN_THING, KNOWN_POLICY_ID, event);
         return resultStage.thenApply(AttributeModified::getDittoHeaders);
     }
 


policies/model/src/main/java/org/eclipse/ditto/policies/model/enforcers/tree/TreeBasedPolicyEnforcer.java
```diff
@@ -339,8 +354,86 @@ private static JsonPointer getPrefixPointerOrThrow(final JsonPointer pointer, fi
...
+
+        final int levelCount = resourcePath.getLevelCount();
+        final Set<JsonPointer> accessiblePaths = new HashSet<>();
+        candidates.stream()
+                .filter(pointerAndValue -> pointerStartsWith(pointerAndValue.pointer, resourcePath))
+                .filter(pointerAndValue -> isAccessible(pointerAndValue.pointer, grantedResources, revokedResources))
+                .forEach(pointerAndValue -> {
+                    final JsonPointer subPointer = pointerAndValue.pointer.getSubPointer(levelCount).orElseThrow(() -> {
+                        final String msgPattern = "JsonPointer did not contain a sub-pointer for level <{0}>!";
+                        return new NullPointerException(MessageFormat.format(msgPattern, levelCount));
```
thjaeckle: I would never instantiate a NullPointerException .. This is IMO misuse of the JVM "null" errors in Java.

Is this a bug in code if this happens? Then I would throw `IllegalStateException`
Is this something which the user can somehow influence? Then I would throw a proper DittoRuntimeException.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
```diff
@@ -0,0 +1,385 @@
...
+ * with restricted READ permissions.
+ */
+public final class ThingEventEnricher {
+
+    private static final DittoLogger LOGGER = DittoLoggerFactory.getLogger(ThingEventEnricher.class);
+
+    private static final TimePlaceholder TIME_PLACEHOLDER = TimePlaceholder.getInstance();
+    private static final HeadersPlaceholder HEADERS_PLACEHOLDER = PlaceholderFactory.newHeadersPlaceholder();
+
+    private static final Map<String, Criteria> RQL_CRITERIA_CACHE = new ConcurrentHashMap<>(32);
```
thjaeckle: Is this intended as a cache?  
I would not do that - the ConcurrentHashMap can in fact make the whole operation even slower than not using a cache.

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
```diff
@@ -0,0 +1,385 @@
...
+            final String rqlCondition = optCondition.get();
+
+            try {
+                final Criteria criteria = RQL_CRITERIA_CACHE.computeIfAbsent(rqlCondition, rql -> {
+                    try {
+                        return QueryFilterCriteriaFactory
+                                .modelBased(RqlPredicateParser.getInstance())
+                                .filterCriteria(rql, DittoHeaders.empty());
+                    } catch (final InvalidRqlExpressionException e) {
+                        LOGGER.warn("Invalid RQL condition <{}> - ignoring: {}", rql, e.getMessage());
```
thjaeckle: Always provide the DittoHeaders to loggers, if they are present - that way e.g. the `correlation-id` and `trace-parent` are automatically added as MDC:
```
LOGGER.withCorrelationId(withDittoHeaders).warn(...
```

Ah, I see this was also missing before .. shame :D

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * Safely gets the policy enforcer stage, handling null cases defensively.
+     *
+     * @param policyId the policy ID to get the enforcer for
+     * @return a non-null CompletionStage, completed with empty Optional if the provider returns null
+     */
+    private CompletionStage<Optional<PolicyEnforcer>> getPolicyEnforcerSafely(@Nullable final PolicyId policyId) {
+        final CompletionStage<Optional<PolicyEnforcer>> stage = policyEnforcerProvider.getPolicyEnforcer(policyId);
+        if (stage == null) {
```
thjaeckle: This can never happen as far as I can see .. the returned CompletionStage is never null

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
```diff
@@ -0,0 +1,385 @@
...
+import org.eclipse.ditto.things.service.utils.ReadGrantIndexer;
+
+/**
+ * Encapsulates functionality in order to enrich ThingEvents with pre-defined {@code extraFields} via DittoHeaders
+ * (fields defined per namespace in the Ditto things configuration) and with partial access paths for subjects
+ * with restricted READ permissions.
+ */
+public final class ThingEventEnricher {
+
+    private static final DittoLogger LOGGER = DittoLoggerFactory.getLogger(ThingEventEnricher.class);
```
thjaeckle, line 70: For every `LOGGER` use in this class, please provide the DittoHeaders as context for MDC.
```
LOGGER.withCorrelationId(dittoHeaders)
```

things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/enrichment/ThingEventEnricher.java
```diff
@@ -0,0 +1,385 @@
...
+     * Constructs a new enricher for ThingEvents based on the provided configuration and policy enforcer.
+     *
+     * @param preDefinedExtraFieldsConfigs the list of config entries for pre-defined extraFields enrichment
+     * @param policyEnforcerProvider the policy enforcer to use in order to check permissions for enriching extraFields
+     */
+    public ThingEventEnricher(
+            final List<PreDefinedExtraFieldsConfig> preDefinedExtraFieldsConfigs,
+            final PolicyEnforcerProvider policyEnforcerProvider
+    ) {
+        this.preDefinedExtraFieldsConfigs = List.copyOf(preDefinedExtraFieldsConfigs);
```
thjaeckle: I would not copy the list - I found out in a heap dump that this costs a lot of memory.

I changed/fixed that on the `master` branch in `PreDefinedExtraFieldsEnricher` in the meantime - instead of providing the `preDefinedExtraFieldsConfigs` as constructor argument, it is now passed into as argument in e.g. `enrichWithPredefinedExtraFields`.

That way, the `ThingEventEnricher` instance will not contain much own state.

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
```diff
@@ -0,0 +1,338 @@
...
+ * </p>
+ *
+ * @since 3.9.0
+ */
+public final class PartialAccessPathCalculator {
+
+    /**
+     * JSON key for the subjects array in the partial access paths structure.
+     */
+    public static final String SUBJECTS_KEY = "subjects";
```
thjaeckle: This should be a `JsonFieldDefinition` instead of a `String` - that way we can define its type as well and get more type safety.

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
```diff
@@ -0,0 +1,338 @@
...
+    
+    /**
+     * JSON key for the paths object in the partial access paths structure.
+     */
+    public static final String PATHS_KEY = "paths";
```
thjaeckle: This should be a `JsonFieldDefinition` instead of a `String` - that way we can define its type as well and get more type safety.

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
```diff
@@ -0,0 +1,338 @@
...
+     * @param thing the current Thing state (may be null for ThingDeleted events)
+     * @param policyEnforcer the PolicyEnforcer to use for permission checks (must not be null)
+     * @return a map from subject ID to list of accessible JsonPointer paths, empty map if no partial subjects exist
+     * @throws NullPointerException if policyEnforcer is null
+     */
+    public static Map<String, List<JsonPointer>> calculatePartialAccessPaths(
+            @Nullable final Thing thing,
+            final PolicyEnforcer policyEnforcer) {
+
+        if (thing == null || policyEnforcer == null) {
```
thjaeckle: `policyEnforcer` cannot be null at this place.
If it can be `null`, the parameter must be declared as `@Nullable`.
However, the only place calling it makes sure it is not null.

policies/model/src/main/java/org/eclipse/ditto/policies/model/enforcers/Enforcer.java
```diff
@@ -313,4 +314,19 @@ default JsonObject buildJsonView(final ResourceKey resourceKey,
...
+     * @param resourceKey the ResourceKey (containing Resource type and path) to start from
+     * @param jsonFields the full JsonFields from which to determine accessible paths
+     * @param authorizationContext the AuthorizationContext containing the AuthorizationSubjects
+     * @param permissions the permissions to check
+     * @return a set of JsonPointer paths that are accessible for the given authorization context and permissions
+     * @throws NullPointerException if any argument is {@code null}.
+     * @since 3.9.0
+     */
+    Set<JsonPointer> getAccessiblePaths(ResourceKey resourceKey, Iterable<JsonField> jsonFields,
+            AuthorizationContext authorizationContext, Permissions permissions);
```
thjaeckle, line 331: Instead of the param `AuthorizationContext authorizationContext` we could also just use `AuthorizationSubject authorizationSubject`.  
As far as I can see, we always only "wrap" a single subject into an `AuthorizationContext` with type `DittoAuthorizationContextType.UNSPECIFIED` - which we do not need.

Keeping `AuthorizationContext` param would only make sense if we invoke it with multiple contained `AuthorizationSubject`s.

things/service/src/main/java/org/eclipse/ditto/things/service/utils/ReadGrantCollector.java
```diff
@@ -0,0 +1,151 @@
...
+            final Set<AuthorizationSubject> partialOnly = new HashSet<>(subjectsWithPartialPermission);
+            partialOnly.removeAll(subjectsWithUnrestrictedPermission);
+
+            final JsonObject thingJson = thing.toJson();
+            final JsonValue pointerValue = thingJson.getValue(pointer).orElse(null);
+            
+            if (pointerValue != null && pointerValue.isObject()) {
+                final JsonObject pointerObject = pointerValue.asObject();
+                
+                for (final AuthorizationSubject subject : partialOnly) {
```
thjaeckle, line 126: Hm, instead of iterating over all of the `partialOnly` subjects - wouldn't it make more sense to pass in all `partialOnly` subjects into a single call of `enforcer.getAccessiblePaths(`?

All other parameters are the same for every invocation in the loop ..

things/service/src/main/java/org/eclipse/ditto/things/service/utils/ReadGrantCollector.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param thing the Thing entity
+     * @param policyEnforcer the PolicyEnforcer to query
+     * @return ReadGrant mapping paths to sets of subject IDs
+     */
+    public static ReadGrant collect(
+            final JsonFieldSelector fields,
+            final Thing thing,
+            final PolicyEnforcer policyEnforcer
+    ) {
+        if (fields == null || thing == null || policyEnforcer == null) {
```
thjaeckle: All parameters are not marked with `@Nullable` - so I would expect this can never evaluate to `true`.
If some parameters can be null, they must be declared so ..

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PartialAccessPathCalculator.java
```diff
@@ -0,0 +1,338 @@
...
+     *   }
+     * }
+     * }</pre>
+     *
+     * @param partialAccessPaths the map to convert (subject ID -> list of accessible paths)
+     * @return an indexed JSON object with subjects array and paths mapping to subject indices
+     * @throws NullPointerException if partialAccessPaths is null
+     */
+    public static JsonObject toIndexedJsonObject(final Map<String, List<JsonPointer>> partialAccessPaths) {
+        if (partialAccessPaths == null || partialAccessPaths.isEmpty()) {
```
thjaeckle: `partialAccessPaths` is not `@Nullable`

things/service/src/main/java/org/eclipse/ditto/things/service/utils/PointerUtils.java
```diff
@@ -0,0 +1,54 @@
...
+        // No instantiation
+    }
+
+    /**
+     * Checks if a pointer represents the root path.
+     *
+     * @param pointer the pointer to check
+     * @return true if the pointer is root (empty or "/")
+     */
+    public static boolean isRoot(final JsonPointer pointer) {
```
thjaeckle: Method is not used, can be removed.

Also, `pointer.isEmpty()` is equivalent to `pointer.equals(ROOT_POINTER)` - does not make much sense.

things/service/src/main/java/org/eclipse/ditto/things/service/utils/ReadGrantCollector.java
hu-ahmed: We can't batch all partialOnly subjects into a single getAccessiblePaths() call because we need per-subject results. The method returns Set<JsonPointer> (a union of paths), but we're building Map<JsonPointer, Set<SubjectId>> to track which subjects can access each path. Batching would lose that mapping. We've optimized by extracting resourceKey and readPermissions outside the loop. To batch, the interface would need to return per-subject mappings (e.g., Map<Subject, Set<Paths>>), which would be a larger change.

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/PartialAccessPathResolver.java
```diff
@@ -0,0 +1,284 @@
...
+     * <p>
+     * For external targets, we filter based ONLY on partial-access subjects' paths, ignoring subjects
+     * with full access. This ensures external systems only receive events filtered for partial-access users.
+     * </p>
+     *
+     * @param partialAccessPathsHeader the header value containing partial access paths JSON
+     * @param dittoHeaders the DittoHeaders containing readGrantedSubjects
+     * @return result containing union of all accessible paths from all partial-access subjects
+     */
+    public static AccessiblePathsResult resolveAllAccessiblePathsFromHeader(
```
thjaeckle: @hu-ahmed this method is not used at the moment.  
Would it make sense to remove it then?

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/PartialAccessPathResolver.java
```diff
@@ -0,0 +1,284 @@
...
+     * @param readGrantedSubjects the set of subjects with read granted permissions from headers
+     * @return result containing accessible paths and filtering decision
+     */
+    public static AccessiblePathsResult resolveAccessiblePathsFromHeader(
+            @Nullable final String partialAccessPathsHeader,
+            @Nullable final AuthorizationContext subscriberAuthContext,
+            final Set<AuthorizationSubject> readGrantedSubjects) {
+
+        final Map<String, List<JsonPointer>> partialAccessPaths =
+                parsePartialAccessPathsAsMap(partialAccessPathsHeader);
```
thjaeckle, line 215: @hu-ahmed this is the same independent from the `subscriberAuthContext` or the `readGrantedSubjects`.  
It would therefore make sense to only calculate this Map once for the `partialAccessPathsHeader` and pass it in as an argument - instead of doing it for every `authContext` (e.g. every `target` in a connection) to avoid unnecessary computation.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
```diff
@@ -233,48 +243,189 @@ private List<MappingOutcome<OutboundSignal.Mapped>> processMappableSignals(
...
+
+    /**
+     * Determines if empty filtered payloads should result in dropping the event for a target.
+     * This only applies to Thing events with partial access information present.
+     *
+     * @param adaptable the adaptable already filtered for partial access
+     * @param rootHeaders the root signal headers (may contain PARTIAL_ACCESS_PATHS)
+     * @return true if empty payloads should be treated as "no access"
+     */
+    private static boolean shouldFilterForTarget(final Adaptable adaptable,
```
thjaeckle: @hu-ahmed this method is not used - can it be removed?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
```diff
@@ -233,48 +243,189 @@ private List<MappingOutcome<OutboundSignal.Mapped>> processMappableSignals(
...
-                                    mappingTimer
-                            ));
+                    
+                    final boolean hasPartialAccessPaths = dittoHeaders.containsKey(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
+                    
+                    if (hasPartialAccessPaths && !targets.isEmpty()) {
+                        return processWithPartialAccessPaths(mappableSignal, source, targets, outboundSignal, mappingTimer, mappers);
+                    } else {
+                        return processWithoutPartialAccessPaths(mappableSignal, source, outboundSignal, mappingTimer, mappers);
+                    }
```
thjaeckle, line 271: @hu-ahmed I wonder if we should move this logic to separate classes to keep the control logic in OutboundMappingProcessor cleaner and better understandable.

So e.g. extract to separate classes: `PartialAccessTargetMapping` and `DefaultTargetMapping` implementing a `TargetMapping` interface and call the one or the other here.  

Just an idea to break down complexity when reading the code here ..

thjaeckle: @hu-ahmed hm, the latest addition does not really provide more abstraction - just 2 classes wrapping which method is called.  
I was more thinking in the direction to also move the logic of the methods `processWithoutPartialAccessPaths` and `processWithoutPartialAccessPaths` to separate classes (which then can also be tested easier) and have much less codelines in OutboundMappingProcessor.

The 2 added internal classes make it even more cumbersome to understand I think ..

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessor.java
```diff
@@ -233,48 +244,240 @@ private List<MappingOutcome<OutboundSignal.Mapped>> processMappableSignals(
...
-                            ));
+                    
+                    final boolean hasPartialAccessPaths = dittoHeaders.containsKey(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
+                    
+                    final TargetMappingStrategy mappingStrategy;
+                    if (hasPartialAccessPaths && !targets.isEmpty()) {
+                        final String partialAccessPathsHeader =
+                                dittoHeaders.get(DittoHeaderDefinition.PARTIAL_ACCESS_PATHS.getKey());
+                        final Map<String, List<JsonPointer>> partialAccessPaths =
+                                PartialAccessPathResolver.parsePartialAccessPathsAsMap(partialAccessPathsHeader);
```
thjaeckle: @hu-ahmed this is still done multiple times being in `flatMap()` - where it is the same for all `mappableSignals` - could be done only once before line 247

thjaeckle:
System tests run is green: https://github.com/eclipse-ditto/ditto/actions/runs/20995923480

PR #2293, opened 2025-12-11
fix false positive ERROR log for when "response-required: false" AND condition does not match

thjaeckle:
Yet another corner-case I did not see before the latest Ditto release 🤦


thjaeckle:
I would merge this one - as it only fixes a "false positive" log ERROR

PR #2302, opened 2026-01-13
Add configurable custom search indexes for MongoDB

thjaeckle:
Add support for configuring additional MongoDB indexes for the search collection via HOCON configuration. Custom indexes are created in addition to the built-in indexes and their lifecycle is managed by the existing activated-index-names mechanism.

Features:
- Support compound indexes with multiple fields
- Configurable ASC/DESC direction per field (default: ASC)
- Index name used as configuration key
- Helm chart support via values.yaml
- Documentation in installation-operating.md

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -1699,7 +1699,7 @@ thingsSearch:
         # - "deleteAt"
       # customIndexes holds a map of index-names as key and custom index config as value
       #  Custom indexes are created in addition to the built-in indexes.
-      #  To activate a custom index, add its name to 'activatedIndexNames' above.
+      #  Custom indexes are automatically activated - no need to add them to 'activatedIndexNames'.
       customIndexes: {}
       # my_custom_idx:
       #   fields:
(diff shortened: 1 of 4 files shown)


deployment/helm/ditto/values.yaml
```diff
@@ -1697,6 +1697,16 @@ thingsSearch:
         # - "policyId",
         # - "referencedPolicies",
         # - "deleteAt"
+      # customIndexes holds a map of index-names as key and custom index config as value
+      #  Custom indexes are created in addition to the built-in indexes.
+      #  To activate a custom index, add its name to 'activatedIndexNames' above.
```
alstanchev: I am wondering if defining a 'customIndex' is not enough to make it "Ditto known" and leave the explicit definition in the 'activatedIndexNames' array only if a custom subset of the known indexes is needed.

thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/20989218613

alstanchev (approved):
Other than the one comment on which i don't have strong opinion but it is worth considering, LGTM

deployment/helm/ditto/values.yaml
thjaeckle: @alstanchev yeah, I was also considering that .. did not want to touch the existing mechanism - but as you now also suggest that, I think I should modify it in that way - otherwise mistakes could be made when configuring those custom indexes and not adding them to the activated ones ..

Thanks :)

PR #2307, opened 2026-01-17
add deleteField mapping

hu-ahmed:
In the current Normalized mapper, partial delete events (AttributeDeleted, FeatureDeleted, etc.) are dropped, and merge‑patch nulls are filtered out. This makes it impossible for downstream consumers to distinguish “field never existed” from “field was explicitly deleted.”

Solution
Introduce an opt‑in mapper option includeDeletedFields. When enabled:

Partial delete events are mapped and recorded in _deletedFields.
Merge‑patch nulls are detected before filtering and recorded in _deletedFields.
The default behavior is unchanged unless the option is enabled.
This preserves backward compatibility while allowing delete‑aware ingestion pipelines.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/NormalizedMessageMapper.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/NormalizedMessageMapper.java
@@ -177,7 +177,7 @@ public final class NormalizedMessageMapper extends AbstractMessageMapper {
         }
 
         if (includeDeletedFields) {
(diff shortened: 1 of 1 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/mapping/NormalizedMessageMapper.java
thjaeckle, line 334: the added `includeDeletedFields` should be added to `toString()`

thjaeckle (changes requested):
Thanks @hu-ahmed - great addition when making use of the normalized mapper.

I just have one minor remark where the added boolean is not added in the `toString()` - apart from that the PR LGTM 👍

Issue #2053, opened 2024-10-28
Periodically log "slowest Ditto search queries" in order to find slow queries


thjaeckle:
Idea:
* Find out the slowest RQL queries in order to be able to optimize use of Ditto search
* Consider both, `query` **and** `count` (count is slower than query)
* Log them with a special logger so that they can be easily found and e.g. shown on dashboards
* Enhance log (or MDC) with hints to the calling application (e.g. originator, origin, etc.)

PR #2308, opened 2026-01-20
#2053: add configurable slow query logging to search service

thjaeckle:
Adds a new feature to log slow search queries that exceed a configurable threshold. When enabled, queries exceeding the threshold will be logged with details including duration, namespaces, RQL filter, and the corresponding MongoDB BSON filter.

Configuration:
- ditto.search.query.slow-query-log.enabled (default: true)
- ditto.search.query.slow-query-log.threshold (default: 1s)

Environment variables:
- THINGS_SEARCH_QUERY_SLOW_QUERY_LOG_ENABLED
- THINGS_SEARCH_QUERY_SLOW_QUERY_LOG_THRESHOLD

Fixes: #2053


alstanchev (approved):
LGTM 👍

Issue #2280, opened 2025-11-26
Configure different durations of "activity-check" "inactive-interval" for Things/Policies per namespace


thjaeckle:
This configuration affects how long an "inactive" (neither modified nor accessed) Thing or Policy will be kept in memory by Ditto.

By allowing to make that configurable per namespace, this allows multiple use-cases in Ditto to be configured more efficiently.  
E.g. a namespace containing devices which only report sensor-data once per month could have a really low "inactive-interval" - which would reduce memory consumption.

PR #2309, opened 2026-01-20
#2280: Add per-namespace activity check configuration

thjaeckle:
Makes `activity-check.inactive-interval` and
`deleted-interval `configurable per namespace, allowing different entity passivation times based on namespace patterns.

Resolves: #2280


alstanchev (approved):
LGTM 👍

PR #2313, opened 2026-01-21
Update building and running Ditto with Java 25

thjaeckle:
* update Scala version
* update Pekko version
* update Maven plugins and config where required to work with Java 25
* update GitHub actions dependencies to latest versions
* add -XX:+UseCompactObjectHeaders JVM flag for Helm chart

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -872,6 +872,13 @@ policies:
     persistence:
       # activityCheckInterval configures to keep policies for that amount of time in memory when no other use did happen:
       activityCheckInterval: 2d
+      # namespaceActivityCheckOverrides allows configuring different passivation intervals for policies in specific namespaces
+      #  First match wins - namespace patterns are evaluated in order.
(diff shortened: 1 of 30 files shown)


deployment/helm/ditto/values.yaml
```diff
@@ -116,6 +116,7 @@ global:
     -XX:+UseG1GC
     -XX:+ParallelRefProcEnabled
     -XX:+DisableExplicitGC
+    -XX:+UnlockExperimentalVMOptions
```
kalinkostashki, line 119: shouldn't this be included for java 25 :
-XX:+UseCompactObjectHeaders

kalinkostashki (changes requested):
small remark

deployment/helm/ditto/values.yaml
thjaeckle: As I mentioned in the chat .. I can't add this right now - as the GitHub action is running Ditto `3.8.10` to test the installation of the chart.  
And this will fail if we add `-XX:+UseCompactObjectHeaders` already now ..

kalinkostashki: sorry slipped my mind... :)

Issue #2319, opened 2026-01-27
Support `{{ thing-json:<json-path> }}` placeholder in `/migrateDefinition` migrationPayload


thjaeckle:
### Motivation / Use Case

When migrating a Thing to a new definition using the `POST /things/{thingId}/migrateDefinition` endpoint, the `migrationPayload` currently only accepts static (literal) values. This makes it impossible to dynamically copy or transform existing data from the Thing during migration.

Consider a scenario where a Thing's definition evolves and a field is renamed or moved:
- **Old definition**: Temperature stored at `features/sensor/properties/temp`
- **New definition**: Temperature stored at `features/sensor/properties/status/currentTemperature`

Currently, the migration must either:
1. Provide a hardcoded default value (losing the existing data), or
2. Require a separate API call to read the Thing first, extract the value, then include it in the migration payload

Supporting the `{{ thing-json:<json-path> }}` placeholder in the `migrationPayload` would allow users to reference and copy data from the existing Thing in a single migration request.

### Proposed Solution

Enable placeholder resolution (specifically `thing-json`) in the `migrationPayload` of the `MigrateThingDefinition` command before merging it with the existing Thing.

The `thing-json` placeholder already exists in Ditto (used in Connections for outbound message transformation) and allows extracting arbitrary values from a Thing's JSON representation using JsonPointer notation.

---

## Example

**Existing Thing** (before migration):
```json
{
  "thingId": "org.eclipse.ditto:my-device",
  "definition": "https://models.example.com/sensor-v1.0.0.tm.jsonld",
  "attributes": {
    "manufacturer": "ACME Corp",
    "location": "Building A, Floor 3"
  },
  "features": {
    "sensor": {
      "properties": {
        "temp": 23.5,
        "humidity": 65
      }
    }
  }
}
```

**Migration Request** with placeholder support:
```bash
POST /api/2/things/org.eclipse.ditto:my-device/migrateDefinition
```
```json
{
  "thingDefinitionUrl": "https://models.example.com/sensor-v2.0.0.tm.jsonld",
  "migrationPayload": {
    "features": {
      "sensor": {
        "properties": {
          "status": {
            "currentTemperature": "{{ thing-json:features/sensor/properties/temp }}",
            "currentHumidity": "{{ thing-json:features/sensor/properties/humidity }}",
            "lastKnownLocation": "{{ thing-json:attributes/location }}"
          }
        }
      }
    }
  }
}
```

**Resulting Thing** (after migration):
```json
{
[...]

PR #2321, opened 2026-01-28
#2319 support placeholder replacement in the migrateDefinition

hu-ahmed:
Resolves: #2319


things/service/src/main/java/org/eclipse/ditto/things/service/persistence/actors/strategies/commands/MigrateThingDefinitionStrategy.java
```diff
@@ -299,6 +321,135 @@ private static boolean isEmptyObject(final JsonValue value) {
...
+                    .build();
+        }
+        return value.get();
+    }
+
+    private static String formatThingJsonPlaceholder(final String pointer) {
+        return "{{ " + THING_JSON_PREFIX_COLON + pointer + " }}";
+    }
+
+    private static final class ThingJsonPlaceholder implements Placeholder<JsonObject> {
```
thjaeckle: @hu-ahmed this is basically a copy of `ImmutableThingJsonPlaceholder` from `ditto-edge-service`, right?

We should probably move `ImmutableThingJsonPlaceholder` to the `ditto-placeholders` module, make it work directly on `JsonObject` (to not have a dependency to the thing model) and re-use it also here.

PR #2326, opened 2026-02-01
adding blog about wot-toolings

hu-ahmed:


thjaeckle (approved):
Thanks a lot for the blogpost @hu-ahmed 

I added you to the `authors.yml` (has to be done only once) and did some minor formatting.

Looks really good 👍

PR #2330, opened 2026-02-04
add swagger openIdConnect

hu-ahmed:


deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,235 @@
+(function (global) {
```
thjaeckle, line 12: This file misses a copyright header

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,235 @@
...
+    "use strict";
+
+    const ENVIRONMENTS_URL_PARAM = "environmentsURL";
+    const PRIMARY_ENV_PARAM = "primaryEnvironmentName";
```
thjaeckle, line 18: Are those 2 intentionally the same params as for the Ditto UI?  
If yes, I think we should document that, e.g. as comment ..

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,235 @@
+(function (global) {
+    "use strict";
+
+    const ENVIRONMENTS_URL_PARAM = "environmentsURL";
```
thjaeckle, line 17: I'm not sure if this will work when there is no URL to fetch the environments config from.  
This is not a requirement of the Ditto UI or the ApiDoc to have this URL to fetch the environments from.

thjaeckle (changes requested):
Thanks @hu-ahmed 

As far as I see, this added OIDC integration will only work if there is an URL serving the environment data.  
The Ditto UI however does not rely on such an endpoint (it is an optional thing to configure) - it falls back to local browser storage to load either a default or manually added/edited environments from this local storage.

I think it makes sense to "re-use" the approach of this environment-URL - in order to effectively be able to just have one endpoint for the Ditto UI and the Swagger UI.  
The mechanism however must also work without having such an endpoint.

Could you have another look in that direction, please?

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
hu-ahmed: The integration now also uses localStorage with the same key as Ditto UI (ditto-ui-env) when no environmentsURL is provided. OIDC config is resolved in this order: URL → localStorage, and if neither has OIDC config it falls back to basic auth.

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -35,6 +35,7 @@ tags:
 security:
   - Google:
       - openid
```
thjaeckle: @hu-ahmed we can remove "Google" from here - as we do no longer have the "Ditto Sandbox" where this was used.

documentation/src/main/resources/openapi/ditto-api-2.yml
```diff
@@ -11253,6 +11254,10 @@ components:
           tokenUrl: /oauth2/google/token
           scopes:
             openid: Access your Google-ID
```
thjaeckle: @hu-ahmed we can remove Google also here

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,269 @@
...
+            clientId: provider.client_id || provider.clientId || provider.clientID,
+            scopes
+        };
+    }
+
+    function getEnvFromEnvs(envs, envName) {
+        if (!envs || typeof envs !== "object") {
+            return null;
+        }
+        const name = envName || Object.keys(envs)[0];
```
thjaeckle: @hu-ahmed the problem with this approach is:  
If our environments contain multiple entries, here a random one gets picked.  
I tried locally with the default Ditto environments (or which were still in my local storage) and it chose an environment which does not even have an `oidc` provider configured.

I think we should support that case if no `envName` is provided explicitly.  
But then already in this method choose a `oidc` supporting provider. If we have multiple, we can choose the first one I would say. 👍

documentation/src/main/resources/openapi/sources/api-2-index.yml
```diff
@@ -48,6 +48,7 @@ tags:
 security:
   - Google:
       - openid
```
thjaeckle: can be removed

documentation/src/main/resources/openapi/sources/api-2-index.yml
```diff
@@ -529,6 +530,8 @@ components:
       $ref: "./security/bearerAuth.yml"
     Google:
       $ref: "./security/google.yml"
```
thjaeckle: remove

deployment/docker/swagger3-index.html
```diff
@@ -42,47 +42,38 @@
...
-                     });
-
-        window.ui = ui;
+    window.onload = async function() {
+        const isApidoc = window.location.pathname.startsWith("/apidoc");
+        const openApiPath = isApidoc ? "/apidoc/openapi/ditto-api-2.yml" : "/openapi/ditto-api-2.yml";
+        await SwaggerOidc.initSwaggerUiWithOidc({
+            specUrl: openApiPath,
+            openApiFileName: "ditto-api-2.yml",
+            oauth2RedirectPath: "/oauth2-redirect.html",
```
thjaeckle: @hu-ahmed are you sure that this works? I would assume this has to be `/apidoc/oauth2-redirect.html`?

I tried the flow and it tried to redirect me to `http://localhost:8080/oauth2-redirect.html`

deployment/docker/swagger3-index.html
```diff
@@ -42,47 +42,38 @@
...
+                password: "ditto"
+            },
+            oauth: {
+                mode: "always",
+                useOidcClientId: true,
+                useOidcScopes: true,
+                config: {
+                    clientId: "",
+                    realm: "eclipse-ditto",
+                    appName: "Eclipse Ditto :: Sandbox",
```
thjaeckle: We should remove `:: Sandbox` - as the sandbox no longer exists

deployment/helm/ditto/swaggerui-config/index.html
```diff
@@ -41,34 +41,29 @@
...
-                                       onComplete: function () {
-                                       }
-                                   });
-        window.ui = ui;
+    window.onload = async function() {
+        await SwaggerOidc.initSwaggerUiWithOidc({
+            urls: [{url: "./openapi/ditto-api-2.yml", name:"Ditto API v2"}],
+            primaryName: "Ditto API v2",
+            openApiFileName: "ditto-api-2.yml",
+            oauth2RedirectPath: "/oauth2-redirect.html",
```
thjaeckle: Same here - before the redirect uri was: `oauth2RedirectUrl: window.location.origin + "./oauth2-redirect.html"` - I don't think an absolute `/oauth2-redirect.html` works

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
thjaeckle: Nice, thanks 👍

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,269 @@
...
+            deepLinking: true,
+            presets: [
+                SwaggerUIBundle.presets.apis,
+                SwaggerUIStandalonePreset
+            ],
+            plugins: [
+                SwaggerUIBundle.plugins.DownloadUrl
+            ],
+            layout: "StandaloneLayout",
+            oauth2RedirectUrl: global.location.origin + (opts.oauth2RedirectPath || "/oauth2-redirect.html"),
```
thjaeckle, line 193: @hu-ahmed ah, I understand .. you wanted to add the `/apidoc` path here ..  
This seems however not to work.
For me this resolved to: `http://localhost:8080` - the apidocs locally served at url `http://localhost:8080/apidoc/`

documentation/src/main/resources/openapi/sources/api-2-index.yml
```diff
@@ -535,8 +534,8 @@ components:
       $ref: "./security/nginxBasic.yml"
     Bearer:
       $ref: "./security/bearerAuth.yml"
-    Google:
-      $ref: "./security/google.yml"
```
thjaeckle, line 539: The `google.yml` file should also be removed

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+
+        const authorizationUrl =
+            provider.authorizationUrl ||
+            provider.authorization_url ||
+            (proxyBase ? `${proxyBase}/auth` : (authority ? `${authority}/auth` : null));
+
+        const tokenUrl =
+            provider.tokenUrl ||
+            provider.token_url ||
+            (proxyBase ? `${proxyBase}/token` : (authority ? `${authority}/token` : null));
```
thjaeckle: The `authorizationUrl` should be fetched from the OpenIdConnect `.well-known/openid-configuration` - I did not see where `provider` would expose `authorizationUrl` or `authorization_url`.  
This `provider` type is defined in the Ditto UI used library https://github.com/authts/oidc-client-ts

And the fallback to `${authority}/auth` might work for some OpenID Connect providers, but not for all.

So I think this `authorizationUrl` and `tokenUrl` lookup is wrong

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+
+        const scopes = provider.scopes || provider.scope || "openid";
+
+        return {
+            openIdConnectUrl,
+            authorizationUrl,
+            tokenUrl,
+            clientId: provider.client_id || provider.clientId || provider.clientID,
+            scopes
+        };
```
thjaeckle, line 90: This should return a properly defined TypeScript type - and not `any` as done right now

thjaeckle: Ah, I just noticed - we have plain javascript here :/ 
Then this is not worth introducing Typescript ..

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+            return null;
+        }
+        const provider = env.authSettings.oidc.providers[providerKey];
+        if (!provider) {
+            return null;
+        }
+
+        const metadataUrl = provider.metadataUrl || provider.metadata_url;
+        const authority = provider.authority ? provider.authority.replace(/\/+$/, "") : null;
+        const proxyPath = normalizeProxyPath(provider.proxyPath || provider.proxy_path);
```
thjaeckle: Where does `proxyPath` or `proxy_path` come from? 
I don't see those included in the underlying `OidcClientSettings` type of the  https://github.com/authts/oidc-client-ts library

thjaeckle: `proxyPath` / `proxy_path` — not part of `OidcClientSettings` or any OIDC standard. This is a deployment-specific concern (reverse proxy routing) leaking into the OIDC config contract. If proxy support is needed, it should be handled at a different layer (e.g., the environments config could provide the full `authority` URL including the proxy prefix).

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+        const openIdConnectUrl = metadataUrl
+            ? metadataUrl
+            : (proxyBase
+                ? `${proxyBase}/.well-known/openid-configuration`
+                : (authority ? `${authority}/.well-known/openid-configuration` : null));
+
+        const authorizationUrl =
+            provider.authorizationUrl ||
+            provider.authorization_url ||
+            (proxyBase ? `${proxyBase}/auth` : (authority ? `${authority}/auth` : null));
```
thjaeckle: `authorizationUrl` / `authorization_url` and `tokenUrl` / `token_url` (line 117) belong to the OIDC discovery document (`OidcMetadata`), not to `OidcClientSettings`. The `oidc-client-ts` library discovers these automatically from the `authority`'s `.well-known/openid-configuration` endpoint. Hardcoding fallback construction like `authority + "/auth"` is fragile — real IdPs use paths like `/protocol/openid-connect/auth` (Keycloak) or `/oauth2/v2.0/authorize` (Azure AD).

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+            provider.token_url ||
+            (proxyBase ? `${proxyBase}/token` : (authority ? `${authority}/token` : null));
+
+        const scopes = provider.scopes || provider.scope || "openid";
+
+        return {
+            openIdConnectUrl,
+            authorizationUrl,
+            tokenUrl,
+            clientId: provider.client_id || provider.clientId || provider.clientID,
```
thjaeckle: The canonical field name in `OidcClientSettings` is `client_id` (snake_case). The aliases `clientId` and `clientID` are not part of the contract. Accepting non-standard names means configs that work here would silently fail with the actual `oidc-client-ts` library (and vice versa).

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+    async function resolveOidcConfig() {
+        const params = new URLSearchParams(global.location.search);
+        const environmentsURL = params.get(ENVIRONMENTS_URL_PARAM);
+        const primaryEnvName = params.get(PRIMARY_ENV_PARAM);
+
+        let envs = null;
+
+        if (environmentsURL) {
+            try {
+                const response = await fetch(environmentsURL);
```
thjaeckle: This `fetch` has no timeout. A hanging environments endpoint would block Swagger UI initialization indefinitely since `window.onload` awaits `resolveOidcConfig()`. Consider adding an `AbortController` with a reasonable timeout:
```js
const controller = new AbortController();
const timeoutId = setTimeout(() => controller.abort(), 5000);
try {
    const response = await fetch(environmentsURL, { signal: controller.signal });
    clearTimeout(timeoutId);
    // ...
} catch (e) {
    clearTimeout(timeoutId);
}
```

deployment/helm/ditto/swaggerui-config/swagger-oidc.js
```diff
@@ -0,0 +1,279 @@
...
+                if (response.ok) {
+                    const contentType = response.headers.get("content-type") || "";
+                    if (contentType.includes("application/json")) {
+                        envs = await response.json();
+                    }
+                }
+            } catch (e) {
```
thjaeckle, line 130: All `catch` blocks in this file (here, line 175, 190, 234) silently swallow errors. If OIDC resolution or spec patching fails, there's zero indication for debugging. At minimum, `console.warn("SwaggerOidc: <context>", e)` would help.

thjaeckle (approved):
Fixes LGTM 👍 
Thanks a lot @hu-ahmed for this .. can't wait to try :)

PR #2331, opened 2026-02-05
Replace k6 with Artillery framework

vvasilevbosch:
Replaced the existing k6-based tests with Artillery to improve maintainability and leverage the Node.js ecosystem.

Deprecate k6 load test

Rationale:
- k6 lacks full Node.js compatibility and npm module support out of the box
- Artillery provides native Node.js environment with access to entire npm ecosystem
- Enables use of official Eclipse Ditto JavaScript client for better maintainability
- Simplifies development with standard Node.js tooling

Implementation highlights:
- Uses @eclipse-ditto/ditto-javascript-client-node for HTTP operations
- Uses kafka.js for high-performance Kafka operations(which also fixes a couple of problems the xk6-kafka plugin had)
- Maintains same test scenarios: WARMUP, READ_THINGS, SEARCH_THINGS, MODIFY_THINGS, DEVICE_LIVE_MESSAGES
- Added new scenario channel: Websocket, using @eclipse-ditto/ditto-javascript-client-node client
- Preserves environment-based configuration compatibility with existing test.env
- Added yml based configuration, with env variables override
- Currently supports only concurrent scenario execution
- Docker image available to run in kubernetes environment


benchmark-tool/artillery-test/processor.ts
```diff
@@ -0,0 +1,288 @@
...
+// ABORT FILE HANDLING
+// this is needed, because artillery manages the
+// threads, and it is not possible to pass data between them
+// in a robust manner, to abort the
+// test and exit gracefully
+// ======================
+
+const ABORT_FILE_DIR = '/tmp'
+const ABORT_FILE_NAME = '.artillery_abort_signal';
+const ABORT_FILE = `${ABORT_FILE_DIR}/ABORT_FILE_NAME`;
```
alstanchev: This will create /tmp/ABORT_FILE_NAME (literal string) instead of /tmp/.artillery_abort_signal.
ABORT_FILE_NAME is missing variable sigil and interpolation delimiters.

benchmark-tool/README.md
```diff
@@ -1,3 +1,5 @@
+# Deprecated, reaplced with artillery based load test in [./artillery-test](./artillery-test/)
```
alstanchev: "reaplced" → "replaced"

benchmark-tool/kubernetes/artillery-test.yaml
```diff
@@ -0,0 +1,69 @@
...
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+        - name: artillery
+          image: artillery-test:0.0.1
+          imagePullPolicy: Never
+          env:
+            - name: CONFIG_PATH
+              value: /app/config.yml
```
alstanchev, line 40: Is this the correct config file or the config-kubernetes.yml should be used?

benchmark-tool/artillery-test/common.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * Get the global config instance.
+ * Loads config on first access.
+ */
+export function getConfig(): TestConfig {
+    if (!_config) {
+        _config = loadConfig();
+    }
+    return structuredClone(_config);
```
alstanchev, line 202: Is this deep clone necessary.  It will make deep clone on every VU iteration. At high rate per sec this will cause pressure on GC. Why not just return _config

benchmark-tool/kubernetes/artillery-test.yaml
vvasilevbosch: both configs are pretty much the same, except the addresses of the ditto service, kafka brokers and mmock, but they need to be configured either way in order to work in kubernetes environment, but maybe its best to have a single one, to avoid confusion

alstanchev (approved):

benchmark-tool/artillery-test/common.ts
vvasilevbosch: each worker thread needs copy of this object, deep clone is needed due to ability to override config via ENV variables, so it's one copy per worker thread, not per vu iteration

PR #2344, opened 2026-02-17
Optimize ThingEventEnricher -> TreeBasedPolicyEnforcer hot path

thjaeckle:
Reduce redundant policy tree walks and thing.toJson() serializations in the per-ThingEvent enrichment path, which JFR profiling identified as the No. 1 CPU hotspot on production things nodes.

- Add ClassifySubjectsVisitor that replaces 3 separate tree walks (partial/unrestricted/effected) with a single pass, returning a SubjectClassification result. Expose as classifySubjects() on the Enforcer interface with an optimized override in TreeBasedPolicyEnforcer.

- Add batch getAccessiblePathsForSubjects() that flattens Thing JSON into leaf pointers once per batch instead of once per subject. Default method on Enforcer with an optimized override in TreeBasedPolicyEnforcer.

- Cache thing.toJson() once at the top of enrichWithPredefinedExtraFields() and thread the JsonObject through ReadGrantCollector, PartialAccessPathCalculator, and the extra fields header builder, eliminating redundant serializations per event.

- Extract ROOT_RESOURCE_POINTER constant in TreeBasedPolicyEnforcer to avoid repeated JsonFactory.newPointer("/") allocations.

### DIFF
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/enforcers/tree/TreeBasedPolicyEnforcer.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/enforcers/tree/TreeBasedPolicyEnforcer.java
@@ -49,6 +49,7 @@ import org.eclipse.ditto.policies.model.SubjectId;
 import org.eclipse.ditto.policies.model.Subjects;
 import org.eclipse.ditto.policies.model.enforcers.EffectedSubjects;
 import org.eclipse.ditto.policies.model.enforcers.Enforcer;
+import org.eclipse.ditto.policies.model.enforcers.SubjectClassification;
 
 /**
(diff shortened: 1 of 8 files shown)


policies/model/src/main/java/org/eclipse/ditto/policies/model/enforcers/tree/TreeBasedPolicyEnforcer.java
```diff
@@ -32,6 +32,7 @@
 
 import org.eclipse.ditto.base.model.auth.AuthorizationContext;
 import org.eclipse.ditto.base.model.auth.AuthorizationSubject;
+import org.eclipse.ditto.base.model.auth.DittoAuthorizationContextType;
```
alstanchev: This import seems not used

alstanchev (approved):
👍

Issue #2221, opened 2025-09-09
Enhance policy imports in a way to customise the "subjects" of imported policy entries


thjaeckle:
Right now, [Policy imports](https://eclipse.dev/ditto/basic-policy.html) work in a way, that Policy entries from an imported policy are effectively "copied".  
There is no possibility to customise an imported policy entry, concretising the "subjects" for which the entry shall be applied.

There are however use cases where this would be very useful, e.g. in order to import the `resources` part of a policy entry but define the `subjects` to be applied to this entry in the importing policy.

Example:

```json
{
  "policyId": "org.eclipse.ditto:root-policy",
  "entries": {
    "operator": {
      "subjects": {
      },
      "resources": {
        "thing:/": {
          "grant": ["READ","WRITE"],
          "revoke": []
        }
      },
      "importable": "explicit-additional-subjects" // we could also do "implicit-additional-subjects"
    }
  }
}
```

The importing policy could then import the explicit policy entry as:
```json
{
  "policyId": "org.eclipse.ditto:operator-policy",
  "imports": {
    "org.eclipse.ditto:root-policy" : {
      "entries": [ "operator" ],
      "subjects": {
        "idp:some-group": {
          "type": "Some group as defined in an IdP"
        }
      }
    }
  }
}
```

Or the implicit policy entry as:
```json
{
  "policyId": "org.eclipse.ditto:operator-policy",
  "imports": {
    "org.eclipse.ditto:root-policy" : {
      "subjects": {
        "idp:some-group": {
          "type": "Some group as defined in an IdP"
        }
      }
    }
  }
}
```

The idea would be that this "merges" the subjects from the imported policy (if there are any) with the ones from the importing policy.

PR #2347, opened 2026-02-18
Add entriesAdditions and allowedImportAdditions for policy imports

thjaeckle:
Resolves: #2221

Introduce `entriesAdditions` on policy imports to allow importing policies to additively merge subjects and resources into imported policy entries. Template policies control what can be extended via `allowedImportAdditions` (enum-backed, secure-by-default: empty set means no additions allowed).

New model types: EntryAddition, EntriesAdditions, AllowedImportAddition enum. Write-time validation ensures entriesAdditions labels are declared in entries. Merge-time logic in PolicyImporter silently skips disallowed additions.

The PR also adds HTTP endpoints for:
* `/api/2/policies/{policyId}/entries/{label}/allowedImportAdditions`
* `/api/2/policies/{policyId}/entries/{label}/importable` (was existing before in the policy model, however no extra endpoint was yet provided)
* `/api/2/policies/{policyId}/imports/{importedPolicyId}/entries` (was existing before in the policy model, however no extra endpoint was yet provided)
* `/api/2/policies/{policyId}/imports/{importedPolicyId}/entriesAdditions`

As adding those separate endpoints and providing also model classes to map requests/responses adds a lot of boilerplate code, the PR got quite big..

### DIFF
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/EntriesAdditions.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/EntriesAdditions.java
@@ -60,7 +60,6 @@ public interface EntriesAdditions extends Iterable<EntryAddition>,
      * @param addition the entry addition to set.
      * @return a new EntriesAdditions containing the given addition.
      * @throws NullPointerException if {@code addition} is {@code null}.
-     * @since 3.9.0
      */
     EntriesAdditions setAddition(EntryAddition addition);
 
@@ -70,7 +69,6 @@ public interface EntriesAdditions extends Iterable<EntryAddition>,
      * @param label the label of the addition to remove.
      * @return a new EntriesAdditions without the given addition.
      * @throws NullPointerException if {@code label} is {@code null}.
-     * @since 3.9.0
      */
     EntriesAdditions removeAddition(Label label);
 
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/PolicyResource.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/PolicyResource.java
@@ -18,6 +18,7 @@ import java.util.Optional;
 
 import org.eclipse.ditto.base.model.signals.commands.ResourceMap;
 import org.eclipse.ditto.json.JsonPointer;
+import org.eclipse.ditto.policies.model.EffectedImports;
 import org.eclipse.ditto.policies.model.Policy;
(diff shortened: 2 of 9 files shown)


policies/model/src/main/java/org/eclipse/ditto/policies/model/PolicyImporter.java
```diff
@@ -88,6 +119,39 @@ private static Stream<PolicyEntry> importEntry(final Collection<Label> importedL
...
+    private static Resource mergeResource(final Resource templateResource, final Resource additionalResource) {
+        final EffectedPermissions templatePerms = templateResource.getEffectedPermissions();
+        final EffectedPermissions additionalPerms = additionalResource.getEffectedPermissions();
+
+        final Set<String> mergedGrants = new LinkedHashSet<>(templatePerms.getGrantedPermissions());
+        mergedGrants.addAll(additionalPerms.getGrantedPermissions());
+
+        final Set<String> mergedRevokes = new LinkedHashSet<>(templatePerms.getRevokedPermissions());
+        mergedRevokes.addAll(additionalPerms.getRevokedPermissions());
+
```
alstanchev, line 148: Not sure what the evaluation order of grants and revokes is but an ambiguous policy state is possible here and an access could bе granted that should be revoked
  - Template policy: grants=[READ, WRITE], revokes=[DELETE]
  - Entry addition: grants=[DELETE], revokes=[WRITE]
  - Merged result: grants=[READ, WRITE, DELETE], revokes=[DELETE, WRITE]
Both DELETE and WRITE appear in both grants and revokes - a logical contradiction!

A possible easy fix might be:
mergedGrants.removeAll(mergedRevokes); // Revokes win - safer default

thjaeckle: @alstanchev that is also possible at the moment - and there is the explicit rule that "revoke wins" always ..
I currently don't find it documented - but is that way implemented :)

thjaeckle: So preventing that would be kind of a breaking behavior change .. you can already in a single policy have such scenarios - e.g. one user has directly "READ" granted on a specific attribute, but the user is also in a group which has "READ" revoked on the same attribute --> Then effectively, the user won't be able to READ it.

FMPOV nothing to change - it just now is even more probable to have such situations.

alstanchev (approved):

Issue #2304, opened 2026-01-14
Limit the namespaces which can be accessed via gateway (HTTP / Websocket) by matching a placeholder, e.g. JWT claim


thjaeckle:
In a Ditto setup where several OpenID Connect providers are configured and where multiple tenants should be supported, it is important to keep data separated.  
While Ditto policies already provide a way to do this, an additional "cheap" way to limit access would be to configure that a certain claim in the JWT (e.g. the Issuer `iss`) would limit the namespaces which could be accessed by this user.

A configuration could look like:
```hocon
ditto.gateway.authentication {
    namespace-access = {
      conditions = [ # conditions must all evaluate to true (AND semantics)
        "{{ jwt:iss | fn:filter('like','https://eclipse.org*') }}"
        "{{ header:someheader | fn:filter('ne','dangerous') }}"
      ]
      allowed-namespaces = [
        "org.eclipse.*",
        "another.concrete.namespace"
      ]
      blocked-namespaces = [
        "forbidden.namespace"
      ]
    }
}
```

* Effects of the allowed/blocked namespaces: Directly in the gateway service, 
   * Ditto would reject a direct GET / PUT / PATCH / DELETE  requests on a policy or thing if the namespace of that policy/thing was not allow-listed to be available for the issuer
   * Ditto would reject operations via WebSocket targeting a policy/thing in a non-allowed namespace
   * When doing a **search**, Ditto gateway would
      * automatically inject the `namespaces` to search in into the search query
      * remove non-allowed namespaces from explicitly defined `namespaces` parameter
      * OPEN: Supporting wildcards in Ditto search (defining `namespaces` parameter) is not yet supported - check if this can be provided as part of this issue as well
   * For events (to be published via WebSocket or SSE), Ditto would filter out the non-allowed namespaces, even if they made it to the streaming session based on access control (policies)
* To support placeholders:
   * JWT
   * headers
   * time (could be useful to already configure a "rule" in the future)
* We can make use of [Ditto placeholder function](https://eclipse.dev/ditto/basic-placeholders.html#function-library) `fn:filter` 
   * starting with a supported placeholder
   * when the condition "retains" any matches (fn:filter removes all non-matches)
   * the condition is fulfilled

Summarizing, this feature can add another layer of security / access enforcement
* in both a multi-tenant setup of Ditto
* but also when using namespaces for different things than tenant separation
[...]

PR #2348, opened 2026-02-19
#2304 Add namespace access control at the gateway level

hu-ahmed:
Resolves: #2304

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/directives/auth/NamespaceAccessEnforcementDirective.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/directives/auth/NamespaceAccessEnforcementDirective.java
@@ -27,9 +27,9 @@ import org.eclipse.ditto.gateway.api.NamespaceNotAccessibleException;
 import org.eclipse.ditto.gateway.service.security.authorization.AuthorizationHeaderJwtExtractor;
 import org.eclipse.ditto.gateway.service.security.authorization.NamespaceAccessValidator;
 import org.eclipse.ditto.gateway.service.security.authorization.NamespaceAccessValidatorFactory;
+import org.eclipse.ditto.internal.utils.pekko.logging.DittoLoggerFactory;
+import org.eclipse.ditto.internal.utils.pekko.logging.ThreadSafeDittoLogger;
 import org.eclipse.ditto.jwt.model.JsonWebToken;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Pekko HTTP directive enforcing namespace access control based on JWT claims, HTTP headers,
@@ -37,7 +37,8 @@ import org.slf4j.LoggerFactory;
  */
 public final class NamespaceAccessEnforcementDirective {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(NamespaceAccessEnforcementDirective.class);
+    private static final ThreadSafeDittoLogger LOGGER =
+            DittoLoggerFactory.getThreadSafeLogger(NamespaceAccessEnforcementDirective.class);
 
     private final NamespaceAccessValidatorFactory validatorFactory;
 
@@ -116,7 +117,8 @@ public final class NamespaceAccessEnforcementDirective {
         final NamespaceAccessValidator validator = createValidator(ctx, dittoHeaders, resourceType);
 
         if (!validator.isNamespaceAccessible(namespace)) {
-            LOGGER.debug("Namespace access denied for: {}", namespace);
+            LOGGER.withCorrelationId(dittoHeaders)
+                    .info("Namespace access denied for resource <{}> in namespace <{}>", resourceType, namespace);
             throw NamespaceNotAccessibleException.forNamespace(namespace, dittoHeaders);
         }
     }
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authorization/NamespaceAccessValidator.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authorization/NamespaceAccessValidator.java
@@ -24,14 +24,14 @@ import javax.annotation.concurrent.Immutable;
 import org.eclipse.ditto.base.model.headers.DittoHeaders;
 import org.eclipse.ditto.gateway.service.security.authentication.jwt.JwtPlaceholder;
 import org.eclipse.ditto.gateway.service.util.config.security.NamespaceAccessConfig;
+import org.eclipse.ditto.internal.utils.pekko.logging.DittoLoggerFactory;
+import org.eclipse.ditto.internal.utils.pekko.logging.ThreadSafeDittoLogger;
 import org.eclipse.ditto.jwt.model.JsonWebToken;
 import org.eclipse.ditto.placeholders.ExpressionResolver;
 import org.eclipse.ditto.placeholders.PipelineElement;
 import org.eclipse.ditto.placeholders.PlaceholderFactory;
 import org.eclipse.ditto.placeholders.TimePlaceholder;
 import org.eclipse.ditto.placeholders.UnresolvedPlaceholderException;
-import org.eclipse.ditto.internal.utils.pekko.logging.DittoLoggerFactory;
-import org.eclipse.ditto.internal.utils.pekko.logging.ThreadSafeDittoLogger;
 import org.eclipse.ditto.utils.jsr305.annotations.AllValuesAreNonnullByDefault;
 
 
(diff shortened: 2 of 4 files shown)


gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/directives/auth/NamespaceAccessEnforcementDirective.java
```diff
@@ -0,0 +1,143 @@
...
+            final DittoHeaders dittoHeaders) {
+        return createValidator(ctx, dittoHeaders).getApplicableNamespacePatterns();
+    }
+
+    private NamespaceAccessValidator createValidator(final RequestContext ctx, final DittoHeaders dittoHeaders) {
+        return validatorFactory.createValidator(dittoHeaders, extractJwtFromRequest(ctx));
+    }
+
+    @Nullable
+    private static JsonWebToken extractJwtFromRequest(final RequestContext ctx) {
```
thjaeckle: In Ditto codebase, we don't use `@Nullable` for return types - only for parameters.  
Please change that to `Optional<JsonWebToken>`.  
Previous to the last commit, but `@Nullable` and `Optional` were used.

gateway/service/src/main/resources/gateway.conf
```diff
@@ -294,6 +294,58 @@ ditto {
...
+      # - allowed-namespaces: List of namespace patterns using SQL-LIKE wildcards (* for any chars, ? for single char)
+      #   Examples: "org.eclipse.*", "concrete.namespace"
+      #   Note: "foo.*" matches e.g. "foo.bar" but not exact "foo". To allow both, configure ["foo", "foo.*"].
+      #   Empty list means no restrictions (all namespaces allowed unless blocked)
+      #
+      # - blocked-namespaces: List of namespace patterns to explicitly block (takes precedence over allowed)
+      #   Examples: "forbidden.*", "internal.test.*"
+      #
+      # namespace-access can be configured as either:
+      # - a single object (example below), or
```
thjaeckle: I would not support both - but only the "list of objects" way.  
It is complex to support both, as we also would need to support both in the Helm values.

Please adjust that to only support the "list of objects".

deployment/helm/ditto/values.yaml
```diff
@@ -2384,6 +2384,22 @@ gateway:
...
         # configure the subject to inject in policy action activateTokenIntegration
         tokenIntegrationSubject: "integration:{{policy-entry:label}}:{{jwt:aud}}"
+        # namespaceAccess contains namespace-based access control configuration
+        #  Allows restricting which namespaces can be accessed based on JWT claims or HTTP headers
+        #  This provides an additional enforcement layer before policy-based access control
+        #  conditions must all evaluate to true (AND semantics)
+        #  Search limitation: if clients do not provide explicit "namespaces", wildcard-only allowedNamespaces
+        #  patterns cannot be translated to exact namespace values; gateway injects an empty namespaces restriction.
+        #  Clients should provide explicit namespaces when wildcard-based search matches are desired.
+        # namespaceAccess:
```
thjaeckle, line 2396: The Helm template only suggests how to use the "single object" config.  
As also mentioned in another comment, please adjust the config so that only the "list of objects" is supported.  

We don't need to support both config options.

deployment/helm/ditto/service-config/gateway-extension.conf.tpl
```diff
@@ -27,6 +27,26 @@ ditto {
         }
       }
 
+      {{- if .Values.gateway.config.authentication.namespaceAccess }}
```
thjaeckle, line 30: This must support array notation.  
As mentioned in other comments already, we should not as well support the "single object" variant - but only the array variant.

thjaeckle (changes requested):
Thanks @hu-ahmed for fixing the first round of review comments.

I added some more, mainly in regards to:
* We should not support "both" config variants (single `namespace-access` object OR array as content of `namespace-access`) .. also the Helm template does only support "single object"
  * Please only support the "array" / list option - to define multiple - and drop the "single entry" variant

And another thing I noticed (which was also missing in the issue description):  
We also need to add a configuration if a single namespace-access `conditions` and `allowed-namespaces` and `blocked-namespaces` are defined for `things` or `policies` or for both.

So I would suggest to add: 
```
resource-types = ["thing","policy"]
```

We already have that pattern also in `entity-creation` config.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/streaming/actors/StreamingSessionActor.java
```diff
@@ -167,6 +174,14 @@ private StreamingSessionActor(final Connect connect,
         this.streamingConfig = streamingConfig;
         this.jwtValidator = jwtValidator;
         this.jwtAuthenticationResultProvider = jwtAuthenticationResultProvider;
+        connectionHeaders = connect.getConnectionHeaders();
+        // The namespace access validator is intentionally snapshotted at connection time using the
+        // initial connection headers (including the JWT present at connect). It is NOT updated when
+        // a token refresh occurs mid-session. This means namespace access rules based on JWT claims
+        // reflect the access granted at session establishment, not the refreshed token.
+        this.namespaceAccessValidator = namespaceAccessValidatorFactory != null
+                ? namespaceAccessValidatorFactory.createValidator(connectionHeaders, ThingConstants.ENTITY_TYPE.toString())
```
thjaeckle: @hu-ahmed this only creates a validator for `"thing"` resource type. As the WebSocket however can also be used for policies, this is not sufficient.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authorization/NamespacePatternMatcher.java
```diff
@@ -0,0 +1,88 @@
...
+                        LOGGER.warn("Namespace pattern <{}> could not be converted to regex and will be ignored.", raw);
+                    }
+                    return regex;
+                })
+                .filter(Objects::nonNull)
+                .map(regex -> {
+                    try {
+                        return Pattern.compile(regex);
+                    } catch (final java.util.regex.PatternSyntaxException e) {
+                        LOGGER.warn("Namespace regex <{}> is invalid and will be ignored: {}", regex, e.getMessage());
```
thjaeckle: @hu-ahmed I would not ignore exceptions here - namespace access is a security feature and should fail with a `DittoConfigError` if a wrong syntax was used.
Otherwise, the Ditto operator might think the config was applied successfully (when not checking the logs for warnings).

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/sse/ThingsSseRouteBuilder.java
```diff
@@ -581,6 +594,42 @@ private static Optional<Charset> determineCharsetFromContentType(final Optional<
                 .map(Charset::forName);
     }
 
+    @Nullable
```
thjaeckle: @hu-ahmed please don't use `@Nullable` for return types.  
Ditto codebase always uses an `Optional` for method return types - `@Nullable` only for method parameters or fields.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authorization/NamespaceAccessValidator.java
```diff
@@ -0,0 +1,198 @@
...
+        if (namespaceAccessRules.isEmpty()) {
+            return true;
+        }
+
+        boolean allowedByAnyRule = false;
+
+        for (final NamespaceAccessRule rule : namespaceAccessRules) {
+            if (allConditionsMet(rule.conditions())) {
+                if (rule.matcher().isBlocked(namespace)) {
+                    LOGGER.debug("Namespace '{}' is blocked by namespace-access rule", namespace);
```
thjaeckle: @hu-ahmed please use the `DittoLoggerFactory` for obtaining the `LOGGER` of this class and
* change this to a `info` log
* add the `DittoHeaders` using `.withCorrelationId()`

That way we will have tracing + correlation-id in the log and be able to correlate the log message.

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/security/authorization/NamespaceAccessValidator.java
```diff
@@ -0,0 +1,198 @@
...
+                    return false;
+                }
+                if (rule.matcher().isAllowedByRule(namespace)) {
+                    allowedByAnyRule = true;
+                }
+            }
+        }
+
+        if (!allowedByAnyRule) {
+            LOGGER.debug("Namespace '{}' is not accessible: no rule granted access", namespace);
```
thjaeckle: @hu-ahmed please use the `DittoLoggerFactory` for obtaining the `LOGGER` of this class and
* change this to a `info` log
* add the `DittoHeaders` using `.withCorrelationId()`

That way we will have tracing + correlation-id in the log and be able to correlate the log message.

Issue #2340, opened 2026-02-16
Support rotation of key used for encryption of secrets in connections.


alstanchev:
The Ditto connectivity service encrypts sensitive connection fields (passwords, credentials, URIs) in MongoDB, but **cannot rotate encryption keys** safely. Changing the encryption key requires a long and cumbersome procedure.

  ## Requirements

  ### Must Have

  1. **Dual-Key Configuration**
     - Support current key (`symmetrical-key`) for encrypting new data
     - Support optional old key (`old-symmetrical-key`) for decrypting legacy data
     - Automatic fallback: try current key, then old key on decrypt

  2. **DevOps Migration Commands**
     - `migrateEncryption` - Start/resume batch re-encryption of persisted data
     - `migrateEncryptionStatus` - Query detailed migration progress
     - `migrateEncryptionAbort` - Stop migration with progress preservation

  3. **Migration Features**
     - Dry-run mode to preview impact
     - Resume capability for long-running migrations
     - Progress tracking (which connection, which document, timing)
     - Configurable throttling to control database load

  4. **Encryption Disable Workflow**
     - Support disabling encryption while preserving access to encrypted data
     - Migrate encrypted values back to plaintext

  ### Non-Functional Requirements

  - Zero downtime (except for configuration restart)
  - No data loss during migration
  - Resumable after abort or failure

  ## Proposed Solution

  ### Configuration

  ```hocon
  ditto.connectivity.connection.encryption {
    encryption-enabled = true
    symmetrical-key = "current-key"      # Encrypts new data
    old-symmetrical-key = "legacy-key"   # Decrypts old data

    migration {
      batch-size = 100
      max-documents-per-minute = 100
    }
  }

  Encryption Logic

  - Encrypt: Always use symmetrical-key
  - Decrypt: Try symmetrical-key, fallback to old-symmetrical-key if decryption fails
  - Migration: Explicit trigger by DevOps piggyback command

  Migration API

  Three piggyback commands:

  1. Start/Resume Migration
  {
    "targetActorSelection": "/user/connectivityRoot/encryptionMigration",
    "piggybackCommand": {
      "type": "connectivity.commands:migrateEncryption",
      "dryRun": false,
      "resume": false
    }
  }

  2. Query Status
  {
    "targetActorSelection": "/user/connectivityRoot/encryptionMigration",
    "piggybackCommand": {
      "type": "connectivity.commands:migrateEncryptionStatus"
    }
  }

  Response includes:
[...]

PR #2350, opened 2026-02-23
  Encryption Key Rotation Support

alstanchev:
This PR implements encryption key rotation for Ditto's connectivity service using a dual-key approach. The feature enables secure rotation
  of AES-256-GCM encryption keys for sensitive connection data (credentials, URIs) without service interruption.

  Key Features:
  - Dual-key configuration: Support for current encryption key (symmetrical-key) and fallback key (old-symmetrical-key) with automatic fallback during
  decryption
  - Migration started via piggyback commands: DevOps-triggered migration that re-encrypts existing MongoDB persistence data from old key to new key in configurable
  batches
  - Progress tracking: Migration progress monitoring with resumption support for failed migrations
  - Encryption disable workflow: Safe migration path to decrypt all data and disable encryption entirely
  - Validation and safety: Configuration validation prevents misconfiguration; dry-run mode allows testing before actual migration
  
  Implements #2340

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
@@ -116,11 +116,14 @@ public final class ConnectivityRootActor extends DittoRootActor {
         startChildActor(ConnectionIdsRetrievalActor.ACTOR_NAME, ConnectionIdsRetrievalActor.props(mongoReadJournal,
                 connectionIdsRetrievalConfig));
 
+        final MongoClientWrapper mongoClientWrapper =
+                MongoClientWrapper.newInstance(connectivityConfig.getMongoDbConfig());
+
         startChildActor(ConnectionPersistenceOperationsActor.ACTOR_NAME,
-                ConnectionPersistenceOperationsActor.props(pubSubMediator, connectivityConfig.getMongoDbConfig(),
+                ConnectionPersistenceOperationsActor.props(pubSubMediator, mongoClientWrapper,
                         config, connectivityConfig.getPersistenceOperationsConfig()));
 
-        startEncryptionMigrationSingleton(actorSystem, connectivityConfig);
+        optionallyStartEncryptionMigrationSingleton(actorSystem, connectivityConfig, mongoClientWrapper);
 
         RootChildActorStarter.get(actorSystem, ScopedConfig.dittoExtension(config)).execute(getContext());
 
@@ -156,15 +159,13 @@ public final class ConnectivityRootActor extends DittoRootActor {
         }).build().orElse(super.getSupervisionDecider());
     }
 
-    private void startEncryptionMigrationSingleton(final ActorSystem actorSystem,
-            final ConnectivityConfig connectivityConfig) {
+    private void optionallyStartEncryptionMigrationSingleton(final ActorSystem actorSystem,
+            final ConnectivityConfig connectivityConfig, final MongoClientWrapper mongoClient) {
         final var encryptionConfig = connectivityConfig.getConnectionConfig().getFieldsEncryptionConfig();
         if (encryptionConfig.isEncryptionEnabled() || encryptionConfig.getOldSymmetricalKey().isPresent()) {
-            final MongoClientWrapper mongoClientWrapper =
-                    MongoClientWrapper.newInstance(connectivityConfig.getMongoDbConfig());
             final String managerName = EncryptionMigrationActor.ACTOR_NAME + "Singleton";
             final ActorRef singletonManager = startClusterSingletonActor(
-                    EncryptionMigrationActor.props(connectivityConfig, mongoClientWrapper), managerName);
+                    EncryptionMigrationActor.props(connectivityConfig, mongoClient), managerName);
 
             final ClusterSingletonProxySettings proxySettings =
                     ClusterSingletonProxySettings.create(actorSystem).withRole(CLUSTER_ROLE);
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/ConnectionPersistenceOperationsActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/ConnectionPersistenceOperationsActor.java
@@ -14,7 +14,6 @@ package org.eclipse.ditto.connectivity.service.messaging.persistence;
 
 import org.eclipse.ditto.connectivity.model.ConnectivityConstants;
 import org.eclipse.ditto.internal.utils.persistence.mongo.MongoClientWrapper;
-import org.eclipse.ditto.internal.utils.persistence.mongo.config.MongoDbConfig;
 import org.eclipse.ditto.internal.utils.persistence.mongo.ops.eventsource.MongoEntitiesPersistenceOperations;
 import org.eclipse.ditto.internal.utils.persistence.mongo.ops.eventsource.MongoEventSourceSettings;
 import org.eclipse.ditto.internal.utils.persistence.operations.AbstractPersistenceOperationsActor;
@@ -51,13 +50,13 @@ public final class ConnectionPersistenceOperationsActor extends AbstractPersiste
      * Create Props of this actor.
      *
      * @param pubSubMediator Pekko pub-sub mediator.
-     * @param mongoDbConfig the MongoDB configuration settings.
+     * @param mongoClient the MongoDB client.
      * @param config configuration with info about event journal, snapshot store and database.
      * @param persistenceOperationsConfig the persistence operations configuration settings.
      * @return a Props object.
      */
     public static Props props(final ActorRef pubSubMediator,
-            final MongoDbConfig mongoDbConfig,
+            final MongoClientWrapper mongoClient,
             final Config config,
             final PersistenceOperationsConfig persistenceOperationsConfig) {
 
@@ -67,7 +66,6 @@ public final class ConnectionPersistenceOperationsActor extends AbstractPersiste
                             false, ConnectionPersistenceActor.JOURNAL_PLUGIN_ID,
                             ConnectionPersistenceActor.SNAPSHOT_PLUGIN_ID);
 
-            final MongoClientWrapper mongoClient = MongoClientWrapper.newInstance(mongoDbConfig);
             final MongoDatabase db = mongoClient.getDefaultDatabase();
 
             final EntityPersistenceOperations entitiesOps =
--- a/connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/persistence/ConnectionPersistenceOperationsActorIT.java
+++ b/connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/persistence/ConnectionPersistenceOperationsActorIT.java
@@ -40,6 +40,7 @@ import org.eclipse.ditto.connectivity.service.config.DittoConnectivityConfig;
 import org.eclipse.ditto.connectivity.service.enforcement.ConnectionEnforcerActorPropsFactory;
 import org.eclipse.ditto.internal.utils.config.DefaultScopedConfig;
 import org.eclipse.ditto.internal.utils.config.ScopedConfig;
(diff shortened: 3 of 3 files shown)


connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/migration/EncryptionMigrationActor.java
```diff
@@ -267,610 +314,346 @@ private void handleMigration(final MigrateConnectionEncryption command) {
...
+                                "lastJournalId={}", progress.phase(),
+                        progress.lastProcessedSnapshotId(), progress.lastProcessedJournalId());
                 sender.tell(MigrateConnectionEncryptionResponse.accepted(
                         true, Instant.now().toString(), dryRun, command.getDittoHeaders()), getSelf());
                 return runMigration(progress, oldKey, newKey, pointers, dryRun);
             });
         } else {
-            migrationResult = deleteProgress().thenCompose(v ->
+            migrationResult = progressTracker.deleteProgress().thenCompose(v ->
                     runMigration(new MigrationProgress(), oldKey, newKey, pointers, dryRun));
```
thjaeckle, line 329: `oldKey` and `newKey` might be null and must therefore be specified as `@Nullable` in the `runMigration` method

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/EncryptionMigrationActor.java
```diff
@@ -267,610 +314,346 @@ private void handleMigration(final MigrateConnectionEncryption command) {
...
-            migrationResult = loadProgress().thenCompose(optProgress -> {
-                if (optProgress.isEmpty() || PHASE_COMPLETED.equals(optProgress.get().phase)) {
+            migrationResult = progressTracker.loadProgress().thenCompose(optProgress -> {
+                if (optProgress.isEmpty() || PHASE_COMPLETED.equals(optProgress.get().phase())) {
                     // No previous migration exists or it already completed — nothing to resume
                     final String reason = optProgress.isEmpty()
                             ? "no previous migration found" : "previous migration already completed";
-                    LOG.info("Resume requested but {}, nothing to do", reason);
+                    log.info("Resume requested but {}, nothing to do", reason);
                     migrationInProgress = false;
```
thjaeckle: Still this modifies the actor state on another thread than the actor thread - which might lead to unexpected behavior.

If you are sure this is no problem in this case, it could be acceptable - but shoul be documented then.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/EncryptionMigrationActor.java
```diff
@@ -267,610 +314,346 @@ private void handleMigration(final MigrateConnectionEncryption command) {
...
                     final MigrationProgress completed = optProgress.orElseGet(MigrationProgress::new)
                             .withPhase(PHASE_COMPLETED);
                     currentProgress = completed;
```
thjaeckle: Still this modifies the actor state on another thread than the actor thread - which might lead to unexpected behavior.

If you are sure this is no problem in this case, it could be acceptable - but shoul be documented then.

thjaeckle: Writing `migrationInProgress` and `currentProgress` here races with `handleStatus(), handleAbort(), and handleMigration()` on the actor thread.

thjaeckle: Suggestion: Pipe the "nothing to resume" result back to the actor via `self.tell()` with a dedicated message (similar to `MigrationCompleted`), and handle the state transition on the actor thread.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/EncryptionMigrationActor.java
```diff
@@ -78,50 +105,39 @@ public final class EncryptionMigrationActor extends AbstractActor {
...
     private final MongoCollection<Document> snapshotCollection;
     private final MongoCollection<Document> journalCollection;
-    private final MongoCollection<Document> progressCollection;
+    private final MigrationProgressTracker progressTracker;
     private final int batchSize;
     private final int maxDocumentsPerMinute;
 
     private boolean migrationInProgress = false;
     private boolean currentDryRun = false;
     private volatile boolean abortRequested = false;
```
thjaeckle: What should the `volatile` do in the context of an actor?

Likely the same "smell" of accessing this field not always via the actor thread, but across several threads. 

This pattern screams a little for concurrency issues coming up and is concerning.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
```diff
@@ -117,8 +120,7 @@ private ConnectivityRootActor(final ConnectivityConfig connectivityConfig,
                 ConnectionPersistenceOperationsActor.props(pubSubMediator, connectivityConfig.getMongoDbConfig(),
                         config, connectivityConfig.getPersistenceOperationsConfig()));
 
-        startChildActor(EncryptionMigrationActor.ACTOR_NAME,
-                EncryptionMigrationActor.props(connectivityConfig));
+        startEncryptionMigrationSingleton(actorSystem, connectivityConfig);
```
thjaeckle: `startEncryptionMigrationSingleton()` is called unconditionally, which means every deployment — including those that never use encryption — pays the cost of:
  - A ClusterSingletonManager + ClusterSingletonProxy actor pair
  - A dedicated MongoClientWrapper with its own connection pool

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
```diff
@@ -154,6 +156,21 @@ protected PartialFunction<Throwable, SupervisorStrategy.Directive> getSupervisio
         }).build().orElse(super.getSupervisionDecider());
     }
 
+    private void startEncryptionMigrationSingleton(final ActorSystem actorSystem,
+            final ConnectivityConfig connectivityConfig) {
+        final MongoClientWrapper mongoClientWrapper =
```
thjaeckle: This will create also its own connection pool, based on the min/max config of the service.

This is a little much, couldn't we reuse the existing MongoClientWrapper, via `MongoClientExtension`?

deployment/helm/ditto/templates/connectivity-deployment.yaml
```diff
@@ -350,6 +350,28 @@ spec:
               value: "{{ .Values.connectivity.config.connections.kafka.producer.parallelism }}"
             - name: PEKKO_HTTP_HOSTPOOL_MAX_CONNECTION_LIFETIME
               value: "{{ .Values.connectivity.config.connections.httpPush.maxConnectionLifetime }}"
+            {{- if .Values.connectivity.config.connections.encryption.enabled }}
```
thjaeckle, line 353: ## disable-encryption workflow won't get the old key

The encryption env vars are only set inside `{{- if .Values.connectivity.config.connections.encryption.enabled }}`.  

But the "disable encryption" workflow requires encryption.enabled = false with an old
  key set. A Helm user following the documented disable workflow would need to manually add the old key env var outside the Helm chart, which is error-prone.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
alstanchev: The MongoClientExtension is only in the search service. Do you mean implementing such in the connectivity too or move the current one to a common place and extend it acordingly?

thjaeckle: @alstanchev ah, true .. did not exactly recall where it was, I just remembered that we have it for such a purpose - to e.g. just use one MongoDB connection pool per service instance.

It might be even easier .. `ConnectionPersistenceOperationsActor` calls `MongoClientWrapper.newInstance(mongoDbConfig)` - we could simply move that to `ConnectivityRootActor` and pass it to both the `ConnectionPersistenceOperationsActor` and the EncryptionMigrationSingleton

alstanchev: But isn't the mongo driver underneath using the same connection pool when the same config (same uri) is provided? If not i will do what you proposed and move the call to ConnectivityRootActor

thjaeckle: Could be, but I am not sure about that - did you find that documented in the mongo driver?

alstanchev: Actually i am wrong, doc states that each Client creates new pool despite of config.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
```diff
@@ -150,6 +156,24 @@ protected PartialFunction<Throwable, SupervisorStrategy.Directive> getSupervisio
         }).build().orElse(super.getSupervisionDecider());
     }
 
+    private void startEncryptionMigrationSingleton(final ActorSystem actorSystem,
+            final ConnectivityConfig connectivityConfig) {
+        final var encryptionConfig = connectivityConfig.getConnectionConfig().getFieldsEncryptionConfig();
+        if (encryptionConfig.isEncryptionEnabled() || encryptionConfig.getOldSymmetricalKey().isPresent()) {
+            final MongoClientWrapper mongoClientWrapper =
+                    MongoClientWrapper.newInstance(connectivityConfig.getMongoDbConfig());
```
thjaeckle: @alstanchev the idea was to create the `MongoClientWrapper` once in `ConnectivityRootActor` and also pass it to `ConnectionPersistenceOperationsActor` (line 120) as well.

That way we guaranteed only have one instance and a shared connection pool.

Could you adjust that, please?

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
```diff
@@ -116,6 +120,8 @@ private ConnectivityRootActor(final ConnectivityConfig connectivityConfig,
                 ConnectionPersistenceOperationsActor.props(pubSubMediator, connectivityConfig.getMongoDbConfig(),
                         config, connectivityConfig.getPersistenceOperationsConfig()));
 
+        startEncryptionMigrationSingleton(actorSystem, connectivityConfig);
```
thjaeckle: @alstanchev I am still wondering if we need to start the singleton actor even when e.g. encryption is disabled in the config.

Is this to handle the case "currently encrypted - no new key provided -> will migrate all encrypted data to be available decrypted"?

alstanchev: It has to start in the case where encryption is disabled but old key is configured which is considered a disable flow. So the actor is need to decrypt. It will not be started in the case where encryption is disabled and old key is not configured. This check is inside the startEncryptionMigrationSingleton() method.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
alstanchev: I got in to refactoring other stuff and thought i did that one.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/ConnectivityRootActor.java
thjaeckle: Aaah, yeah .. all good then 👍 
Thanks :)

PR #2355, opened 2026-02-26
Optimize MongoReadJournal pipelines and bump dependency versions

thjaeckle:
- Extract inline BsonDocument.parse() calls into static final constants (FILTER_PRIORITY_TAGS_EXPR, DOCDB_MAP_PRIORITY_TAGS_EXPR) to avoid re-parsing on every ping interval
- Add $project stages before $sort/$group to reduce document size in memory
- Replace recursive $redact with targeted $addFields/$filter on the items array in listNewestActiveSnapshotsByBatch; skip stage entirely when includeDeleted=true
- Enable snapshot aggregation index (pid+sn+id) by default for MongoDB 6+
- Update pekko-persistence-mongodb 1.3.1 -> 1.4.0
- Update mongo-java-driver 5.6.0 -> 5.6.4
- Update netty 4.2.4.Final -> 4.2.10.Final
- Fix IndexInitializationIT to match activated index configuration

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -45,7 +45,7 @@
 
         <pekko-bom.version>1.4.0</pekko-bom.version>
         <pekko-http-bom.version>1.3.0</pekko-http-bom.version>
-        <pekko-persistence-mongodb.version>1.3.1</pekko-persistence-mongodb.version>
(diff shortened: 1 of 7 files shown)


alstanchev (approved):
👍

PR #2356, opened 2026-02-26
support checkPermissions for all protocols

hu-ahmed:
The checkPermissions API was previously only reachable via the HTTP gateway (POST /api/v2/checkPermissions). This meant that devices or backend services communicating with Ditto through other protocols — such as WebSocket, AMQP, or MQTT — had no way to use it, even
  though permission checking is a cross-cutting concern that makes sense regardless of the transport layer.

  Changes

  To make checkPermissions protocol-agnostic, the following was done:

  - Relocated core classes (CheckPermissions, CheckPermissionsResponse, ImmutablePermissionCheck) from gateway.service.endpoints.routes.checkpermissions into base.api.common.checkpermissions, making them available to the full Ditto stack instead of being
  gateway-specific.
  - Added Ditto Protocol support by introducing the necessary adapters, signal mappers, and mapping strategies (CheckPermissionsCommandAdapter, CheckPermissionsCommandResponseAdapter, CheckPermissionsSignalMapper, CheckPermissionsCommandMappingStrategies, etc.). The
  command is handled on the topic path _/_/common/commands/checkPermissions.
  - Integrated into the edge service via CheckPermissionsActor and InboundDispatchingSink, so the command can be dispatched and processed independently of whether it arrives over HTTP, WebSocket, or a connectivity protocol.

### DIFF
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissions.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissions.java
@@ -36,6 +36,7 @@ import org.eclipse.ditto.json.JsonPointer;
 import org.eclipse.ditto.json.JsonValue;
 import org.eclipse.ditto.policies.model.PolicyId;
 import org.eclipse.ditto.policies.model.signals.commands.PolicyCommand;
+import org.eclipse.ditto.policies.model.signals.commands.query.PolicyQueryCommand;
 
 /**
  * Command to check multiple permissions for different resources in a single request.
@@ -45,7 +46,7 @@ import org.eclipse.ditto.policies.model.signals.commands.PolicyCommand;
 @Immutable
 @JsonParsableCommand(typePrefix = PolicyCommand.TYPE_PREFIX, name = CheckPermissions.NAME)
 public final class CheckPermissions extends AbstractCommand<CheckPermissions> implements
-        PolicyCommand<CheckPermissions> {
+        PolicyQueryCommand<CheckPermissions> {
 
     /**
      * The type of the command.
@@ -60,7 +61,7 @@ public final class CheckPermissions extends AbstractCommand<CheckPermissions> im
     private static final JsonFieldDefinition<JsonObject> PERMISSION_CHECKS_FIELD = JsonFactory.newJsonObjectFieldDefinition(
             "permissionChecks", FieldType.REGULAR, JsonSchemaVersion.V_2
     );
-    private static final PolicyId ENTITY_ID = PolicyId.of("ditto:check-permissions");
+    private static final PolicyId ENTITY_ID = PolicyId.of(":check-permissions");
 
     private final Map<String, ImmutablePermissionCheck> permissionChecks;
 
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissionsResponse.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissionsResponse.java
@@ -38,7 +38,7 @@ import org.eclipse.ditto.json.JsonObjectBuilder;
 import org.eclipse.ditto.json.JsonPointer;
 import org.eclipse.ditto.json.JsonValue;
 import org.eclipse.ditto.policies.model.PolicyId;
-import org.eclipse.ditto.policies.model.signals.commands.PolicyCommandResponse;
+import org.eclipse.ditto.policies.model.signals.commands.query.PolicyQueryCommandResponse;
 
 /**
  * Response for a {@link CheckPermissions} command in the Eclipse Ditto framework.
@@ -53,7 +53,7 @@ import org.eclipse.ditto.policies.model.signals.commands.PolicyCommandResponse;
 @Immutable
 @JsonParsableCommandResponse(type = CheckPermissionsResponse.TYPE)
 public final class CheckPermissionsResponse extends AbstractCommandResponse<CheckPermissionsResponse>
-        implements WithEntity<CheckPermissionsResponse>, PolicyCommandResponse<CheckPermissionsResponse> {
+        implements WithEntity<CheckPermissionsResponse>, PolicyQueryCommandResponse<CheckPermissionsResponse> {
 
     /**
      * The type identifier for the response.
@@ -67,7 +67,7 @@ public final class CheckPermissionsResponse extends AbstractCommandResponse<Chec
             JsonFieldDefinition.ofJsonObject("permissionResults", FieldType.REGULAR, JsonSchemaVersion.V_2);
 
     private static final HttpStatus HTTP_STATUS = HttpStatus.OK;
-    private static final PolicyId ENTITY_ID = PolicyId.of("ditto:check-permissions");
+    private static final PolicyId ENTITY_ID = PolicyId.of(":check-permissions");
 
     private final JsonObject permissionResults;
 


thjaeckle:
@hu-ahmed could you please add some description to the PR? I know what it is about - but a little more context would be helpful for the PR and why this is added

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
```diff
@@ -262,6 +263,7 @@ protected BaseClientActor(final Connection connection,
         // log the default client ID for tracing
         logger.info("Using default client ID <{}>", getDefaultClientId());
 
+        this.commandForwarderActor = commandForwarderActor;
```
thjaeckle: Why did you add this additionally as `ActorRef`?  
It is already available as `commandForwarderActorSelection` - which we can also use to send messages to or use `Patterns.ask`.

protocol/src/main/java/org/eclipse/ditto/protocol/TopicPath.java
```diff
@@ -224,7 +224,14 @@ enum Group {
...
          */
-        CONNECTIONS("connections", ConnectivityConstants.ENTITY_TYPE);
+        CONNECTIONS("connections", ConnectivityConstants.ENTITY_TYPE),
+
+        /**
+         * Common group (no entity type).
+         *
+         * @since 3.9.0
+         */
+        COMMON("common", null);
```
thjaeckle: This is now unused and can be removed again

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -33,14 +34,20 @@
...
  *
  * @since 3.7.0
  */
 @Immutable
 public final class ImmutablePermissionCheck implements Jsonifiable<JsonObject> {
 
-    private final ResourceKey resource;
+    private static final String POLICY_RESOURCE_PREFIX = "policy:";
+
+    private final String resource;
```
thjaeckle: Now that `ImmutablePermissionCheck` is in the policy model, it should use the type `ResourceKey` instead of plain String here.

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -51,36 +58,40 @@ public final class ImmutablePermissionCheck implements Jsonifiable<JsonObject> {
...
      *
-     * @param resourceKey the resourceKey for which permissions are being checked.
+     * @param resource the resource string in {@code "resourceType:resourcePath"} format
+     *                 (e.g., {@code "thing:/features"}, {@code "policy:/"}).
      * @param entityId the entity ID associated with the resource.
      * @param hasPermissions the list of permissions being checked.
      * @return a new {@link ImmutablePermissionCheck} instance.
      */
-    public static ImmutablePermissionCheck of(final ResourceKey resourceKey, final String entityId,
+    public static ImmutablePermissionCheck of(final String resource, final String entityId,
```
thjaeckle: Now that `ImmutablePermissionCheck` is in the policy model, it should keep using the type `ResourceKey` instead of plain String here.

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/ImmutablePermissionCheck.java
```diff
@@ -51,36 +58,40 @@ public final class ImmutablePermissionCheck implements Jsonifiable<JsonObject> {
...
+     * in {@code "resourceType:resourcePath"} format (e.g., {@code "thing:/features"}).
+     * <p>
+     * Callers that need a typed {@code ResourceKey} can convert via
+     * {@code ResourceKey.newInstance(getResource())}.
      *
-     * @return the resource ResourceKey.
+     * @return the resource string.
      */
-    public ResourceKey getResourceKey() {
+    public String getResource() {
```
thjaeckle: Should return type `ResourceKey`

gateway/service/src/main/java/org/eclipse/ditto/gateway/service/streaming/actors/StreamingSessionActor.java
```diff
@@ -312,33 +285,21 @@ private Receive createIncomingSignalBehavior() {
                 .match(CommandResponse.class, this::forwardAcknowledgementOrLiveCommandResponse)
                 .match(ThingSearchCommand.class, this::forwardSearchCommand)
                 .match(StreamingSubscriptionCommand.class, this::forwardStreamingSubscriptionCommand)
+                .match(CheckPermissions.class, checkPermissions -> {
+                    final Duration timeout = checkPermissions.getDittoHeaders().getTimeout()
+                            .orElse(Duration.ofSeconds(60));
```
thjaeckle: The timeout should not be a "magic number" in the middle of the code
* either extract a constant of the class - with a proper name (I think I would do that)
* or make configurable

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/InboundDispatchingSink.java
```diff
@@ -547,6 +552,13 @@ private int dispatchIncomingSignal(final IncomingSignal incomingSignal) {
             }
 
             return 1;
+        } else if (signal instanceof CheckPermissions checkPermissions) {
+            final Duration timeout = checkPermissions.getDittoHeaders().getTimeout()
+                    .orElse(Duration.ofSeconds(60));
```
thjaeckle: The timeout should not be a "magic number" in the middle of the code
* either extract a constant of the class - with a proper name (I think I would do that)
* or make configurable

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissions.java
```diff
@@ -39,14 +42,14 @@
  */
 @Immutable
 @JsonParsableCommand(typePrefix = CheckPermissions.TYPE_PREFIX, name = CheckPermissions.NAME)
-public final class CheckPermissions extends CommonCommand<CheckPermissions> {
+public final class CheckPermissions extends AbstractCommand<CheckPermissions> {
```
thjaeckle: `CheckPermissions` should now implement `PolicyCommand` - as we moved it to the policy module.

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissionsResponse.java
```diff
@@ -50,9 +51,11 @@
  */
 @Immutable
 @JsonParsableCommandResponse(type = CheckPermissionsResponse.TYPE)
-public final class CheckPermissionsResponse extends CommonCommandResponse<CheckPermissionsResponse>
+public final class CheckPermissionsResponse extends AbstractCommandResponse<CheckPermissionsResponse>
```
thjaeckle, line 55: `CheckPermissionsResponse` should now implement `PolicyCommandResponse` - as we moved it to the policy module.

protocol/src/main/java/org/eclipse/ditto/protocol/adapter/ProtocolAdapter.java
```diff
@@ -131,7 +133,9 @@ static TopicPath.Channel determineChannel(final Signal<?> signal) {
      * @return the default channel determined from the signal
      */
     static TopicPath.Channel determineDefaultChannel(final Signal<?> signal) {
-        if (signal instanceof PolicyCommand || signal instanceof PolicyCommandResponse ||
+        if (signal instanceof CheckPermissions || signal instanceof CheckPermissionsResponse) {
```
thjaeckle: This should no longer be needed if `CheckPermissions` implements `PolicyCommand` and `CheckPermissionsResponse` implements `PolicyCommandResponse`

protocol/src/main/java/org/eclipse/ditto/protocol/TopicPath.java
```diff
@@ -255,6 +262,13 @@ public String getName() {
             return name;
         }
 
+        /**
+         * Returns the entity type for this group, or {@code null} for groups that have no associated
+         * entity type (e.g., {@link #COMMON}).
+         *
+         * @return the entity type, or {@code null}.
+         */
+        @Nullable
```
thjaeckle: `@Nullable` should be removed again - as we also remove `COMMON` enum which needed it to be `null`

policies/model/src/main/java/org/eclipse/ditto/policies/model/signals/commands/checkpermissions/CheckPermissions.java
```diff
@@ -39,14 +42,14 @@
...
 
-    static final String TYPE_PREFIX = CommonCommand.TYPE_PREFIX;
+    static final String TYPE_PREFIX = "policies." + TYPE_QUALIFIER + ":";
 
     /**
      * The name of the command.
      */
-    static final String NAME = "checkPermissions";
+    public static final String NAME = "checkPermissions";
```
thjaeckle: this can stay package private

Issue #1638, opened 2023-05-16
Configure certain policies to be always imported by all policies of a namespace


thjaeckle:
Ditto shall support configuring that per namespace a declared set of policy IDs are always imported, acting as kind of "namespace/tenant root policy".

It is sufficient to configure that statically, requiring a Ditto service restart for changes to be applied.

PR #2367, opened 2026-03-10
Support namespace root policy

hu-ahmed:
Resolves: #1638

  ## Summary

  This PR adds support for namespace root policies in Ditto policy enforcement, including wildcard-based namespace mappings.

  A namespace can be mapped to one or more root policy IDs. During enforcer creation, Ditto transparently merges entries from those root policies into policies of that namespace.

  ## What changed

  - Added namespace-root merge support in policy enforcer resolution.
  - Added config abstraction:
    - `NamespacePoliciesConfig`
    - `DefaultNamespacePoliciesConfig`
  - Added support for namespace policy patterns:
    - exact namespace: `org.example.devices`
    - prefix wildcard: `org.example.devices.*`
    - catch-all: `*`
  - Added deterministic precedence for overlapping patterns:
    - exact namespace
    - more specific prefix wildcard
    - broader prefix wildcard
    - catch-all `*`
  - Wired namespace policy resolution into:
    - cache loader path
    - create-policy enforcement path (`PolicyEnforcerActor`)
  - Extended cache invalidation:
    - when a namespace root policy changes, cached policies in covered namespaces are invalidated
  - Added canonical base config file:
    - `internal/utils/config/.../ditto-namespace-policies.conf`
    - included via `ditto-service-base.conf`
  - Updated Helm templates and values to service-scoped config:
    - `policies.config.namespacePolicies`
    - `things.config.namespacePolicies`
  - Updated chart docs accordingly.
  - Added config validation for unsupported wildcard syntax at startup.

  ## Behavior / rules

  - Only entries with `importable = "implicit"` are merged.
  - Entries with `importable = "explicit"` or `importable = "never"` are not merged.
  - Local policy entries win on label conflicts.
  - If multiple namespace root policies match, they are applied in deterministic precedence order:
    1. exact namespace
    2. more specific wildcard prefix
    3. broader wildcard prefix
    4. `*`
  - If a configured root policy is missing or deleted, entries are skipped and an error is logged.
  - Stored policy JSON is not modified; merging happens only at enforcer-build time.
  - Unsupported namespace policy patterns are rejected at config load time.

  ## Supported config syntax

  - `org.example.devices`
  - `org.example.devices.*`
  - `*`

  Unsupported examples:

  - `org.*.devices`
  - `foo*`
  - `**`

  ## Example config

  ```yaml
  policies:
    config:
      namespacePolicies:
        org.example.devices:
          - org.example:tenant-root-exact
        org.example.devices.*:
          - org.example:tenant-root-devices
        org.example.*:
          - org.example:tenant-root-general

  things:
    config:
      namespacePolicies:
        org.example.devices:
          - org.example:tenant-root-exact
        org.example.devices.*:
          - org.example:tenant-root-devices
        org.example.*:
          - org.example:tenant-root-general

### DIFF
--- a/internal/utils/config/src/main/resources/ditto-namespace-policies.conf
+++ b/internal/utils/config/src/main/resources/ditto-namespace-policies.conf
@@ -1,0 +1,26 @@
+# namespace root policies
+# Maps namespace patterns to lists of policy IDs whose implicit entries are transparently merged
+# into every matching policy's enforcer at build time.
+#
+# Pattern syntax (key side):
+#  - exact string      e.g. "org.example.devices"  matches only that namespace
+#  - prefix wildcard   e.g. "org.example.*"        matches any namespace starting with "org.example."
+#  - catch-all         "*"                         matches every namespace
+#
+# Rules:
+#  - only entries with importable = "implicit" are merged; "explicit" and "never" are skipped.
+#  - local policy entries always win on label conflicts (namespace root cannot override local entries)
+#  - matching patterns are applied in this precedence: exact > more specific prefix wildcard > less specific prefix wildcard > "*"
+#  - unsupported wildcard syntax is rejected at config load time
+#  - if a configured root policy is missing or deleted, its entries are skipped and an ERROR is logged
+#  - changing a root policy triggers cache invalidation for all cached policies in matching namespaces
+#  - this config is read by all services that perform policy enforcement (policies, things, thingsearch)
+#
+# For Helm deployments, configure this via the global.namespacePolicies value,
+# which is rendered into the respective service extension config files.
+ditto.namespace-policies {
+//  "org.example.devices"  = ["org.example:tenant-root"]           # exact namespace
+//  "org.example.devices.*" = ["org.example:devices-root"]         # more specific prefix wildcard
+//  "org.example.*"         = ["org.example:tenant-root"]          # broader prefix wildcard
+//  "*"                     = ["root:catch-all-policy"]            # every namespace
+}
--- a/policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcerCache.java
+++ b/policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcerCache.java
@@ -27,6 +27,7 @@ import org.eclipse.ditto.internal.utils.cache.Cache;
 import org.eclipse.ditto.internal.utils.cache.CacheFactory;
 import org.eclipse.ditto.internal.utils.cache.config.CacheConfig;
 import org.eclipse.ditto.internal.utils.cache.entry.Entry;
+import org.eclipse.ditto.policies.enforcement.config.NamespacePoliciesConfig;
 import org.eclipse.ditto.policies.model.Policy;
 import org.eclipse.ditto.policies.model.PolicyId;
 import org.eclipse.ditto.policies.model.PolicyImport;
@@ -39,11 +40,14 @@ final class PolicyEnforcerCache implements Cache<PolicyId, Entry<PolicyEnforcer>
 
     private final Cache<PolicyId, Entry<PolicyEnforcer>> delegate;
     private final Map<PolicyId, Set<PolicyId>> policyIdToImportingMap;
+    private final NamespacePoliciesConfig namespacePoliciesConfig;
 
     PolicyEnforcerCache(final AsyncCacheLoader<PolicyId, Entry<PolicyEnforcer>> policyEnforcerCacheLoader,
             final ExecutionContextExecutor cacheDispatcher,
-            final CacheConfig cacheConfig) {
+            final CacheConfig cacheConfig,
+            final NamespacePoliciesConfig namespacePoliciesConfig) {
         policyIdToImportingMap = new ConcurrentHashMap<>();
+        this.namespacePoliciesConfig = namespacePoliciesConfig;
         this.delegate = CacheFactory.createCache(
                 (policyId, executor) -> policyEnforcerCacheLoader.asyncLoad(policyId, executor)
                         .whenCompleteAsync(((policyEnforcerEntry, throwable) -> policyEnforcerEntry.get()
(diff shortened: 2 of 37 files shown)


policies/service/src/main/java/org/eclipse/ditto/policies/service/persistence/actors/PolicyEnforcerActor.java
```diff
@@ -39,13 +41,16 @@
...
+    private final NamespacePoliciesConfig namespacePoliciesConfig;
+
     @SuppressWarnings("unused")
     private PolicyEnforcerActor(final PolicyId policyId,
             final PolicyCommandEnforcement policyCommandEnforcement,
             final PolicyEnforcerProvider policyEnforcerProvider,
             final LocalAskTimeoutConfig localAskTimeoutConfig
     ) {
         super(policyId, policyCommandEnforcement, policyEnforcerProvider, localAskTimeoutConfig);
+        this.namespacePoliciesConfig = DefaultNamespacePoliciesConfig.of(getContext().system().settings().config());
```
thjaeckle: ### HIGH: Config parsed per actor instance

`DefaultNamespacePoliciesConfig.of()` parses the HOCON config on every `PolicyEnforcerActor` construction. Since there is one actor per active policy entity, this creates many identical immutable objects.

The config is static and the same for the entire actor system. Consider either:
- Passing the pre-parsed config as a `Props` constructor parameter (parsed once in the `props()` factory method)
- Or caching it as a `static` lazy singleton

The `CachingPolicyEnforcerProvider` already does this correctly (parses once), so this actor is the outlier.

policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/config/NamespacePoliciesConfig.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param pattern the pattern from the config (exact, prefix wildcard, or {@code *}).
+     * @return {@code true} if {@code namespace} matches {@code pattern}.
+     * @since 3.9.0
+     */
+    static boolean namespaceMatchesPattern(final String namespace, final String pattern) {
+        if ("*".equals(pattern)) {
+            return true;
+        }
+        if (pattern.endsWith(".*")) {
+            return namespace.startsWith(pattern.substring(0, pattern.length() - 1));
```
thjaeckle, line 112: ### MEDIUM: Two edge cases worth documenting and testing

**1. `org.example.*` does NOT match `org.example` itself**
`startsWith("org.example.")` requires at least one segment after the prefix. An operator configuring `org.example.*` might expect it to cover `org.example` too. This is probably the intended behavior ("sub-namespaces of"), but it should be explicitly documented and tested.

**2. Multi-level sub-namespaces match implicitly**
`org.example.*` also matches `org.example.devices.sensors.temp` (arbitrary depth), because `startsWith` doesn't restrict to a single level. The Javadoc says "matches any namespace starting with `org.example.`" which is accurate, but consider adding a note that this is not single-level-only — operators coming from glob semantics might expect `.*` to mean one level.

Suggested test additions:
```java
assertThat(namespaceMatchesPattern("org.example", "org.example.*")).isFalse();
assertThat(namespaceMatchesPattern("org.example.a.b.c", "org.example.*")).isTrue();
```

policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcerCache.java
```diff
@@ -92,35 +96,63 @@ public Optional<Entry<PolicyEnforcer>> getBlocking(final PolicyId key) {
...
+        if (!namespacePoliciesConfig.getAllNamespaceRootPolicyIds().contains(policyId)) {
+            return false;
+        }
+
+        final Set<String> affectedNamespaces = namespacePoliciesConfig.getNamespacesForRootPolicy(policyId);
+        if (affectedNamespaces.isEmpty()) {
+            return false;
+        }
+
+        return delegate.asMap().keySet().stream()
```
thjaeckle, line 167: ### MEDIUM: Full cache iteration on namespace root invalidation

`delegate.asMap().keySet().stream()` iterates **every** cached policy entry and checks each against the affected patterns. This is O(n) where n = total cache size, executed on every change to a namespace root policy.

For typical deployments this is fine, but worth a brief comment noting the trade-off. In large-scale deployments (100k+ cached policies with frequently-changing root policies), consider tracking namespace membership at cache-load time (e.g., a reverse `Map<String, Set<PolicyId>>` for namespace → cached policies) to make this O(affected) instead of O(total).

policies/enforcement/src/test/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcerNamespacePoliciesTest.java
```diff
@@ -0,0 +1,249 @@
...
+    private static NamespacePoliciesConfig namespaceConfig(final PolicyId rootPolicyId) {
+        final NamespacePoliciesConfig config = mock(NamespacePoliciesConfig.class);
+        when(config.isEmpty()).thenReturn(false);
+        when(config.getRootPoliciesForNamespace(NAMESPACE)).thenReturn(Collections.singletonList(rootPolicyId));
+        when(config.getAllNamespaceRootPolicyIds()).thenReturn(Set.of(rootPolicyId));
+        when(config.getNamespacesForRootPolicy(rootPolicyId)).thenReturn(Set.of(NAMESPACE));
+        return config;
+    }
+
+}
```
thjaeckle, line 315: ### MEDIUM: Missing test for root policies with their own explicit imports

The merge code calls `rootPolicyOpt.get().withResolvedImports(policyResolver)` — resolving the root policy's own explicit imports before merging its entries. But no test verifies this scenario.

Suggested test: root policy A has an explicit import from policy B. B has an implicit entry `B_READER`. Verify that after merge, the child policy's enforcer includes both A's implicit entries **and** the entries A imported from B.

Also missing:
- Duplicate `PolicyId` appearing in both an exact and wildcard pattern (verifies `distinct()` dedup + precedence order)
- Root policy deletion triggering cache invalidation (end-to-end with `PolicyEnforcerCache`)

internal/utils/config/src/main/resources/ditto-namespace-policies.conf
```diff
@@ -0,0 +1,27 @@
...
+#  - matching patterns are applied in this precedence: exact > more specific prefix wildcard > less specific prefix wildcard > "*"
+#  - unsupported wildcard syntax is rejected at config load time
+#  - if a configured root policy is missing or deleted, its entries are skipped and an ERROR is logged
+#  - changing a root policy triggers cache invalidation for all cached policies in matching namespaces
+#  - this config is read by all services that perform policy enforcement (policies, things)
+#
+# For Helm deployments, configure this via the policies.config.namespacePolicies and
+# things.config.namespacePolicies values, which are rendered into the respective
+# service extension config files.
+ditto.policies.namespace-policies {
```
thjaeckle: I don't think this is a good idea to put it in a config file defined here and with key `ditto.policies`.  
`ditto.policies` contains policies service related configuration (in file `policies.conf`).

And this is a little confusing then to have that key also in "things" configuration. 

Plus, this file would also be imported for all other Ditto services which do not make use of this (gateway, connectivity).

I would explicitly add the imports to `policies.conf` and `things.conf`.

PR #2368, opened 2026-03-12
#2325 add namespace support for policies

### LINKED ISSUE #2325: Add optional `namespaces` field to policy entries for namespace-scoped authorization
## Summary

Add an optional `namespaces` field to policy entries that restricts the entry's applicability to things within matching namespaces. This enables namespace-scoped authorization without changes to the search index structure.

## Motivation

Policy entries currently apply universally regardless of the thing's namespace. This limitation affects:

1. **Multi-Tenant Scenarios**: Organizations need tenant isolation without maintaining separate policies per tenant
2. **Namespace-Based Organization**: Different user groups need different permissions per namespace (e.g., vehicle managers vs. facility managers)

## Proposed Solution

Add an optional `namespaces` array to policy entries:

```json
{
  "policyId": "com.acme:shared-policy",
  "entries": {
    "vehicle-admins": {
      "subjects": { "nginx:fleet-manager": { "type": "generated" } },
      "resources": { "thing:/": { "grant": ["READ", "WRITE"], "revoke": [] } },
      "namespaces": ["com.acme.vehicles", "com.acme.vehicles.*"]
    },
    "analysts": {
      "subjects": { "nginx:data-analyst": { "type": "generated" } },
      "resources": { "thing:/": { "grant": ["READ"], "revoke": [] } }
    }
  }
}
```

### Semantics

| `namespaces` field | Behavior |
|--------------------|----------|
| Absent | Entry applies to **all namespaces** (backward compatible) |
| Empty array `[]` | Same as absent |
| Array with patterns | Entry **only applies** when thing's namespace matches at least one pattern |

### Namespace Pattern Matching

- `com.acme` - exact match only
- `com.acme.*` - matches everything below `com.acme` but **not** `com.acme` itself

To match both: `["com.acme", "com.acme.*"]`

## Multi-Tenant Example

```json
{
  "policyId": "platform:multi-tenant-base",
  "entries": {
    "tenant-a-users": {
      "subjects": { "oidc:tenant-a-group": {} },
      "resources": { "thing:/": { "grant": ["READ", "WRITE"], "revoke": [] } },
      "namespaces": ["com.tenant-a", "com.tenant-a.*"]
    },
    "tenant-b-users": {
      "subjects": { "oidc:tenant-b-group": {} },
      "resources": { "thing:/": { "grant": ["READ", "WRITE"], "revoke": [] } },
      "namespaces": ["com.tenant-b", "com.tenant-b.*"]
    }
  }
}
```

Tenant A users **cannot** access Tenant B things (and vice versa).

## Implementation Notes

- **Search index unchanged**: Namespace filtering applied at index write time in `EvaluatedPolicy.of()` - entries filtered by thing's namespace before extracting subjects
- **Backward compatible**: Absent field means "applies to all namespaces"
- **Validation**: Reuse existing `RegexPatterns.NAMESPACE_PATTERN` with `*` allowed as trailing wildcard

## Related

- #2304 - Complementary feature for gateway-level namespace enforcement based on JWT claims
- #2347 - "entriesAdditions" - will be enhanced to also be able to configure additional "namespaces" as part of this issue

hu-ahmed:
Resolves #2325

Add support for namespace-scoped policy entries via a new optional `namespaces` field on `PolicyEntry`.

A scoped policy entry only applies to Things whose namespace matches at least one configured pattern.
If `namespaces` is omitted or empty, the entry remains globally applicable for backward compatibility.

Supported matching semantics:
- `com.acme` matches only the exact namespace `com.acme`
- `com.acme.*` matches nested namespaces such as `com.acme.vehicles`, but not `com.acme` itself

### DIFF
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyBuilder.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyBuilder.java
@@ -207,9 +207,9 @@ final class ImmutablePolicyBuilder implements PolicyBuilder {
         revokedPermissions.put(label, new LinkedHashMap<>());
 
         setResourcesFor(entry.getLabel(), entry.getResources());
-        namespaces.put(label, entry.getNamespaces());
+        namespaces.put(label, entry.getNamespaces().orElse(null));
         setImportableFor(label, entry.getImportableType());
-        allowedImportAdditions.put(label, entry.getAllowedImportAdditions());
+        allowedImportAdditions.put(label, entry.getAllowedImportAdditions().orElse(null));
     }
 
     private void putAllSubjects(final PolicyEntry policyEntry) {
@@ -404,13 +404,11 @@ final class ImmutablePolicyBuilder implements PolicyBuilder {
 
         final Collection<PolicyEntry> policyEntries = allLabels.stream()
                 .map(lbl -> {
-                    final List<String> entryNamespaces = getNamespaces(lbl);
                     final ImportableType importableType =
                             getImportableType(lbl).orElse(ImportableType.IMPLICIT);
-                    final Set<AllowedImportAddition> additions =
-                            getAllowedImportAdditions(lbl);
                     return PoliciesModelFactory.newPolicyEntry(lbl, getSubjectsForLabel(lbl),
-                            getResourcesForLabel(lbl), entryNamespaces, importableType, additions);
+                            getResourcesForLabel(lbl), getNamespaces(lbl).orElse(null),
+                            importableType, getAllowedImportAdditions(lbl).orElse(null));
                 })
                 .collect(Collectors.toList());
 
@@ -432,12 +430,12 @@ final class ImmutablePolicyBuilder implements PolicyBuilder {
         return Optional.ofNullable(importableTypes.get(Label.of(label)));
     }
 
-    private Set<AllowedImportAddition> getAllowedImportAdditions(final CharSequence label) {
-        return allowedImportAdditions.getOrDefault(Label.of(label), Collections.emptySet());
+    private Optional<Set<AllowedImportAddition>> getAllowedImportAdditions(final CharSequence label) {
+        return Optional.ofNullable(allowedImportAdditions.get(Label.of(label)));
     }
 
-    private List<String> getNamespaces(final CharSequence label) {
-        return namespaces.getOrDefault(Label.of(label), Collections.emptyList());
+    private Optional<List<String>> getNamespaces(final CharSequence label) {
+        return Optional.ofNullable(namespaces.get(Label.of(label)));
     }
 
     private Resources getResourcesForLabel(final CharSequence label) {
--- a/policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyEntry.java
+++ b/policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyEntry.java
@@ -25,6 +25,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 import org.eclipse.ditto.base.model.exceptions.DittoJsonException;
@@ -46,20 +47,23 @@ final class ImmutablePolicyEntry implements PolicyEntry {
     private final Label label;
     private final Subjects subjects;
     private final Resources resources;
-    private final List<String> namespaces;
+    @Nullable private final List<String> namespaces;
     private final ImportableType importableType;
-    private final Set<AllowedImportAddition> allowedImportAdditions;
+    @Nullable private final Set<AllowedImportAddition> allowedImportAdditions;
 
     private ImmutablePolicyEntry(final Label theLabel, final Subjects theSubjects, final Resources theResources,
-            final List<String> namespaces, final ImportableType importableType,
-            final Set<AllowedImportAddition> allowedImportAdditions) {
+            @Nullable final List<String> namespaces, final ImportableType importableType,
+            @Nullable final Set<AllowedImportAddition> allowedImportAdditions) {
         label = checkNotNull(theLabel, "label");
         subjects = theSubjects;
         resources = theResources;
-        this.namespaces = Collections.unmodifiableList(new ArrayList<>(new LinkedHashSet<>(namespaces)));
+        this.namespaces = namespaces != null
+                ? Collections.unmodifiableList(new ArrayList<>(new LinkedHashSet<>(namespaces)))
+                : null;
         this.importableType = importableType;
-        this.allowedImportAdditions = Collections.unmodifiableSet(
-                new LinkedHashSet<>(allowedImportAdditions));
+        this.allowedImportAdditions = allowedImportAdditions != null
+                ? Collections.unmodifiableSet(new LinkedHashSet<>(allowedImportAdditions))
+                : null;
     }
 
     /**
@@ -74,8 +78,8 @@ final class ImmutablePolicyEntry implements PolicyEntry {
     public static PolicyEntry of(final Label label, final Subjects subjects, final Resources resources) {
         checkNotNull(subjects, "subjects");
         checkNotNull(resources, "resources");
-        return new ImmutablePolicyEntry(label, subjects, resources, Collections.emptyList(),
-                ImportableType.IMPLICIT, Collections.emptySet());
+        return new ImmutablePolicyEntry(label, subjects, resources, null,
+                ImportableType.IMPLICIT, null);
     }
 
     /**
@@ -94,8 +98,8 @@ final class ImmutablePolicyEntry implements PolicyEntry {
         checkNotNull(subjects, "subjects");
         checkNotNull(resources, "resources");
         checkNotNull(importableType, "importableType");
-        return new ImmutablePolicyEntry(label, subjects, resources, Collections.emptyList(),
-                importableType, Collections.emptySet());
+        return new ImmutablePolicyEntry(label, subjects, resources, null,
+                importableType, null);
     }
 
     /**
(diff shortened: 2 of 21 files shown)


policies/enforcement/src/main/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcer.java
```diff
@@ -110,4 +114,28 @@ public Enforcer getEnforcer() {
...
+    public PolicyEnforcer forNamespace(final String thingNamespace) {
+        if (policy == null) {
+            return this;
+        }
+        final boolean anyEntryHasNamespaces = StreamSupport.stream(policy.spliterator(), false)
+                .anyMatch(entry -> !entry.getNamespaces().isEmpty());
+        if (!anyEntryHasNamespaces) {
+            return this;
+        }
+        final List<PolicyEntry> filteredEntries = StreamSupport.stream(policy.spliterator(), false)
```
thjaeckle: ### MEDIUM: Double iteration over policy entries

The code iterates through all policy entries twice:
1. Lines 129-130: `anyMatch(entry -> !entry.getNamespaces().isEmpty())` — full scan to check if any entry has namespaces
2. Lines 134-136: `filter(entry -> entry.appliesToNamespace(...))` — full scan again to filter

This can be merged into a single pass:
```java
final List<PolicyEntry> filteredEntries = StreamSupport.stream(policy.spliterator(), false)
        .filter(entry -> entry.appliesToNamespace(thingNamespace))
        .collect(Collectors.toList());
if (filteredEntries.size() == (int) StreamSupport.stream(policy.spliterator(), false).count()) {
    return this; // no entries were filtered out
}
```

Or more efficiently, track whether any entry was actually filtered:
```java
boolean anyFiltered = false;
final List<PolicyEntry> filtered = new ArrayList<>();
for (final PolicyEntry e : policy) {
    if (e.appliesToNamespace(thingNamespace)) {
        filtered.add(e);
    } else {
        anyFiltered = true;
    }
}
if (!anyFiltered) return this;
```

Note: `forNamespace()` is called on every enforcement via `ThingEnforcerActor.providePolicyEnforcer()`, so this runs on every thing command.

policies/model/src/main/java/org/eclipse/ditto/policies/model/PolicyEntry.java
```diff
@@ -111,6 +112,48 @@ default JsonSchemaVersion[] getSupportedSchemaVersions() {
...
+     * @param namespace the thing's namespace to check against.
+     * @return {@code true} if this entry applies to the given namespace.
+     * @since 3.9.0
+     */
+    default boolean appliesToNamespace(final String namespace) {
+        final List<String> patterns = getNamespaces();
+        if (patterns.isEmpty()) {
+            return true;
+        }
+        for (final String pattern : patterns) {
```
thjaeckle: ### MEDIUM: Inconsistency with `NamespacePoliciesConfig.namespaceMatchesPattern()` from PR #2367

This method supports two pattern types: exact match and `prefix.*` wildcard.
PR #2367's `NamespacePoliciesConfig.namespaceMatchesPattern()` supports the same two PLUS `*` (catch-all).

While the validator in `ImmutablePolicyEntry.isValidNamespacePattern()` correctly rejects `*` for policy entries, the two matching methods have divergent implementations:
- Here: `namespace.startsWith(prefix + ".")` (concatenates `"."` each call)
- PR #2367: `namespace.startsWith(pattern.substring(0, pattern.length() - 1))` (the substring already includes the dot)

Consider extracting a shared static utility method to avoid subtle divergence as the codebase evolves. Both PRs introduce namespace pattern matching independently — having two slightly different implementations is a maintenance risk.

policies/enforcement/src/test/java/org/eclipse/ditto/policies/enforcement/PolicyEnforcerTest.java
```diff
@@ -0,0 +1,94 @@
...
+                Collections.singletonList(Subject.newInstance(SubjectId.newInstance(subjectId))),
+                Collections.singletonList(PoliciesModelFactory.newResource("thing", "/",
+                        EffectedPermissions.newInstance(
+                                PoliciesModelFactory.newPermissions(Permission.READ),
+                                PoliciesModelFactory.noPermissions()))),
+                ImportableType.IMPLICIT,
+                Collections.emptySet(),
+                namespaces);
+    }
+}
```
thjaeckle, line 105: ### MEDIUM: Missing test for `forNamespace()` edge case

The production code handles `policy == null` at `PolicyEnforcer.java:126`:
```java
if (policy == null) {
    return this;
}
```

But no test covers this path. Consider adding:
```java
@Test
public void forNamespaceWithNullPolicyReturnsSameInstance() {
    final PolicyEnforcer enforcer = new PolicyEnforcer(null,
            PolicyEnforcers.defaultEvaluator(Collections.emptyList()));
    assertThat(enforcer.forNamespace("any.ns")).isSameAs(enforcer);
}
```

Also: `forNamespace()` doesn't guard against `thingNamespace == null`. If a thing has no namespace set (shouldn't happen but defensively), `startsWith(prefix + ".")` would NPE. Consider adding a null check or documenting the precondition.

policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyEntry.java
```diff
@@ -176,6 +213,44 @@ private static Set<AllowedImportAddition> readAllowedImportAdditions(final JsonO
...
+            throw PolicyEntryInvalidException.newBuilder()
+                    .description("The value '" + value + "' in field '" +
+                            JsonFields.NAMESPACES.getPointer() +
+                            "' is not a valid namespace pattern. Valid patterns are namespaces" +
+                            " like 'com.acme' or wildcard patterns like 'com.acme.*'.")
+                    .build();
+        }
+    }
+
+    private static boolean isValidNamespacePattern(final String value) {
```
thjaeckle: ### MEDIUM: No duplicate validation for namespace patterns

The list `["com.acme", "com.acme"]` passes validation since each element is individually valid. While not harmful at runtime (the duplicate just causes a redundant match check), it could indicate a user misconfiguration.

Consider either:
- Deduplicating silently (e.g., `new ArrayList<>(new LinkedHashSet<>(namespaces))`)
- Or rejecting duplicates with a descriptive error

Also: the list preserves insertion order, which is good for deterministic behavior. But consider whether `Set<String>` would better represent the semantics (unordered, no duplicates). The current `List<String>` choice is fine for JSON round-tripping though.

documentation/src/main/resources/pages/ditto/basic-policy.md
```diff
@@ -340,7 +354,8 @@ Example of a policy specifying different types of `importable` entries and allow
       "subjects": { ... },
       "resources": { ... },
       "importable": "implicit",
-      "allowedImportAdditions": [ "subjects" ]
+      "allowedImportAdditions": [ "subjects" ],
+      "namespaces": [ "com.acme", "com.acme.*" ]
```
thjaeckle: I would move that to below `subjects` and `resources` - does not really matter from JSON perspective - but is more logically to have those 3 "together".

documentation/src/main/resources/openapi/sources/schemas/policies/policyEntry.yml
```diff
@@ -19,6 +19,8 @@ properties:
     $ref: 'importable.yml'
   allowedImportAdditions:
     $ref: 'allowedImportAdditions.yml'
+  namespaces:
```
thjaeckle: I would move that up to below `resources` definition

documentation/src/main/resources/pages/ditto/basic-policy.md
thjaeckle: This needs to be enhanced with `"namespaces"` as well.

policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyBuilder.java
```diff
@@ -56,6 +58,7 @@ private ImmutablePolicyBuilder() {
         revokedPermissions = new LinkedHashMap<>();
         importableTypes = new LinkedHashMap<>();
         allowedImportAdditions = new LinkedHashMap<>();
+        namespaces = new LinkedHashMap<>();
```
thjaeckle: I would also move this declaration up - below `revokedPermissions`

policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyBuilder.java
```diff
@@ -42,6 +43,7 @@ final class ImmutablePolicyBuilder implements PolicyBuilder {
     private final Map<Label, Map<ResourceKey, Permissions>> revokedPermissions;
     private final Map<Label, ImportableType> importableTypes;
     private final Map<Label, Set<AllowedImportAddition>> allowedImportAdditions;
+    private final Map<Label, List<String>> namespaces;
```
thjaeckle: please move this declaration up - above `importableTypes`

policies/model/src/main/java/org/eclipse/ditto/policies/model/ImmutablePolicyEntry.java
```diff
@@ -111,7 +118,35 @@ public static PolicyEntry of(final Label label, final Subjects subjects, final R
...
+     * @param importableType specifies whether and how this entry is allowed to be imported by others.
+     * @param allowedImportAdditions which types of additions are allowed when importing this entry.
+     * @param namespaces the namespace patterns restricting which thing namespaces this entry applies to.
+     * @return a new {@code PolicyEntry} object.
+     * @throws NullPointerException if any argument is {@code null}.
+     * @since 3.9.0
+     */
+    public static PolicyEntry of(final Label label, final Subjects subjects, final Resources resources,
+            final ImportableType importableType, final Set<AllowedImportAddition> allowedImportAdditions,
+            final List<String> namespaces) {
```
thjaeckle: I would not create the `namespaces` parameter as last one - but move behind `resources`

thjaeckle (approved):
@hu-ahmed PR is in a great shape, thanks a lot 👍 

Did a final review - and added as part of that to not by default fallback to `"namespaces": []` if none were explicitly present/persisted (e.g. when querying the new `GET /.../namespaces` this was returned for policies without any namespaces defined).

PR #2375, opened 2026-03-22
Add horizontally-resizable pane functionality

### LINKED ISSUE #2374: UI: Make the split panels resizable horizontally and vertically
Currently, tables are resizable vertically via a resize handle (which is easy to miss), but this only applies to the whole table and only vertically.

However, in certain scenarios, it would be useful to have the option to resize the different panes horizontally. For example, when there are multiple fields for the things table, it would be good to be able to expand the search table itself (and therefore make the details/management pane smaller)

There should be proper drag-to-resize dividers between the main panes in each tab, both horizontally (between the list column and the detail/editor column) and vertically (between sections like the things table, attributes and features in the Things tab).

NunoGFernandes:
Make Explorer UI panes resizable via draggable splitters (Split.js) and persist sizes

Adds a split-pane system to the Explorer UI so list/detail panels can be horizontally resized by drag handles.

Closes #2374

### DIFF
--- a/ui/modules/utils/splitPanes.ts
+++ b/ui/modules/utils/splitPanes.ts
@@ -18,11 +18,16 @@ import Split from "split.js";
  * containers. Their direct children with class "split-panel"
  * are the resizable panels.
  */
-document.addEventListener("mouseup", () => {
-  document.body.classList.remove("split-dragging-h", "split-dragging-v");
-});
+let splitPanesInitialized = false;
 
 export function initSplitPanes(): void {
+  if (!splitPanesInitialized) {
+    document.addEventListener("mouseup", () => {
+      document.body.classList.remove("split-dragging-h", "split-dragging-v");
+    });
+    splitPanesInitialized = true;
+  }
+
   document.querySelectorAll<HTMLElement>("[data-split]").forEach((container) => {
     const direction = container.dataset.split as "horizontal" | "vertical";
     const sizesAttr = container.dataset.splitSizes;
@@ -57,9 +62,18 @@ export function initSplitPanes(): void {
       sizes = panels.map(() => 100 / panels.length);
     }
 
+    let minSize = 50;
+    if (minAttr) {
+      try {
+        minSize = JSON.parse(minAttr);
+      } catch {
+        /* ignore */
+      }
+    }
+
     Split(panels, {
       sizes,
-      minSize: minAttr ? JSON.parse(minAttr) : 50,
+      minSize,
       gutterSize: 6,
       direction,
       snapOffset: 0,
@@ -78,9 +92,18 @@ export function initSplitPanes(): void {
           localStorage.setItem(`split-sizes-${storageKey}`, JSON.stringify(newSizes));
         }
       },
-      onDrag: () => {
-        window.dispatchEvent(new Event("resize"));
-      },
+      onDrag: (() => {
+        let rafPending = false;
+        return () => {
+          if (!rafPending) {
+            rafPending = true;
+            requestAnimationFrame(() => {
+              window.dispatchEvent(new Event("resize"));
+              rafPending = false;
+            });
+          }
+        };
+      })(),
     });
   });
 }
--- a/ui/modules/environments/environments.html
+++ b/ui/modules/environments/environments.html
@@ -10,8 +10,8 @@
   ~
   ~ SPDX-License-Identifier: EPL-2.0
   -->
-<div class="row resizable_pane" style="height: 80vh">
-    <div class="col-md-4 resizable_flex_column">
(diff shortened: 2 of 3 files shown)


ui/package-lock.json
```diff
@@ -5375,6 +5376,12 @@
         "source-map": "^0.6.0"
       }
     },
+    "node_modules/split.js": {
+      "version": "1.6.5",
+      "resolved": "https://nexus.cbsdev.net/repository/npm-all/split.js/-/split.js-1.6.5.tgz",
```
thjaeckle: This seems to cause the build to fail as it cannot download from `https://nexus.cbsdev.net`.

Why would that not be `https://registry.npmjs.org` as for other dependencies?

ui/modules/utils/splitPanes.ts
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
thjaeckle: Please use the current year for newly created files.

ui/package-lock.json
NunoGFernandes: Ups! My bad. I forgot to change the default registry from my custom one

ui/modules/utils/splitPanes.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Initializes all split panes in the document.
+ * Elements with data-split="horizontal" become resizable split
+ * containers. Their direct children with class "split-panel"
+ * are the resizable panels.
+ */
+document.addEventListener("mouseup", () => {
+  document.body.classList.remove("split-dragging-h", "split-dragging-v");
+});
```
thjaeckle: shouldn't this be moved to `initSplitPanes()`, also avoiding double-registration?

ui/package-lock.json
```diff
@@ -5375,6 +5376,11 @@
         "source-map": "^0.6.0"
       }
     },
+    "node_modules/split.js": {
+      "version": "1.6.5",
+      "resolved": "https://registry.npmjs.org/split.js/-/split.js-1.6.5.tgz",
+      "integrity": "sha512-mPTnGCiS/RiuTNsVhCm9De9cCAUsrNFFviRbADdKiiV+Kk8HKp/0fWu7Kr8pi3/yBmsqLFHuXGT9UUZ+CNLwFw=="
```
thjaeckle, line 5382: The integrity hash did not change after changing the registry to npmjs.

Could you run `npm install` to make sure the integrity hash is correct?

ui/modules/utils/splitPanes.ts
```diff
@@ -0,0 +1,86 @@
...
+        /* ignore */
+      }
+    }
+    if (!sizes) {
+      sizes = panels.map(() => 100 / panels.length);
+    }
+
+    Split(panels, {
+      sizes,
+      minSize: minAttr ? JSON.parse(minAttr) : 50,
```
thjaeckle: For the other `JSON.parse` calls you catched errors and ignored them. Why not here for `minAttr`?

ui/modules/utils/splitPanes.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+      onDragEnd: (newSizes) => {
+        document.body.classList.remove("split-dragging-h", "split-dragging-v");
+        if (storageKey) {
+          localStorage.setItem(`split-sizes-${storageKey}`, JSON.stringify(newSizes));
+        }
+      },
+      onDrag: () => {
+        window.dispatchEvent(new Event("resize"));
+      },
```
thjaeckle: This fires potentially hundreds of resize events per second during a drag. Ace editors and other listeners will re-layout on each event. This should be debounced (e.g., `requestAnimationFrame` or a 50ms throttle).

ui/modules/utils/splitPanes.ts
NunoGFernandes: The module-level registration was indeed an oversight. Moved the mouseup listener to initSplitPanes() with an initialization guard flag so that it is registered only once

ui/package-lock.json
NunoGFernandes: I did, and the hash hasn't changed

ui/modules/utils/splitPanes.ts
NunoGFernandes: Made the change for consistency

ui/modules/utils/splitPanes.ts
NunoGFernandes: Replaced this with a requestAnimationFrame based throttle, making the event fire at most once per anumation frame

thjaeckle (approved):
Changes LGTM - thanks a lot for your first contribution to Ditto @NunoGFernandes 

Hoping for even more :)

PR #2384, opened 2026-03-31
use string subject IDs in PRE_DEFINED_EXTRA_FIELDS_READ_GRANT_OBJECT …

hu-ahmed:
## Summary                                                                                                                                                                              
                                                            
  - Fix: `ThingEventEnricher` now produces `PRE_DEFINED_EXTRA_FIELDS_READ_GRANT_OBJECT`                                                                                                   
    header with string subject IDs instead of integer indices, matching the format                                                                                                        
    expected by `DittoCachingSignalEnrichmentFacade`                                                                                                                                      
  - Add `ReadGrant.toJson()` to serialize grants as `{"/path": ["subject1", "subject2"]}`                                                                                                 
  - Remove dead code: `IndexedReadGrant` and `ReadGrantIndexer` (no remaining usages)                                                                                                     
  - Add regression tests verifying string format in both `ReadGrantCollectorTest`                                                                                                         
    and `ThingEventEnricherTest`                                                                                                                                                          
                                                                                                                                                                                          
 ## Root cause                                                                                                                                                                           
                                                            
  `ReadGrantIndexer.index()` converted `ReadGrant` to an indexed format where
  subject IDs are replaced by integer indices to reduce header size:                                                                                                                      
                                                                    
  ```json                                                                                                                                                                                 
  {"/attributes": [0]}                                      
  ```                 
     
  Where `0` maps to a subject ID string. But
  `DittoCachingSignalEnrichmentFacade.filterAskedForFieldSelectorToGrantedFields()`                                                                                                       
  compares the array values directly against the requesting connection's authorization
  subject ID strings. `JsonValue(integer)` never equals `JsonValue(string)`,                                                                                                              
  so the filter returned an empty field selector, causing all pre-defined extra fields                                                                                                    
  to be silently dropped from enriched events.                                                                                                                                            
                                                                                                                                                                                          
  ## Impact                                                                                                                                                                               
                                                                                                                                                                                          
  All connections relying on pre-defined extra fields enrichment (e.g.
  `extraFields=attributes`) received events with empty `extra`, causing                                                                                                                   
  downstream consumers to fail when attempting to read expected fields.


things/service/src/test/java/org/eclipse/ditto/things/service/utils/ReadGrantCollectorTest.java
```diff
@@ -223,6 +228,61 @@ private static Policy createPolicyWithNestedPartialAccess() {
...
+    /**
+     * Regression test: verifies that {@link ReadGrant#toJson()} produces string subject IDs
+     * (not integer indices), which is required by
+     * {@code DittoCachingSignalEnrichmentFacade.filterAskedForFieldSelectorToGrantedFields()}.
+     * <p>
+     * The facade matches authorization subject IDs directly against the array values in the
+     * {@code PRE_DEFINED_EXTRA_FIELDS_READ_GRANT_OBJECT} header. If the values were integer
+     * indices instead of strings, the match would always fail and extra fields would be silently
+     * dropped — causing the "Could not resolve attributes from extra" error seen in production
+     * after the Ditto 3.9.0-M1 deployment.
```
thjaeckle: This comment should be removed - irrelevant.

Issue #2180, opened 2025-06-02
Add Helm-Chart compatibility with any ingress controller


mladBlum:
**Current Situation:**

The ingress resources in the Helm chart contain numerous configurations (annotations) specific to the Kubernetes ingress-nginx controller ([ingress-nginx GitHub repository](https://github.com/kubernetes/ingress-nginx)) and are not universally compatible with other ingress controllers.
For individuals using a different ingress controller, the current options are:

1. Disable all ingress and create them manually
2. Override all the defaults in values.yaml

Both methods require considerable effort,  initially and also before each update of Ditto. I think the aim should be that someone who uses the helm chart should have as little effort as possible with installation and updates.

**Proposal:**

I believe we can achieve a big improvement if we manage to remove the ingress-nginx specific annotations to create a "neutral" ingress configuration. This ingress will align with typical ingress configurations in other charts, allowing operators to add their specific settings for whatever they want to achieve.

While I was validating this approach, I discovered that this should be quiety easy. Ditto-nginx already serves as entry point to ditto and the specific annotations in the current ingress resources are already present in the ditto-nginx config.

See ditto-nginx:

- etc/nginx/nginx.conf
- etc/nginx/nginx-cors.conf

Current ingress annotations:

- nginx.ingress.kubernetes.io/server-snippet
- nginx.ingress.kubernetes.io/configuration-snippet

If everything aligns well, the necessary changes might be minimal and simplifies the current ingress resources.
I welcome feedback if I've missed any points. I know, of course, that many things are not as easy as they seem at first.

Btw. I think there is something odd with the current setup anyway. 
The Ingress ressources apply nginx configurations that are already applied by ditto-nginx. So there are two almost identical configurations that are applied twice. 

```mermaid
flowchart LR
subgraph chart [Helm-Chart]
ingress_controller --> ditto-nginx

headers_chart((nginx.config from ingress)) --> ingress_controller
headers_ditto((nginx.config included in ditto-nginx)) --> ditto-nginx

ditto-nginx --> ui(ui)
ditto-nginx --> swagger(swagger)
ditto-nginx --> any(...)
ingress_controller --> gateway
end

client --> ingress_controller(Ingress Controller)
```

PR #2386, opened 2026-04-01
remove ingress-nginx controller

### LINKED ISSUE #2056: Simplify helm chart by resolving ingress controller as an external dependency
Currently we have the nginx-ingress controller which does its job well, but there is some overhead introduced there.
They idea of having all its necessary resources as templates maybe wasn't optimal.

I would like to suggest that we move on to the official nginx-ingress [charts](https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx) as a dependency.
That way we could only rework the values.yaml to coform to the requirements of that helm chart and remove manual resources we are currently supporting :)

kalinkostashki:
- resolves #2056
- ingress-nginx is deprecated as of 31.03.2026
- controller should be something the user of the helm chart defines
- removed all ingress-nginx specific annotations
- removed nginx-ingress-auth.yaml and nginx-ingress.yaml -> those should now be defined externally
- added a blog post in order to explain the approach and reasons behind the decision + tested each scenario in the blogpost to make sure it works

### DIFF
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -63,9 +63,9 @@ jobs:
         # which a folder exists at
         # https://github.com/yannh/kubernetes-json-schema/
         k8s:
-          - v1.28.13
-          - v1.29.8
-          - v1.30.4
+          - v1.33.7
+          - v1.34.3
+          - v1.35.1
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


mladBlum:
Looks good for me. As I see you solved also the hazle with ditto-ui access when removing the nginx stuff! That i was stuck on. I am glad that this is finally done. :+1: 
We can close #2180

kalinkostashki:
> Looks good, I suppose 😄
> 
> @kalinkostashki that means we would already introduce Helm chart version 4.0.0 together with Ditto release 3.9.0?
> 
> Might lead to confusion - we should document that somewhere that now Helm chart and Ditto appVersion are no longer aligned. It was only a matter of time until that happens, so it's ok ..
> 
> Could you resolve the remaining merge conflicts?

done  :)

PR #2390, opened 2026-04-01
Fix spurious "Unknown message" warnings in ConnectionPersistenceActor

thjaeckle:
During connection recovery, ConnectionPersistenceActor logged WARN-level "Unknown message: ReceiveTimeout" and "Unknown message: PA_RECOVERED" messages. These were harmless but added noise to logs, especially during rolling deployments when many connections recover simultaneously.

Root causes:
- ReceiveTimeout: The 2s receive timeout set in ConnectionSupervisorActor .preStart() could get stashed during startup and later forwarded to the persistence actor via the supervisor's catch-all matchAny handler, since activeBehaviour had no explicit ReceiveTimeout handler.
- PA_RECOVERED: The NoOpEnforcerActor's blanket matchAny echoed all messages back to the sender. When the supervisor forwarded PA_RECOVERED to the enforcer with sender=PA, the NoOpEnforcerActor bounced it right back to the persistence actor.

Fixes:
- ConnectionSupervisorActor: Handle ReceiveTimeout in activeBehaviour
- NoOpEnforcerActorPropsFactory: Consume PA_RECOVERED control signal instead of echoing it back to the persistence actor
- ConnectionPersistenceActor: Defense-in-depth handlers for both messages at DEBUG level in matchAnyAfterInitialization

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/enforcement/NoOpEnforcerActorPropsFactory.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/enforcement/NoOpEnforcerActorPropsFactory.java
@@ -16,6 +16,8 @@ import org.eclipse.ditto.connectivity.model.ConnectionId;
 
 import com.typesafe.config.Config;
 
+import org.eclipse.ditto.policies.enforcement.AbstractEnforcerActor;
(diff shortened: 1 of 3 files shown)


hu-ahmed:
LGTM!

PR #2397, opened 2026-04-07
introduced empty rql filter

### LINKED ISSUE #2377: Support empty() RQL function to filter for absent or empty fields
## Problem

Users with array-valued attributes (e.g. `attributes/tags` containing `["tag1", "tag2"]`) can currently search for things containing a specific value:

```
eq(attributes/tags,"some-array-value")
```

They can also check for field absence using:

```
not(exists(attributes/tags))
```

However, there is **no way to find things where a field exists but is empty** — e.g. an empty array `[]`, an empty object `{}`, or an empty string `""`.

A common real-world use case is: *"find all things that either have no tags at all, or have an empty tags array"*. This currently requires two conditions, one of which (`empty array check`) is not expressible in RQL.

## Proposed solution: `empty()` RQL function

Introduce a new RQL filter function `empty(<field>)` that returns `true` when a field is **absent or void of content**, regardless of its JSON type.

### Syntax

```
empty(attributes/tags)
```

Negation via `not()`:

```
not(empty(attributes/tags))
```

### Semantics

`empty(field)` matches when:
- The field does not exist
- The field is `null`
- The field is an empty array `[]`
- The field is an empty object `{}`
- The field is an empty string `""`

`not(empty(field))` matches when:
- The field exists AND has meaningful content (non-empty array, non-empty object, non-empty string, number, boolean)

### Examples

| RQL                                   | Meaning                                           |
|---------------------------------------|---------------------------------------------------|
| `empty(attributes/tags)`              | Tags absent, null, or empty                       |
| `not(empty(attributes/tags))`         | Tags exist and contain at least one value          |
| `or(empty(attributes/tags),eq(attributes/tags,"default"))` | Tags empty or contain "default" |

## Implementation: MongoDB mapping

A key benefit of this approach is that **the type of the field does not need to be known upfront**. MongoDB's operators are type-tolerant in `$or` branches — each branch simply doesn't match when the type doesn't apply, without errors.

### `empty(field)` maps to:

```json
{
  "$or": [
    { "t.attributes.tags": { "$exists": false } },
    { "t.attributes.tags": null },
    { "t.attributes.tags": { "$size": 0 } },
    { "t.attributes.tags": { "$eq": {} } },
    { "t.attributes.tags": { "$eq": "" } }
  ]
}
```

### `not(empty(field))` maps to:

```json
{
  "$and": [
    { "t.attributes.tags": { "$exists": true } },
    { "t.attributes.tags": { "$ne": null } },
    { "t.attributes.tags": { "$not": { "$size": 0 } } },
    { "t.attributes.tags": { "$ne": {} } },
    { "t.attributes.tags": { "$ne": "" } }
  ]
}
```

### Type-agnostic behavior

Each `$or` branch targets a specific type and is silently skipped for non-matching types:

| Field value | `$exists: false` | `null` | `$size: 0` | `$eq: {}` | `$eq: ""` |
|-------------|-------------------|--------|------------|-----------|-----------|
| missing     | ✅                |        |            |           |           |
| `null`      |                   | ✅     |            |           |           |
| `[]`        |                   |        | ✅         |           |           |
| `{}`        |                   |        |            | ✅        |           |
| `""`        |                   |        |            |           | ✅        |
| `[1,2]`     |                   |        |            |           |           |
| `{"a":1}`   |                   |        |            |           |           |
| `"hello"`   |                   |        |            |           |           |

All conditions are **index-friendly** — no `$expr` or other slow operators are needed.

## Affected components

- **RQL parser** (`rql/parser`): Add `empty` as a new single-logical operator (similar to `exists`)
- **RQL model** (`rql/model`): Add AST node type for `empty`
- **RQL query** (`rql/query`): Add criteria and visitor support
- **Things search persistence** (`thingsearch/service`): Add BSON translation in the predicate/criteria visitors

hu-ahmed:
Resolves: #2377  
## Summary                                                                                                                                                                 
                                                                                                                                                                           
  Introduces a new RQL filter function `empty(<field>)` that returns true when a field is absent                                                                             
  or void of content (null, empty array, empty object, empty string), regardless of its JSON type.
                                                                                                                                                                             
  This enables use cases like finding all things where e.g. `attributes/tags` either doesn't exist                                                                           
  or is an empty array — which was previously not expressible in a single RQL filter.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/Thing3ValuePredicateVisitor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/persistence/Thing3ValuePredicateVisitor.java
@@ -29,6 +29,7 @@ import org.eclipse.ditto.rql.query.expression.ExistsFieldExpression;
 import org.eclipse.ditto.rql.query.expression.FieldExpression;
 import org.eclipse.ditto.rql.query.expression.FilterFieldExpression;
 import org.eclipse.ditto.rql.query.expression.visitors.FieldExpressionVisitor;
+import org.eclipse.ditto.rql.query.things.EmptyThingPredicateVisitor;
 import org.eclipse.ditto.rql.query.things.ExistsThingPredicateVisitor;
 import org.eclipse.ditto.rql.query.things.FilterThingPredicateVisitor;
 import org.eclipse.ditto.rql.query.things.ThingPredicatePredicateVisitor;
@@ -101,6 +102,13 @@ final class Thing3ValuePredicateVisitor implements CriteriaVisitor<Function<Thin
         return thing -> isUnknownField(fieldExpression)
                 ? Trilean.UNKNOWN
                 : Trilean.lift(ExistsThingPredicateVisitor.apply(fieldExpression, placeholderResolvers).test(thing));
+    }
+
+    @Override
+    public Function<Thing, Trilean> visitEmpty(final ExistsFieldExpression fieldExpression) {
+        return thing -> isUnknownField(fieldExpression)
+                ? Trilean.UNKNOWN
+                : Trilean.lift(EmptyThingPredicateVisitor.apply(fieldExpression, placeholderResolvers).test(thing));
     }
 
     @Override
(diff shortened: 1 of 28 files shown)


thjaeckle:
System tests run revealed some errors:
https://github.com/eclipse-ditto/ditto/actions/runs/24084958670

thjaeckle:
System tests run:
https://github.com/eclipse-ditto/ditto/actions/runs/24084958670

Just contains 1 error which is known to sometimes fail (unrelated to the changes) ✅

thjaeckle (approved):
LGTM - thanks a lot for the added feature @hu-ahmed 🎉

PR #2401, opened 2026-04-08
fix Kafka consumer crash loop on messages with blank header values

thjaeckle:
Blank tracing header values (e.g. empty correlation-id) caused IllegalArgumentException in Tag.of() during span creation, which was outside the try-catch in KafkaMessageTransformer. With QoS 1, the uncommitted offset led to re-consuming the same poison message indefinitely.

Three defensive fixes:
- SpanTagging: skip blank values like null in correlationId/connectionId/entityId
- KafkaMessageTransformer: move span creation inside try-catch block
- Consumer streams: filter null results from transform() to prevent NPE


hu-ahmed:
LGTM!

PR #2403, opened 2026-04-09
Add importsAliases to policies for transparent subject fan-out across imported entries

### LINKED ISSUE #2402: Add importsAliases concept to policies for transparent subject fan-out across imported entries
## Summary

Add a new `importsAliases` concept to the Policy model that allows mapping a single label to one or more `entriesAdditions` targets across policy imports. Subject operations on the alias label transparently fan out to all targets, enabling backward-compatible API migration when template policies split a single entry into multiple namespace-scoped entries.

The concept is named `importsAliases` (rather than `subjectAliases`) to reflect its tight coupling with policy imports and to allow future extension beyond subject-only operations (e.g., resources, namespaces).

## Motivation

With the introduction of `entriesAdditions` in policy imports (Ditto 3.9.0), template-based policies enable reuse of resource definitions. However, a migration gap exists:

Consider a building management system where thousands of policies grant an `"operator"` entry with permissions across multiple feature domains (HVAC control, lighting, access control). Each policy duplicates the same resource definitions:

```json
{
  "policyId": "buildingmgmt:floor-3-wing-a",
  "imports": {
    "buildingmgmt:admin-access": {}
  },
  "entries": {
    "operator": {
      "subjects": {},
      "resources": {
        "thing:/features/hvac-control": { "grant": ["READ"], "revoke": [] },
        "thing:/features/lighting": { "grant": ["READ"], "revoke": [] },
        "thing:/features/access-control": { "grant": ["READ"], "revoke": [] },
        "message:/features/hvac-control/inbox": { "grant": ["WRITE"], "revoke": [] },
        "message:/features/lighting/inbox": { "grant": ["WRITE"], "revoke": [] },
        "message:/features/access-control/inbox": { "grant": ["WRITE"], "revoke": [] }
      },
      "importable": "implicit"
    }
  }
}
```

An external team manages operator subjects via:
```
PUT /api/2/policies/buildingmgmt:floor-3-wing-a/entries/operator/subjects
```

Moving to a template-based approach with `entriesAdditions` eliminates the resource duplication. However, for better namespace scoping, the template may need to **split** the single `"operator"` entry into multiple entries — e.g., `"operator-hvac"` (namespace: `buildingmgmt.hvac`) and `"operator-lighting"` (namespace: `buildingmgmt.lighting`).

This creates a problem: the external team's API call `PUT /entries/operator/subjects` no longer works because there is no local `"operator"` entry — the subjects now live in `entriesAdditions` spread across multiple imported entries.

## Proposed Solution: `importsAliases`

A new top-level field on the Policy, closely associated with `imports`:

```json
{
  "policyId": "buildingmgmt:floor-3-wing-a",
  "imports": {
    "buildingmgmt:operator-template": {
      "entries": ["operator-hvac", "operator-lighting"],
      "entriesAdditions": {
        "operator-hvac": {
          "subjects": { "oauth2:jane@building.com": { "type": "operator" } }
        },
        "operator-lighting": {
          "subjects": { "oauth2:jane@building.com": { "type": "operator" } }
        }
      }
    }
  },
  "importsAliases": {
    "operator": {
      "targets": [
        { "import": "buildingmgmt:operator-template", "entry": "operator-hvac" },
        { "import": "buildingmgmt:operator-template", "entry": "operator-lighting" }
      ]
    }
  },
  "entries": {}
}
```

### Key behaviors

- **Subject operations** on `/entries/operator/subjects` transparently fan out to all `entriesAdditions` targets referenced by the alias
- **Label exclusivity**: A label cannot exist as both an `importsAlias` and a local `entry` simultaneously (enforced with 409 Conflict)
- **Subject-only** (currently): Aliases only handle subject operations. Resource/namespace operations on an alias label return "entry not found"
- **Read operations** return the `entriesAdditions` subjects from the first target (all targets are kept in sync by fan-out writes)
- **Direct management API**: New CRUD endpoints under `/importsAliases/{label}` for managing alias definitions
- **Target validation**: Alias targets must reference existing imports; empty targets are rejected
- **Size validation**: Fan-out size is validated after expansion to prevent policy size limit bypass
- **Duplicate deduplication**: Duplicate targets are automatically deduplicated

### New API endpoints

```
GET/PUT/DELETE /api/2/policies/{id}/importsAliases
GET/PUT/DELETE /api/2/policies/{id}/importsAliases/{label}
DELETE         /api/2/policies/{id}/imports              (also new)
```

### Edge cases

- Deleting an import referenced by an alias is rejected (must remove alias first)
- Deleting all imports is rejected if any alias exists with targets
- Direct modification of `entriesAdditions` that are alias targets is allowed (next alias write re-syncs all targets)
- Targets may span multiple imports

## Acceptance Criteria

- [x] New `importsAliases` field in the Policy JSON model
- [x] Label conflict validation between entries and aliases
- [x] Transparent fan-out for `ModifySubjects`, `ModifySubject`, `DeleteSubject` through alias
- [x] Transparent read resolution for `RetrieveSubjects`, `RetrieveSubject` through alias
- [x] CRUD API for alias definitions (`/importsAliases`)
- [x] DELETE API for bulk imports (`/imports`)
- [x] Dedicated event types for alias CRUD and subject fan-out operations
- [x] Import deletion protection when referenced by alias
- [x] Target validation at alias creation time
- [x] Size validation after fan-out
- [x] Duplicate target deduplication
- [x] Empty targets rejection
- [x] Ditto Protocol adapter support (WebSocket/connectivity)
- [x] OpenAPI specification
- [x] JSON schema
- [x] Documentation
- [x] Unit tests

thjaeckle:
## Summary

Closes #2402

Adds `importsAliases` — a new top-level Policy concept that maps a label to one or more `entriesAdditions` targets across policy imports. Subject operations on the alias label transparently fan out to all targets via the existing `/entries/{label}/subjects` API.

This enables backward-compatible migration from duplicated policy entries to template-based policies, even when the template splits a single entry into multiple namespace-scoped entries.

## What's included

  - **Model**: `ImportsAlias`, `ImportsAliasTarget`, `ImportsAliases` types integrated into `Policy`
  - **API**: CRUD endpoints at `/importsAliases` and `/importsAliases/{label}`, plus `DELETE /imports`
  - **Transparent routing**: `ModifySubjects`, `ModifySubject`, `DeleteSubject`, `RetrieveSubjects`, `RetrieveSubject` resolve alias labels and fan out to all targets' `entriesAdditions`
  - **Validation**: label conflict (entry vs alias), target import existence, empty targets rejection, duplicate target deduplication, size validation after fan-out
  - **Guards**: deleting an import referenced by an alias is rejected
  - **Events**: dedicated CRUD events + fan-out events for event sourcing
  - **Ditto Protocol**: full adapter support (WebSocket/connectivity) with `PolicyResource` and `PolicyPathMatcher` integration
  - **Documentation**: OpenAPI specs, JSON schema, `basic-policy.md` section
  - **Tests**: 11 strategy tests (fan-out, reads, deletes, conflict, guard) + 148 protocol adapter round-trip tests

  ## Example

  ```json
  {
    "imports": {
      "template:roles": {
        "entries": ["operator-reactor", "operator-turbine"],
        "entriesAdditions": {
          "operator-reactor": { "subjects": { "oauth2:homer": { "type": "employee" } } },
          "operator-turbine": { "subjects": { "oauth2:homer": { "type": "employee" } } }
        }
      }
    },
    "importsAliases": {
      "operator": {
        "targets": [
          { "import": "template:roles", "entry": "operator-reactor" },
          { "import": "template:roles", "entry": "operator-turbine" }
        ]
      }
    }
  }
```

`PUT /entries/operator/subjects` transparently writes to both targets. The caller is unaware of the alias.


hu-ahmed:
## Review: importsAliases — Transparent subject fan-out                                                                                                                                                                         

 Thanks for the new improvements             
  Exercised the new alias CRUD and subject fan-out flows against a local Ditto run and rechecked the relevant policy/import code paths. The feature works, but a few issues worth addressing before merge.                        
                                                                                                                                                                                                                                  
  ---                                                                                                                                                                                                                             
                                                            
  ### 1. Top-level `PUT /policies/{id}` accepts malformed `importsAliases.targets` and silently drops invalid elements

  `ImmutableImportsAlias.fromJson()` filters non-object array members instead of rejecting them, so payloads like `"targets": ["bogus"]` are accepted through the top-level policy API and end up stored as `targets: []`. Mixed arrays like `[{valid}, 42]` silently drop the invalid element. The dedicated `PUT /importsAliases/{label}` path correctly rejects the same input with 400, so the two API paths disagree on the same payload.
                                                                                                                                                                                                                                  
  ### 2. Alias-definition CRUD does not run `PolicyCommandSizeValidator`

  Every comparable mutation strategy validates policy size after mutation, but `ModifyImportsAlias` and `ModifyImportsAliases` currently skip that step. The fan-out subject writes do validate size after expansion, so the gap is limited to the alias definition itself — practical impact is moderate since alias definitions are small, but it breaks the pattern.
                                                                                                                                                                                                                                  
  ### 3. Alias-related endpoints do not emit ETag, although the new OpenAPI docs advertise it

  In local testing, regular subject reads returned an ETag header, while `GET /entries/{alias}/subjects`, `GET /importsAliases`, and `GET /importsAliases/{label}` did not. The retrieve strategies return `Optional.empty()` for those entity tags because they resolve via `getEntryFor(label)` which only finds local entries, never aliases. The documented conditional-request behavior is not actually available on these endpoints.
                                                                                                                                                                                                                                  
  ### 4. `DELETE /entries/{alias}/subjects/{id}` for a missing subject returns 204, while the non-alias path returns 404  This looks intentional and is test-covered, so not calling it a hard bug. Still, it is inconsistent with the existing delete semantics on regular entries — every other Ditto delete operation returns 404 when the target does not exist. Worth confirming as the intended contract before merge.
                                                                                                                                                                                                                                  
  ### 5. Minor: dead code in the delete strategies          

  `DeleteImportsAliasesStrategy` discards the result of an immutable update, and `DeletePolicyImportsStrategy` constructs a variable that is never used. Harmless, but should be removed.                                         
   
  ### 6. Minor: user-visible wording still says "subject alias" in ~19 places, and ~13 places say "a imports alias" instead of "an imports alias"                                                                                 
                                                            
  Not behavioral, but worth cleaning up for consistency after the rename to `importsAliases`.                                                                                                                                     
                                                            
  ---

  > **Note:** I would not raise "alias target must validate imported entry-label existence" as a PR-specific issue. The existing direct entriesAdditions path already allows undeclared labels too — that is a broader  pre-existing gap rather than something introduced by this PR.

hu-ahmed:
LGTM!
Thanks for the quick adaptation

hu-ahmed (approved):

thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/24396347497

PR #2404, opened 2026-04-09
fix partial access filtering for subscribers with multiple auth subjects

thjaeckle:
When a subscriber has multiple authorization subjects — some with partial READ access and some with unrestricted access — the partial access path resolver incorrectly skipped the unrestricted access check. This caused the subscriber to be treated as having only partial access, stripping fields like thingId, _revision, _modified, and _created from SSE events.

The fix checks for unrestricted access before falling through to partial access filtering: if ANY subscriber subject has unrestricted access, the full payload is returned immediately.

### DIFF
--- a/gateway/service/src/test/java/org/eclipse/ditto/gateway/service/endpoints/routes/sse/ThingsSseRouteBuilderTest.java
+++ b/gateway/service/src/test/java/org/eclipse/ditto/gateway/service/endpoints/routes/sse/ThingsSseRouteBuilderTest.java
@@ -400,6 +400,40 @@ public final class ThingsSseRouteBuilderTest extends EndpointTestBase {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


hu-ahmed:
Thanks for the fix, the bug analysis is spot-on and the fix is clean and minimal. Two small things:                                                                                                                           
                                                                                                                                                                                                                                 
**Dead code**: `allSubscriberSubjectIds` is now populated but never read. It was only used in the old `subscriberSubjectIds.isEmpty()` branch which has been replaced. Should be removed.                                              
                                                                                                                                                                                                                                 
**Missing test**: Consider adding a test with the subject order reversed (full-reader first, partial-reader second) to explicitly verify order-independence. The current loop logic handles it, but having the assertion documents  the contract.

thjaeckle:
@hu-ahmed done ✅

hu-ahmed:
Thanks 
LGTM!

PR #2408, opened 2026-04-12
fix aggregation metrics retaining stale values for vanished group-by …

hu-ahmed:
## Summary
                                                                                                                                                                             
  - **Fix stale gauges**: When a group-by bucket vanishes from MongoDB `$group` results,                                                                                     
    the corresponding Prometheus gauge is now zeroed and removed within one scrape cycle                                                                                     
    instead of waiting up to 2-3 hours                                                                                                                                       
  - **Per-metric cleanup timers**: Each metric's cleanup timer now uses its own scrape
    interval instead of the global maximum across all metrics                                                                                                                
  - **Timeout safety**: Added 2-minute `completionTimeout` on the aggregation stream to
    prevent hung MongoDB connections from permanently leaking actor closures

### DIFF
--- a/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/AggregateThingsMetricsActor.java
+++ b/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/starter/actors/AggregateThingsMetricsActor.java
@@ -15,27 +15,21 @@
 package org.eclipse.ditto.thingsearch.service.starter.actors;
 
 import java.time.Duration;
-import java.util.concurrent.CompletableFuture;
 
 import org.apache.pekko.NotUsed;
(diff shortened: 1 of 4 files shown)


thjaeckle (approved):
LGTM - thanks for the fixes .. this improves, summarizing:

When a group-by bucket disappears from MongoDB aggregation results (e.g., no more things match a particular filter combination), the corresponding Prometheus gauge previously retained its last stale value for up to 2-3 hours.  
This PR makes stale gauges get zeroed and removed within one scrape cycle by batching all results and reconciling against known gauges. It also fixes cleanup timers to be per-metric instead of globally throttled to the slowest metric's interval, and adds a 2-minute timeout to prevent hung MongoDB connections from leaking permanently.

PR #2409, opened 2026-04-13
Dynamically scope WoT Thing Description by user permissions

### LINKED ISSUE #2144: Dynamically reduce the scope of a requested WoT Thing Description in Ditto to the user's permissions
Currently, any authenticated (not authorized - this is currently not checked) user may fetch the WoT "Thing Description" of a Thing which links to a WoT Thing Model in its `definition` field.

This is done providing the `accept` header: `Accept: application/td+json` (content negotiation).  
However:
* as said, any authenticated user may fetch the TD of any known thingId
* and the user gets back the complete functionality of the thing - which may be much more than the user is able to see or use from the thing's API

So the idea of this issue is to scope the generated TD based on the user's permissions:
* only include the WoT "properties" which the user is able to READ
* dynamically generate "readOnly" based on the permissions
* only add "actions" which the user is allowed to invoke (WRITE to)

thjaeckle:
Filter the generated WoT Thing Description in the enforcement layer based on the requesting user's policy permissions.  
Resolves: #2144

- Properties the user cannot READ are removed from the TD
- Properties the user cannot WRITE are marked readOnly
- Actions the user cannot invoke (WRITE on message resource) are removed
- Events the user cannot subscribe to (READ on message resource) are removed
- Sub-model links to inaccessible features are removed
- Root-level forms are cleaned up when affordances are filtered

Added behind feature toggle `WOT_TD_PERMISSION_FILTERING_ENABLED` (default: true). Setting the env var
`DITTO_DEVOPS_FEATURE_WOT_TD_PERMISSION_FILTERING_ENABLED=false` restores the previous behavior of unfiltered TDs.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/signals/FeatureToggle.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/signals/FeatureToggle.java
@@ -74,6 +74,16 @@ public final class FeatureToggle {
             "ditto.devops.feature.policy-enforcement-use-throughput-optimized-evaluator-enabled";
 
     /**
+     * System property name of the property defining whether WoT Thing Description responses are filtered based on
+     * the requesting user's policy permissions. When enabled, TDs only contain properties, actions, and events the
+     * user is authorized to access. When disabled, TDs are returned unfiltered (legacy behavior).
+     *
+     * @since 3.9.0
+     */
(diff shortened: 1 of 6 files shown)


thjaeckle:
System tests passed (except 1 known flaky test): https://github.com/eclipse-ditto/ditto/actions/runs/24351770952

PR #2411, opened 2026-04-13
fix: collapse partial-access-paths header to ancestor pointers

hu-ahmed:
## Problem                                                

 Testing Ditto 3.9.0-M2, we observed ~10 000 `BufferOverflowException` per hour                                                                                                       
  on the things-service logged by `org.apache.pekko.remote.artery.Encoder`:
    
```                                                                                                                                                                         
  Failed to serialize message [org.eclipse.ditto.internal.utils.pubsub.api.PublishSignal]                                                                                    
  Caused by: java.nio.BufferOverflowException                                                                                                                                
    at java.base/java.nio.ByteBuffer.put(ByteBuffer.java:1220)                                                                                                               
    at ...ByteBufferOutputStream.write(...)                                                                                                                                  
    at ...AbstractJsonifiableWithDittoHeadersSerializer.toBinary(...)                                                                                                        
    at ...MessageSerializer$.serializeForArtery(...)                                                                                                                         
 ```
                                                                                                                                                                        
  Each failure silently drops the `ThingEvent` for every cluster subscriber                                                                                                  
  — SSE, WebSocket, connectivity targets, and thingsearch all stop seeing                                                                                                    
  changes for the affected Thing.                                                                                                                                            
                                                                                                                                                                             
  ## Root cause                                                                                                                                                              
                                                            
  `TreeBasedPolicyEnforcer.getAccessiblePathsForSubjects` (added in #2287)                                                                                                   
  flattens the Thing JSON to leaf pointers. For a Thing whose features
  carry a few hundred child entries with several properties each, the                                                                                                        
  per-subject leaf set reaches several thousand entries. Serialised into
  the `ditto-partial-access-paths` internal DittoHeader (indexed JSON                                                                                                        
  format), the header alone exceeds 350 KiB, well over the default Pekko                                                                                                     
  Artery 256 KiB frame.                                                                                                                                                      
                                                                                                                                                                             
  Example measurement (synthetic reproduction): a Thing with 1000 child                                                                                                      
  entries under each of two features, and a partial reader granted the                                                                                                       
  two feature resources:                                                                                                                                                     
                                                                                                                                                                             
  | | Path count | Indexed header bytes |                                                                                                                                    
  |---|---:|---:|                                                                                                                                                            
  | Before fix | 5 011 | 367 564 |                                                                                                                                           
  | After fix  |     2 |      94 |                          
                                                                                                                                                                             
  ## Fix
                                                                                                                                                                             
  Two coordinated changes:                                  

  **1. Producer — `PartialAccessPathCalculator` collapses leaves to ancestor pointers**                                                                                      
   
  After the enforcer returns the per-subject leaf set, walk the Thing JSON                                                                                                   
  top-down; if every leaf of a subtree is in the subject's accessible set,
  replace them with a single ancestor pointer. Mirrors the policy                                                                                                            
  semantics — a READ grant on an ancestor implicitly grants read on every                                                                                                    
  descendant.                                                                                                                                                                
                                                                                                                                                                             
  **2. Consumer — `JsonPartialAccessFilter` honours ancestor grants**
                                                                                                                                                                             
  New `PathTrie.hasAncestorMatch(path)` check. When a field has an                                                                                                           
  accessible ancestor the whole subtree is included, keeping filtering
  semantically identical to pre-collapse leaf enumeration.

### DIFF
--- a/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
+++ b/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
@@ -222,6 +222,14 @@ public final class AdaptablePartialAccessFilter {
             return true;
         }
 
+        final int eventLevelCount = eventPath.getLevelCount();
+        for (int i = 1; i < eventLevelCount; i++) {
+            final JsonPointer prefix = eventPath.getPrefixPointer(i).orElse(null);
+            if (prefix != null && accessiblePaths.contains(prefix)) {
+                return true;
+            }
+        }
+
         return false;
     }
 
--- a/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/JsonPartialAccessFilter.java
+++ b/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/JsonPartialAccessFilter.java
@@ -184,6 +184,16 @@ public final class JsonPartialAccessFilter {
     /**
      * Recursively filters a JSON object based on accessible paths.
      * Uses a PathTrie for efficient prefix matching.
+     * <p>
+     * A field is included when any of the following holds:
+     * <ul>
+     *   <li>The field path is an exact match in the accessible set.</li>
+     *   <li>The field has an accessible descendant (partial match — recurse).</li>
+     *   <li>A proper ancestor of the field path is an exact match — in which case the entire
+     *       subtree rooted at this field is included as-is (no further filtering needed). This
+     *       mirrors Ditto policy semantics: a READ grant on an ancestor grants read on every
+     *       descendant.</li>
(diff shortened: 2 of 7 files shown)


thjaeckle:
Nice fix! One potential gap I noticed:

`AdaptablePartialAccessFilter.isPathAccessibleForNonObjectPayload` (line 211) only checks exact match and root match — it does not check ancestor prefixes. After this PR, the accessible paths set can contain ancestor pointers like `/features/A` instead of leaf paths. A non-object (scalar) event at e.g. `/features/A/properties/x` would fail both checks and return `false`, potentially causing missed event delivery for scalar change events under an ancestor grant.

The same `hasAncestorMatch` logic that was added to the object-filtering path in `filterJsonRecursive` should probably also be applied here.

thjaeckle:
System tests passed (except 1 known test to be flaky): https://github.com/eclipse-ditto/ditto/actions/runs/24387481305

thjaeckle (approved):

PR #2396, opened 2026-04-07
Surface enforcement/validation errors for fire-and-forget commands

thjaeckle:
Previously, fire-and-forget commands (timeout=0, no ack requests) immediately returned HTTP 202 Accepted without waiting for enforcement or validation results. This meant policy violations and WoT validation errors were silently swallowed.

Now the gateway waits for the enforcement/validation result before responding. If an error arrives (e.g. policy check fails), it is returned to the client. If no error arrives within the default command timeout, HTTP 202 is returned.

Resolves #2392

PR #2413, opened 2026-04-14
Fix fire-and-forget commands delayed by full request timeout

thjaeckle:
The fire-and-forget enforcement change (PR #2396) waits for enforcement errors before returning HTTP 202. However, it used the full command default timeout (= request-timeout = 60s) as the wait duration. Since things-service does not send a response for fire-and-forget commands on success, the gateway waited the full 60s before responding.

Add a dedicated `fire-and-forget-enforcement-timeout` config option (default: 5s) and use it instead. Enforcement errors arrive within milliseconds, so 5s is sufficient to catch them without delaying successful fire-and-forget responses.


thjaeckle:
This solves the integration-test errors on master.  
This run has 0 test errors: https://github.com/eclipse-ditto/ditto/actions/runs/24391511654

PR #2414, opened 2026-04-14
fix: checkPermissions ignoring permissions from imported policies

hu-ahmed:
**Summary**
                                                                                                                                                                                                                                  
  - `/api/2/checkPermissions `returned false for policies that derive permissions entirely from imported policies, even though the actual operations (e.g., mute) succeeded                                                         
  - Root cause: `PolicyCheckPermissionsStrategy` used `PolicyEnforcer.of(entity)` (raw policy only), while normal enforcement uses PolicyEnforcerProvider which resolves imports                                                      
  - Fix: constructor-inject `PolicyEnforcerProvider` into the strategy, matching the established pattern for strategies that need Pekko infrastructure


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/24405563608

thjaeckle (approved):
LGTM, thanks for the fix 👍

PR #2418, opened 2026-04-17
Fix Fluency thread leak in connection logger publisher

hu-ahmed:
## Problem                                                                                                                                                                                                                      
                                                                                                                                                                                                                                  
  Enabling the Fluency-based connection log publisher                                                                                                                                                                             
  (`CONNECTIVITY_LOGGER_PUBLISHER_ENABLED=true`) causes an unbounded thread leak in the connectivity service. On our dev cluster with 37 connections,
  ~2,100 orphaned `Flusher.runLoop` threads accumulated within 17 hours,  with pool IDs climbing past 24,000.                       
                                                                                                                                                                                                                                  
  Two root causes:                                                                                                                                                                                                                
                                                                                                                                                                                                                                  
  1. **Orphaned Fluency instances.** `ConnectionLogger.getInstance()` and `ConnectionLoggerRegistry.fromConfig()` created a new                                                                                                                                                                        
     `ConnectionLoggerRegistry` on every call, each constructing a new Fluency client with its own flusher thread pool. Because the logger                                                                                                                                                          
     cache (`LOGGERS`) is static, `computeIfAbsent` returned already-cached loggers — the newly created Fluency instance was immediately orphaned with no owner to close it, but its flusher thread kept running.
                                              
  2. **Shared Fluency closed by individual loggers.** `FluentPublishingConnectionLogger.close()` called `fluencyForwarder.close()` on the shared Fluency instance. When `invalidateLoggers()` closed one connection's loggers, it shut down the Fluency forwarder for all other connections. The Fluency library then spawned new internal sender threads trying to recover, amplifying the leak.                                
                                                                                                                                                                                                                                  
  ## Solution                                                                                                                                                                                                                     
                                                                                                                                                                                                                                  
  - **Cache registries by config.** `ConnectionLoggerRegistry.fromConfig()` now uses a static `INSTANCES` map with `computeIfAbsent`, so only one registry — and one Fluency client — exists per configuration for the lifetime of the JVM.                                                                                                                                                                                                          
  - **No-op logger close.** `FluentPublishingConnectionLogger.close()` no longer closes the shared Fluency instance. The forwarder is owned by the registry, not by individual loggers.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/BaseClientActor.java
@@ -275,7 +275,7 @@ public abstract class BaseClientActor extends AbstractFSMWithStash<BaseClientSta
         protocolAdapter = protocolAdapterProvider.getProtocolAdapter(null);
         final var monitoringConfig = connectivityConfig().getMonitoringConfig();
         connectionCounterRegistry = ConnectivityCounterRegistry.newInstance(connectivityConfig);
-        connectionLoggerRegistry = ConnectionLoggerRegistry.fromConfig(monitoringConfig.logger());
+        connectionLoggerRegistry = ConnectionLoggerRegistry.fromConfig(monitoringConfig.logger(), system);
         connectionLoggerRegistry.initForConnection(connection);
         connectionLogger = connectionLoggerRegistry.forConnection(connection.getId());
         connectionCounterRegistry.initForConnection(connection);
@@ -1956,6 +1956,7 @@ public abstract class BaseClientActor extends AbstractFSMWithStash<BaseClientSta
                 outboundMappingProcessorActor,
                 getSelf(),
                 getContext(),
+                getContext().getSystem(),
                 connectivityConfig,
                 dittoHeadersValidator,
                 getResponseValidationFailureConsumer());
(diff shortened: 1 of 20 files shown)


thjaeckle:
Thanks a lot for the fix @hu-ahmed 

I noticed that the Fluency instance is not closed e.g. during graceful shutdown. We might want to flush the logs during graceful shutdown so that no logs are "dropped".

I think we could follow the `SshClientProvider` pattern (defining a Pekko Extension) to extract the Fluency singleton into an Extension and adding the `CoordinatedShutdown` task there as well.

Could you check that, please?

hu-ahmed:
Sure, I did the cheap easy solution :), but it's not the right way

thjaeckle:
Triggered system-tests as we have some tests for the log-forwarding there: https://github.com/eclipse-ditto/ditto/actions/runs/24563397908

thjaeckle:
@hu-ahmed system-tests passed - but strangely unit tests were now twice failing with the same error

thjaeckle (commented):
## Review of updated PR

Nice work following the `SshClientProvider` pattern — the `FluencyProvider` Extension correctly centralizes the Fluency singleton lifecycle and integrates with `CoordinatedShutdown`. The thread leak root cause (each `ConnectionLoggerRegistry` creating its own `Fluency` instance) is properly addressed.

### Issues

**1. Missing `checkNotNull(actorSystem)` in `DefaultConnectionMonitorRegistry.fromConfig`**

The updated Javadoc says `@throws java.lang.NullPointerException if any argument is {@code null}`, but the implementation only checks `connectivityConfig`:

```java
public static DefaultConnectionMonitorRegistry fromConfig(final ConnectivityConfig connectivityConfig,
        final ActorSystem actorSystem) {
    checkNotNull(connectivityConfig);
    // actorSystem is not checked — will NPE later with an unclear message
```

Add `checkNotNull(actorSystem);` to match the Javadoc contract and `ConnectionLoggerRegistry.fromConfig` which does check both.

Overall this looks good — just the small item above, then 👍 from my side.

thjaeckle (approved):
LGTM, thanks a lot @hu-ahmed 👍

PR #2421, opened 2026-04-20
harden GitHub Actions against supply chain attacks

hu-ahmed:
**Background**                                                                                                                                                                                                                      
                                                                                                                                                                                                                                  
  The https://www.stepsecurity.io/blog/harden-runner-detection-aqua-security-trivy-action-supply-chain-attack — where an attacker force-pushed 76 of 77 version tags in aquasecurity/trivy-action to malicious commits — showed that any CI workflow consuming third-party actions by mutable reference (tags, branches) is one upstream compromise away from shipping attacker code. Ditto's workflows were in that category.                                  
                                                                                                                                                                                                                                  
  This PR applies the hardening baseline recommended by the Eclipse Foundation Security Team in https://blog.mbarbero.com/ (Mikaël Barbero, 2026-03-24), adapted to Ditto's specific workflows.                                   
   
  **What changed**                                                                                                                                                                                                                    
                                                            
  1. Every third-party action pinned to commit SHA. 17 distinct actions across 10 workflows. Each uses: owner/name@vX is now uses: owner/name@<full-40-char-sha> # <exact-semver>. SHAs resolved and verified with https://github.com/suzuki-shunsuke/pinact (the tool cited in the Eclipse blog post) — not by hand. A tag can be silently repointed; a SHA cannot.
                                                                                                                                                                                                                                  
  2. .github/dependabot.yml added. Weekly github-actions ecosystem updates, grouped into a single PR, with a 7-day cooldown so we don't adopt a compromised release during the first 24 hours of an incident. Config matches the  Eclipse security team's template verbatim.
                                                                                                                                                                                                                                  
  ▎ Dependabot still needs to be enabled at the repo level (Settings → Security → Dependabot). The file alone is not enough — Eclipse Foundation admins can flip this on.                                                         
   
  3. step-security/harden-runner added in audit mode to every GitHub-hosted job (9 workflows, 11 jobs). Records network egress and file modifications without blocking. Had this been active during the Trivy incident, exfiltration to the typo-squatted scan.aquasecurtiy.org domain would have been immediately visible. The self-hosted ditto-runner job in system-tests.yml is intentionally excluded — harden-runner is designed for GitHub-hosted runners and needs runner-owner validation before enabling.                                                                                                                                                                     
                                                            
  4. Docker image pinned by digest. maven:3.9.12-eclipse-temurin-25 → maven:3.9.12-eclipse-temurin-25@sha256:4f82a03a... in all 4 uses in system-tests.yml. Container tags have the same mutability problem as action tags.       
   
  5. Supply-chain linters added/run:                                                                                                                                                                                              
  - https://github.com/boostsecurityio/poutine run locally: zero failures on unpinnable_action, injection, default_permissions_on_risky_events, known_vulnerability_in_build_component, and all other blocking rules. The 6  note-level findings are informational flags on unverified-creator actions (dorny, helm, jitterbit, peaceiris) — widely used but not on GitHub's Marketplace verified list. No action required.                                  
  - [zizmor](https://github.com/zizmorcore/zizmor) added as a new CI job (`.github/workflows/zizmor.yml`). Runs on every PR (and push to `master`) that touches  `.github/workflows/**` or `.github/dependabot.yml`, uploads SARIF to the GitHub Security tab, and is **gating** — the action fails CI on new findings. Accepted findings  are tracked in `.github/zizmor.yml` with a written justification per entry (currently: scoped `cache-poisoning` suppressions for workflows that don't publish artifacts or don't populate a cache). This complements harden-runner: zizmor catches issues at lint time, harden-runner catches anomalies at runtime.
   6. **Default `GITHUB_TOKEN` permissions narrowed to read-only.** Every workflow now declares an explicit top-level `permissions: contents: read` block, with job-level  elevation only where actually needed: `gh-pages.yml` keeps `contents: write` on the deploy job (publishes to the `gh-pages` branch), and `system-tests.yml` adds `checks:  write` on its single job (required by `dorny/test-reporter` to publish results to the Checks API). The DockerHub publishers (`push-dockerhub*.yml`, `helm-chart-release.yml`, `docker-nightly.yml`) need no elevation — they auth to Docker Hub via `DOCKER_HUB_TOKEN`, not `GITHUB_TOKEN`. This clears 12 medium-severity  `excessive-permissions` findings from zizmor and makes least-privilege explicit at the file level rather than relying on the org/repo default.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
+    open-pull-requests-limit: 10
+    labels:
+      - dependencies
+    commit-message:
+      prefix: ci
+      include: scope
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/docker-nightly.yml
+++ b/.github/workflows/docker-nightly.yml
@@ -28,6 +28,8 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       -
         name: Set up QEMU
         uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
--- a/.github/workflows/docker-nightly.yml
+++ b/.github/workflows/docker-nightly.yml
@@ -21,15 +21,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       -
+        name: Harden Runner
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        with:
+          egress-policy: audit
+      -
         name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       -
         name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
       -
         name: Set up Docker Buildx
         id: buildx
-        uses: docker/setup-buildx-action@v3 # setup buildx in order to do build and push multi-architecture images
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0 # setup buildx in order to do build and push multi-architecture images
       -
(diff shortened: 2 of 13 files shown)


thjaeckle:
Thanks for the thorough hardening work — the SHA pinning, Harden-Runner audit-mode rollout, image-digest pinning, and zizmor gating all look good.

One item from the Eclipse blog post that I think should still be addressed in this PR before merging: **default `GITHUB_TOKEN` to read-only.**

Currently only `.github/workflows/zizmor.yml` declares a top-level `permissions:` block. The other ten workflows have no explicit token-permissions declaration, so they rely on the org/repo default. Best practice (and what the blog calls out) is to declare it explicitly per workflow:

```yaml
permissions:
  contents: read
```

…at the top of each workflow, with job-level overrides where writes are actually needed — e.g.:
- `gh-pages.yml` already has a job-level `contents: write` (publish to gh-pages branch) — keep it.
- `helm-chart-release.yml`, `push-dockerhub.yml`, `push-dockerhub-on-demand.yml`, `docker-nightly.yml` push artifacts but don't actually need elevated `GITHUB_TOKEN` permissions for that (Docker Hub uses `DOCKER_HUB_TOKEN`), so a top-level `contents: read` is sufficient.
- `system-tests.yml` only reads code and uploads artifacts — `contents: read` + `actions: read` should be enough; check if `actions/upload-artifact@v4` needs anything beyond default.

Adding this is cheap, makes the principle of least privilege explicit at the file level (so zizmor and human reviewers can verify it without having to know the org-level default), and is the one item from the blog's list that's both clearly in-scope for this PR and not yet done.

Side note: the PR description states zizmor runs with `continue-on-error: true`, but the actual workflow doesn't set that — it's gating. Worth updating the description to match.

thjaeckle (approved):
LGTM, thanks a lot @hu-ahmed 🥳

PR #2422, opened 2026-04-21
Add transitiveImports for selective multi-level policy import resolution

### LINKED ISSUE #2420: Support transitive resolution of policy imports via transitiveImports
## Summary

Policy imports are currently resolved one level deep: if policy A imports from policy B, A gets B's inline entries only. If B itself imports from policy C (e.g., a shared template with `entriesAdditions`), those entries are **not** visible to A.

This makes it impossible to build multi-level policy template hierarchies where:
- A **global template** defines roles with resources, namespaces, and `allowedImportAdditions`
- **Intermediate policies** import from the template and add concrete subjects via `entriesAdditions`
- **Leaf policies** import from the intermediate policy and need to see the combined result (resources from the template + subjects from the intermediate)

## Why this requires `entriesAdditions`

Transitive import resolution is a natural complement to `entriesAdditions` (introduced in Ditto 3.9.0) — and would be meaningless without it. Before `entriesAdditions`, imported entries were always taken as-is from the imported policy's persisted entries. There was no mechanism for an intermediate policy to *add* anything to entries it imported from elsewhere, so resolving through it would yield nothing beyond what a direct import already provides.

`entriesAdditions` changes this by allowing an intermediate policy to hold **per-import state** (subjects, resources, namespaces) that only materializes during resolution. A consuming policy that directly imports the global template gets the raw entries (empty subjects). A consuming policy that directly imports the intermediate policy gets nothing (its inline entries are empty). Only by resolving *through* the intermediate policy can the consuming policy obtain the combined result: template resources merged with the intermediate policy's subject additions.

## Proposal

Add an optional `transitiveImports` field to policy imports. It contains an explicit list of policy IDs that the directly imported policy itself imports from, which should be resolved before extracting entries.

### Example

**Global template** (`acme:fleet-roles`):
```json
{
  "policyId": "acme:fleet-roles",
  "entries": {
    "driver": {
      "subjects": {},
      "resources": {
        "thing:/features/location": { "grant": ["READ"], "revoke": [] },
        "thing:/features/fuel": { "grant": ["READ"], "revoke": [] },
        "message:/features/fuel/inbox": { "grant": ["WRITE"], "revoke": [] }
      },
      "namespaces": ["acme.vehicle"],
      "allowedImportAdditions": ["subjects"],
      "importable": "implicit"
    }
  }
}
```

**Regional fleet policy** (`acme:fleet-west`) imports the template and adds drivers:
```json
{
  "policyId": "acme:fleet-west",
  "imports": {
    "acme:fleet-roles": {
      "entriesAdditions": {
        "driver": {
          "subjects": {
            "oauth2:alice@acme.com": { "type": "employee" },
            "oauth2:bob@acme.com": { "type": "employee" }
          }
        }
      }
    }
  },
  "entries": {}
}
```

**Vehicle policy** needs the `driver` entry with resources from the template AND subjects from the fleet policy:
```json
{
  "policyId": "acme.vehicle:truck-42",
  "imports": {
    "acme:fleet-west": {
      "entries": ["driver"],
      "transitiveImports": ["acme:fleet-roles"]
    }
  }
}
```

### Resolution behavior

1. Load `acme:fleet-west` (persisted state — inline entries are empty)
2. Because `transitiveImports` lists `acme:fleet-roles`, resolve that import on the loaded fleet policy first
3. This merges `fleet-roles`'s `driver` entry into `fleet-west`, applying `fleet-west`'s `entriesAdditions` (subjects)
4. Vehicle policy extracts `driver` from the enriched result — getting both the resources from the template and the subjects from the fleet policy

Without `transitiveImports`, this composition would be impossible: a direct import of `fleet-roles` yields the entry without subjects, and a direct import of `fleet-west` yields nothing (no inline entries).

### API

- `GET/PUT /api/2/policies/{policyId}/imports/{importedPolicyId}/transitiveImports` — manage the transitive imports array
- The field is also readable/writable as part of the full import at `GET/PUT /api/2/policies/{policyId}/imports/{importedPolicyId}`

### Key design properties

- **Explicit whitelist** — only the listed policy IDs are resolved, not a recursive flag. This prevents surprise permission expansion and makes the dependency graph auditable.
- **Search index tracking** — transitive policy IDs are tracked in `__referencedPolicies`, so changes to the template trigger re-indexing of dependent things.
- **Cache invalidation** — transitive policy IDs are registered in the cache dependency map, enabling cascade invalidation.
- **Backward compatible** — absent `transitiveImports` ≡ current single-level behavior.
- **Cycle safe** — the importing policy's own ID must not appear in `transitiveImports` (validated at write time).

thjaeckle:
Adds a new optional `transitiveImports` field to policy imports that enables multi-level import resolution. This complements `entriesAdditions` by allowing consuming policies to resolve through intermediate policies that add subjects to template-defined entries.

Without this feature, policy imports are single-level: if policy A imports from B, and B imports from C with entriesAdditions, A cannot see C's entries enriched with B's subject additions. The `transitiveImports` field explicitly lists which of the imported policy's own imports should be resolved first.

Changes:
- Model: EffectedImports, PolicyImport, PoliciesModelFactory, PolicyImporter
- Validation: cycle prevention in PolicyImportsValidator, including dedicated validateTransitiveImports for the sub-resource endpoint
- Resolution: cycle detection via visited set + depth limit, single-prefix labels (no double-prefixing in transitive chains), O(n) allOf merge replacing O(n²) reduce/combineSets, no temp Policy copies in transitive resolution
- HTTP API: GET/PUT /imports/{id}/transitiveImports endpoint
- Signal commands: Retrieve/Modify + responses, event, strategies
- Protocol mapping: Ditto Protocol adapter for all new signals
- Cache: PolicyEnforcerCache tracks transitive IDs for cascade invalidation, deregisters stale mappings on policy reload, guards against NPE on exceptional cache load
- Search: ResolvedPolicyCacheLoader uses ConcurrentHashMap.newKeySet for thread safety, awaits transitive tag futures before returning; BackgroundSyncStream tracks transitive policy IDs in __referencedPolicies
- Documentation: basic-policy.md with examples, deeper nesting explanation
- OpenAPI: schema, path definition, api-2-index registration
- JSON Schema: policy.json updated
- Tests: PolicyImporterTest (2/3-level resolution, cycle detection, EXPLICIT importable with transitive), cache invalidation, search index consistency, model serialization round-trips

Closes: #2420

### DIFF
--- a/documentation/src/main/resources/jsonschema/policy.json
+++ b/documentation/src/main/resources/jsonschema/policy.json
@@ -57,6 +57,15 @@
                   }
                 }
               }
+            },
+            "transitiveImports": {
+              "title": "Transitive imports",
+              "type": "array",
+              "description": "List of policy IDs from the imported policy's own imports that should be resolved transitively before extracting entries. This enables multi-level import chains where a template policy defines resources and an intermediate policy adds subjects via entriesAdditions.",
+              "items": {
+                "type": "string",
+                "description": "Policy ID of a policy that the imported policy itself imports from."
+              }
             }
           }
         }
--- a/documentation/src/main/resources/openapi/ditto-api-2.yml
+++ b/documentation/src/main/resources/openapi/ditto-api-2.yml
@@ -7618,6 +7618,141 @@ paths:
                 $ref: '#/components/schemas/AdvancedError'
         '412':
           $ref: '#/components/responses/PreconditionFailed'
+  '/api/2/policies/{policyId}/imports/{importedPolicyId}/transitiveImports':
+    get:
+      summary: Retrieve the transitive resolution policy IDs of a specific policy import
+      description: |-
+        Returns the "transitiveImports" array of the policy import identified by the `policyId` path
+        parameter and the `importedPolicyId` path parameter.
+
+        The array lists policy IDs from the imported policy's own imports that should be resolved
+        transitively before extracting entries. This enables multi-level import chains.
+      tags:
+        - Policies
+      parameters:
+        - $ref: '#/components/parameters/PolicyIdPathParam'
+        - $ref: '#/components/parameters/ImportedPolicyIdPathParam'
+        - $ref: '#/components/parameters/IfMatchHeaderParamHash'
+        - $ref: '#/components/parameters/IfNoneMatchHeaderParam'
+        - $ref: '#/components/parameters/TimeoutParam'
+      responses:
+        '200':
+          description: The request successfully returned. The transitiveImports array is returned.
+          headers:
+            ETag:
+              description: |-
+                The (current server-side) ETag for this (sub-)resource. For top-level resources it is in the format
+                "rev:[revision]", for sub-resources it has the format "hash:[calculated-hash]".
+              schema:
+                type: string
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TransitiveImports'
+        '304':
+          $ref: '#/components/responses/NotModified'
+        '400':
+          description: |-
+            The request could not be completed. Possible reasons:
+
+              * the `policyId` or the `importedPolicyId` does not conform to the namespaced entity ID notation (see [Ditto documentation on namespaced entity IDs](https://www.eclipse.dev/ditto/basic-namespaces-and-names.html#namespaced-id))
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '401':
+          description: The request could not be completed due to missing authentication.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '404':
+          description: |-
+            The request could not be completed. The policy with the given ID or
+            the policy import was not found in the context of the authenticated
+            user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '412':
+          $ref: '#/components/responses/PreconditionFailed'
+    put:
+      summary: Modify the transitive resolution policy IDs of a specific policy import
+      description: |-
+        Modify the "transitiveImports" array of the policy import identified by the `policyId` path
+        parameter and the `importedPolicyId` path parameter.
+
+        The array lists policy IDs from the imported policy's own imports that should be resolved
+        transitively before extracting entries.
+      tags:
+        - Policies
+      parameters:
+        - $ref: '#/components/parameters/PolicyIdPathParam'
+        - $ref: '#/components/parameters/ImportedPolicyIdPathParam'
+        - $ref: '#/components/parameters/IfMatchHeaderParamHash'
+        - $ref: '#/components/parameters/IfNoneMatchHeaderParam'
+        - $ref: '#/components/parameters/IfEqualHeaderParam'
+        - $ref: '#/components/parameters/TimeoutParam'
+        - $ref: '#/components/parameters/ResponseRequiredParam'
+      responses:
+        '204':
+          description: The transitiveImports array was successfully updated.
+          headers:
+            ETag:
+              description: |-
+                The (current server-side) ETag for this (sub-)resource. For top-level resources it is in the format
+                "rev:[revision]", for sub-resources it has the format "hash:[calculated-hash]".
+              schema:
+                type: string
+        '400':
+          description: |-
+            The request could not be completed. Possible reasons:
+
+              * the `policyId` or the `importedPolicyId` does not conform to the namespaced entity ID notation (see [Ditto documentation on namespaced entity IDs](https://www.eclipse.dev/ditto/basic-namespaces-and-names.html#namespaced-id))
+              * the JSON body is not a valid JSON array of policy ID strings
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '401':
+          description: The request could not be completed due to missing authentication.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '403':
+          description: |-
+            The request could not be completed. Possible reasons:
+            * the caller has insufficient permissions.
+              You need `WRITE` permission on the `policy:/imports/{importedPolicyId}` resource,
+              without any revoke in a deeper path of the policy resource.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '404':
+          description: |-
+            The request could not be completed. The policy with the given ID or
+            the policy import was not found in the context of the authenticated
+            user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AdvancedError'
+        '412':
+          $ref: '#/components/responses/PreconditionFailed'
+        '413':
+          $ref: '#/components/responses/EntityTooLarge'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/TransitiveImports'
+            example:
+              - 'org.eclipse.ditto:policy-template'
+        description: JSON array of policy IDs to resolve transitively.
+        required: true
   '/api/2/policies/{policyId}/importsAliases':
     get:
       summary: Retrieve all imports aliases of a policy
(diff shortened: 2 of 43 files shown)


hu-ahmed (commented):
## Review: transitiveImports for multi-level policy import resolution                                                                                                                                                           
                                                                                                                                                                                                                                
  **Summary:** 3 blocking, 6 high, 5 medium. Findings marked *[verified-live]* were reproduced against a running gateway.                                                                                                         
   
  ---                                                                                                                                                                                                                             
                  
  ### Blocking

  **B1. Ditto Protocol path `/imports/<id>/transitiveImports` is not routable** *[verified-live]*                                                                                                                                 
   
  `PolicyResource.java:31-49` and `PolicyPathMatcher.java:38-56` have no entry for transitiveImports. The `ResourceMap` at `PolicyResource.java:54-62` does not register `EffectedImports.JsonFields.TRANSITIVE_IMPORTS`. HTTP works because the route constructs commands directly; WebSocket and Connectivity fail.
                                                                                                                                                                                                                                  
  Reproduced via WS: server returns `{"status":400,"error":"protocoladapter:unknown.path","message":"The path '/imports/com.example:policyB/transitiveImports' does not correspond to any known command, response or event."}`.   
   
  Fix: add `POLICY_IMPORT_TRANSITIVE_IMPORTS` to the enum, register it in the `ResourceMap` under `POLICY_IMPORT` keyed on `EffectedImports.JsonFields.TRANSITIVE_IMPORTS`, and map to `"policyImportTransitiveImports"` in  `PolicyPathMatcher`. Add parametrized adapter tests.
                                                                                                                                                                                                                                  
  **B2. Silent `.filter(JsonValue::isString)` drops non-string array elements** *[verified-live]*                                                                                                                                 
   
  - `ImmutableEffectedImports.java:141-146` (parent `/imports/<id>` PUT)                                                                                                                                                          
  - `PolicyImportsRoute.java:288-295` (sub-resource PUT)
                                                                                                                                                                                                                                  
  Reproduced: `PUT ["com.example:policyC", 42, null, true]` returns 204; readback is `["com.example:policyC"]`. Same via parent endpoint. Same anti pattern as the prior `namespaces` issue — must validate and reject with `PolicyImportInvalidException`, not filter.                                                                                                                                                                                     
                                                                                                                                                                                                                                  
  **B3. `invalidateConditionally` destroys the reverse-mapping regardless of outcome**                                                                                                                                            
   
  `PolicyEnforcerCache.java:163-165` calls `policyIdToImportingMap.remove(policyId)` even when `valueCondition` is false and the primary cache entry is not invalidated. Subsequent changes to `policyId` no longer cascade to importers that remain cached.
                                                                                                                                                                                                                                  
  ---                                                                                                                                                                                                                             
   
  ### High                                                                                                                                                                                                                        
                  
  **H1. Cycle test asserts only `isNotNull()`**                                                                                                                                                                                   
   
  `PolicyImporterTest.java:1199-1204` — passes for empty sets and wrong sets alike. Combined with H2 this is a ticking bomb. Assert the expected truncated entry set.                                                             
                  
  **H2. Silent truncation at depth 10 and silent cycle drop** *[verified-live]*                                                                                                                                                   
                  
  `PolicyImporter.java:167-169` and `172-173` drop data with no log, no metric, no error. Reproduced 3-policy cycle `A→B(t=[C]), B→C(t=[A]), C→A(t=[B])`: thing GET returns 200 in ~14ms with no diagnostic. Also `MAX_TRANSITIVE_RESOLUTION_DEPTH = 10` is hardcoded package-private.
                                                                                                                                                                                                                                  
  Fix: log WARN with correlation ID on both paths; move the constant to `PolicyConfig`.                                                                                                                                           
   
  **H3. Write-time validation only catches one-hop self-reference** *[verified-live]*                                                                                                                                             
                  
  `PolicyImportsValidator.java:32-97` — cross-policy cycles (A→B→C→A) pass. Reproduced: three policies with 3-way cycle all PUT with 204. Document the limitation in Javadoc and `basic-policy.md`.                               
                  
  **H4. Cache registration race in `whenCompleteAsync`**                                                                                                                                                                          
                  
  `PolicyEnforcerCache.java:52-77` registers the importer→imported mapping *after* `asyncLoad` completes. If the imported policy mutates between load and callback, the importer is cached stale with no mapping, and the next invalidation misses it. Register eagerly before load, or use a revision check.
                                                                                                                                                                                                                                  
  **H5. Order-dependent merge on conflicting labels**                                                                                                                                                                             
   
  `PolicyImporter.java:96-102` — `LinkedHashSet.addAll` keeps first on `equals` collisions. Two imports with the same label but different subjects both survive the set, and downstream resolution is last-wins by JSON key order. Document precedence and test it, or reject conflicts at validation.
                                                                                                                                                                                                                                  
  **H6. Transitive IDs not validated against intermediate policy's actual imports** *[verified-live]*                                                                                                                             
   
  Reproduced: `PUT ["com.example:totally-fake","com.example:also-fake"]` returns 204. At resolution the IDs are silently filtered out. Either document lenient semantics (forward references) with a test, or reject at write time.           
                                                                                                                                                                                                                                  
  ---             

  ### Medium                                                                                                                                                                                                                      
   
  **M1.** `ResolvedPolicyCacheLoader.java:134` uses `getNow(Optional.empty())` after `allOf`, masking partial-completion bugs. Use `.join()`.                                                                                     
                  
  **M2.** `BackgroundSyncStream.java:284-287` and `ResolvedPolicyCacheLoader.java:161-170` both follow transitives one level — correct today but kept in sync by convention only. Extract a shared `PolicyImports.expectedReferencedPolicyIds(Policy)` helper in the model module and consume from both.
                                                                                                                                                                                                                                  
  **M3.** Inner merge loops in `PolicyImporter.mergeResources` and `mergeSubjects` are O(k²) (copy-on-write). The commit message's "O(n)" claim applies only to the outer merge — trim the claim or fix the inner loops.          
   
  **M4.** `setEntity` on `ModifyPolicyImportTransitiveImports.java:188-194` and `PolicyImportTransitiveImportsModified.java:191-198` does `asArray().asString()` without `wrapJsonRuntimeException`. Bad payloads surface as 500 instead of 400. Wrap the flow or add tests.
                                                                                                                                                                                                                                  
  **M5.** Label collision: B transitively importing both C and D where both export label `X` yields two entries labelled `B:X` with different content. Document the collision semantics in `basic-policy.md`.

thjaeckle:
Thanks for the thorough review — all 14 findings addressed in db77eb4:

### Blocking
- **B1** — Registered `POLICY_IMPORT_TRANSITIVE_IMPORTS` in `PolicyResource` enum + `ResourceMap` + `PolicyPathMatcher`. WS/Connectivity paths now resolve correctly.
- **B2** — Replaced silent `.filter(JsonValue::isString)` with validation that rejects non-string elements via `PolicyImportInvalidException` in both `ImmutableEffectedImports.fromJson` and `PolicyImportsRoute.createTransitiveImportsForPut`.
- **B3** — `invalidateConditionally` now only removes the reverse-mapping when the primary cache entry is actually invalidated. Uses `.get()` (read-only) when the condition doesn't match.

### High
- **H1** — Cycle test now asserts `isNotEmpty()`, `hasSize(1)`, and that the entry label contains `"ROLE"`.
- **H2** — Made `MAX_TRANSITIVE_RESOLUTION_DEPTH` public with comprehensive Javadoc documenting the silent truncation behavior and why logging isn't possible in the model module (no SLF4J dependency, by design).
- **H3** — Documented cross-policy cycle limitation in `PolicyImportsValidator` Javadoc and `basic-policy.md` Key Rules section.
- **H4** — Replaced `whenCompleteAsync` (side-effect after completion) with `thenApplyAsync` (transformation before the value is returned) so import mappings are registered before the cache entry becomes visible to readers.
- **H5** — Documented first-wins merge precedence in `mergeImportedPolicyEntries` Javadoc.
- **H6** — Documented lenient/forward-reference semantics in code comment and `basic-policy.md`.

### Medium
- **M1** — Replaced `getNow(Optional.empty())` with `.join()`.
- **M2** — Extracted `PolicyImports.getExpectedReferencedPolicyIds()` as shared helper; `BackgroundSyncStream` now uses it. `ResolvedPolicyCacheLoader` references it in a comment (it needs only the transitive subset for tag tracking, so direct use would be less clean).
- **M3** — Added complexity clarification comment on inner merge loops.
- **M4** — Wrapped `setEntity` with `wrapJsonRuntimeException` in both `ModifyPolicyImportTransitiveImports` and `PolicyImportTransitiveImportsModified`.
- **M5** — Documented label collision semantics in `basic-policy.md` Key Rules section.

hu-ahmed:
Thanks for the quick fix

hu-ahmed (approved):

thjaeckle:
System tests passed, except one test related to "response diversion" which is known to be flaky:  
https://github.com/eclipse-ditto/ditto/actions/runs/24711913875

PR #2425, opened 2026-04-24
Fix the handling of subscriptions with multiple topics and extra fields

abalarev:
**Summary**
Resolves incorrect handling of subscriptions for multiple topics, containing at least one extra field.

**Issue**
- multiple topics without extraFields works ok - sends to the matching ones
- having multiple topics with extraFields works for only one of them (when it's matching), but not deterministic for which one
- having a topic or topics with extraFields and some without extraFields works again only for one of the topics with extraFields (when it's matching and not for each topic), but for none of the topics without extraFields

**What's changed**
OutboundMaooingProcessorActor flow changed as follows:
- retrieving of extra fields for all topics (per target) combined in a single request
- the filter is now applied using already retrieved extra fields as it needs them to filter correctly. Only those extra fields defined in the matching filter are left in the outbound signal
- splitting of targets with and without extra fields is now removed as it became redundant
- the new flow is still optimized for targets with no extra fields in the topics. 
- outbound signal ordering is now changed, therefore the order of unit tests expectations is also changed
- more unit tests are created to verify the handling of the problem cases described above

Added system tests: https://github.com/eclipse-ditto/ditto-testing/pull/33

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
@@ -70,6 +70,7 @@ import org.eclipse.ditto.connectivity.api.OutboundSignal.Mapped;
 import org.eclipse.ditto.connectivity.api.OutboundSignalFactory;
 import org.eclipse.ditto.connectivity.model.Connection;
 import org.eclipse.ditto.connectivity.model.FilteredTopic;
+import org.eclipse.ditto.connectivity.model.Topic;
 import org.eclipse.ditto.connectivity.model.LogCategory;
 import org.eclipse.ditto.connectivity.model.LogType;
 import org.eclipse.ditto.connectivity.model.MetricDirection;
@@ -103,7 +104,6 @@ import org.eclipse.ditto.json.JsonObject;
 import org.eclipse.ditto.json.JsonPointer;
 import org.eclipse.ditto.json.JsonValue;
 import org.eclipse.ditto.placeholders.ExpressionResolver;
-import org.eclipse.ditto.placeholders.HeadersPlaceholder;
 import org.eclipse.ditto.placeholders.PipelineElement;
 import org.eclipse.ditto.placeholders.PlaceholderFactory;
 import org.eclipse.ditto.placeholders.PlaceholderResolver;
@@ -149,7 +149,6 @@ public final class OutboundMappingProcessorActor
     private static final ResourcePlaceholder RESOURCE_PLACEHOLDER = ResourcePlaceholder.getInstance();
     private static final TimePlaceholder TIME_PLACEHOLDER = TimePlaceholder.getInstance();
     private static final ThingJsonPlaceholder THING_JSON_PLACEHOLDER = ThingJsonPlaceholder.getInstance();
-    private static final HeadersPlaceholder HEADERS_PLACEHOLDER = PlaceholderFactory.newHeadersPlaceholder();
 
     private final ActorRef clientActor;
     private final Connection connection;
@@ -396,7 +395,7 @@ public final class OutboundMappingProcessorActor
                                     outboundPair.first(),
                                     outboundMappingProcessor,
                                     Source.single(outboundPair.first())
-                                            .via(splitByTargetExtraFieldsFlow())
+                                            .via(pairTargetsWithTopicsFlow())
                                             .mapAsync(mappingConfig.getParallelism(), this::enrichAndFilterSignal)
                                             .mapConcat(x -> x)
                                             .map(outbound -> handleOutboundSignal(outbound, outboundMappingProcessor))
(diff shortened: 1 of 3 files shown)


thjaeckle:
@abalarev will you also provide a system-test which tests this fix - seems important enough to me to cover it with an integration-test

abalarev:
@thjaeckle Sounds reasonable, I'll add some system test(s)

thjaeckle:
@abalarev thanks for the added system tests 👍  
I just ran them and seems there are some related test failures for MQTT, could you check them please?  
https://github.com/eclipse-ditto/ditto/actions/runs/25307537926

abalarev:
@thjaeckle Please run the system tests again, I've pushed a fix.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -498,17 +509,35 @@ private CompletionStage<Collection<OutboundSignalWithSender>> enrichAndFilterSig
...
+            final JsonObject extra,
+            final JsonFieldSelector allExtraFields) {
+
+        return topic.getExtraFields()
+                .flatMap(fields -> getExtraFields(expressionResolver, Collections.singletonList(fields)))
+                .map(neededFields -> {
+                    final var builder = extra.toBuilder();
+                    allExtraFields.getPointers().stream()
+                            .filter(pointer -> !neededFields.getPointers().contains(pointer))
+                            .forEach(pointer -> pointer.getRoot().ifPresent(builder::remove));
```
thjaeckle: The trim removes by `pointer.getRoot()`, which deletes the whole top-level object. With `allExtraFields = ["/features/f1", "/features/f2", "/attributes"]` and `neededFields = ["/features/f1", "/attributes"]`, the loop hits `/features/f2`, removes the `features` root, and also wipes `/features/f1` that the matched topic actually requested.

The trimming should operate at full-pointer depth (remove the value at the exact pointer, or rebuild the JSON containing only `neededFields`) rather than at the root segment.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -411,61 +410,54 @@ protected void messageDiscarded(final OutboundSignal message, final QueueOfferRe
...
-                Resolvers.forSignal(outboundSignal.getSource(), connection.getId());
-        final Optional<JsonFieldSelector> extraFieldsOptional = getExtraFields(expressionResolver, filteredTopic);
-        if (extraFieldsOptional.isEmpty()) {
+        final Set<FilteredTopic> topics = outboundSignalWithExtraFields.second();
+
+        List<JsonFieldSelector> allExtraFields = topics.stream()
+                .map(FilteredTopic::getExtraFields)
+                .flatMap(Optional::stream)
+                .toList();
+        boolean topicWithNoFilterNoExtraFieldsExists = topics.stream().anyMatch(topic -> topic.getFilter().isEmpty() && topic.getExtraFields().isEmpty());
```
thjaeckle: `topics` is `target.getTopics()` unfiltered by streaming type — `pairTargetsWithTopics` now passes all of a target's topics whenever any one of them has extra fields. So a target with e.g. a `TWIN_EVENTS` extras-topic plus a `LIVE_MESSAGES` topic with no filter and no extras would short-circuit enrichment for a `TWIN_EVENTS` signal, even though the `LIVE_MESSAGES` topic is irrelevant to that signal.

This `anyMatch` should apply the same `streamingType == StreamingType.fromTopic(topic.getTopic().getPubSubTopic())` guard used in `pairTargetsWithTopics`.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -411,61 +410,54 @@ protected void messageDiscarded(final OutboundSignal message, final QueueOfferRe
...
+        if (allExtraFields.isEmpty() || topicWithNoFilterNoExtraFieldsExists) {
```
thjaeckle: Behavior change worth pinning down: a target that mixes `{ TWIN_EVENTS, no filter, no extras }` with `{ TWIN_EVENTS, filter=X, extras=definition }` will now be delivered **without** extras (the catch-all wins via this early return). Previously the same signal also produced an enriched copy.

Deduplication is good, but is it intentional that the catch-all topic wins over the topic that explicitly requested enrichment? Either way, please document the precedence rule in the javadoc of `enrichAndFilterSignal` so the semantics are unambiguous.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -474,21 +466,40 @@ private CompletionStage<Collection<OutboundSignalWithSender>> enrichAndFilterSig
...
+                                    outboundSignal.getSource(),
+                                    allExtraFieldsOptional.get(),
+                                    extra).orElse(null);
+                            return topics.stream()
+                                    .filter(_ -> enrichedThing != null || topicWithNoFilterExists)
+                                    .flatMap(topic -> applyFilter(outboundSignal, enrichedThing, topic)
+                                            .map(signal -> setTrimmedExtra(signal, topic, expressionResolver,
+                                                    extra, allExtraFieldsOptional.get()))
+                                            .stream())
+                                    .findFirst()
```
thjaeckle, line 502: `topics` is a `Set<FilteredTopic>`, which has no defined iteration order. If a signal matches more than one topic on the target, `findFirst()` picks an arbitrary one, so the trimmed extras delivered downstream become non-deterministic — the very symptom this PR sets out to fix.

Consider iterating in a deterministic order: derive from `target.getTopics()` via a structure that preserves insertion order, or sort topics by a stable key, so the chosen match is reproducible.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -411,61 +410,54 @@ protected void messageDiscarded(final OutboundSignal message, final QueueOfferRe
...
+                            .map(targetAndSelector -> Pair.create(
+                                    outboundSignal.setTargets(Collections.singletonList(targetAndSelector.first())),
+                                    targetAndSelector.second()))
+                            .toList();
                 });
     }
 
-
     // Called inside stream; must be thread-safe
     // precondition: whenever filteredTopic != null, it contains an extra fields
```
thjaeckle: This precondition comment is now stale — the parameter is `Set<FilteredTopic>`, not a single nullable `FilteredTopic`. Please update or remove.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -411,61 +410,54 @@ protected void messageDiscarded(final OutboundSignal message, final QueueOfferRe
      * Create a flow that splits 1 outbound signal into many as follows.
      * <ol>
      * <li>
-     * Targets with matching filtered topics without extra fields are grouped into 1 outbound signal, followed by
+     *     For each target without extra fields, it produces a pair of outbound signal and empty set of topics.
+     *     As these targets have already passed pre-filtering in an early stage, no more filtering is needed.</li>
+     * <li>
+     *     For each target containing any extra field in its topics, it produces a pair of outbound signal and a set of its target topics.
+     *     As the filter could incLude extra fields, an additional filtering must be performed after extracting the extra fields.
```
thjaeckle: Typo: `incLude` → `include`.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
```diff
@@ -411,61 +410,54 @@ protected void messageDiscarded(final OutboundSignal message, final QueueOfferRe
...
+            final Pair<OutboundSignalWithSender, Set<FilteredTopic>> outboundSignalWithExtraFields) {
         final OutboundSignalWithSender outboundSignal = outboundSignalWithExtraFields.first();
-        final FilteredTopic filteredTopic = outboundSignalWithExtraFields.second();
-        final ExpressionResolver expressionResolver =
-                Resolvers.forSignal(outboundSignal.getSource(), connection.getId());
-        final Optional<JsonFieldSelector> extraFieldsOptional = getExtraFields(expressionResolver, filteredTopic);
-        if (extraFieldsOptional.isEmpty()) {
+        final Set<FilteredTopic> topics = outboundSignalWithExtraFields.second();
+
+        List<JsonFieldSelector> allExtraFields = topics.stream()
```
thjaeckle: The rest of this file is consistently `final` on every local. Please add `final` to the new locals introduced in this method: `allExtraFields` (here), `topicWithNoFilterNoExtraFieldsExists`, `topicWithNoFilterExists`, `allExtraFieldsOptional`, and `partialThingOptional`.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActorTest.java
```diff
@@ -274,19 +256,226 @@ public void expectNoTargetIssuedAckRequestInPublishedSignals() {
...
+                                .isFalse();
+                    });
+        }};
+    }
+
+    @Test
+    public void multipleTopicsWithoutExtraFieldsFastProcessed() {
+        new TestKit(actorSystemResource.getActorSystem()) {{
+            // Create a target with multiple topics all without extra fields.
+            // Must send outbound signal, skipping filtering as the pe-filtering has already done the job
```
thjaeckle: Typo: `pe-filtering` → `pre-filtering`.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActorTest.java
```diff
@@ -274,19 +256,226 @@ public void expectNoTargetIssuedAckRequestInPublishedSignals() {
...
+                        protocolAdapterProvider.getProtocolAdapter("test"),
+                        AbstractMessageMappingProcessorActorTest.mockLoggingAdapter(), null)
         );
     }
 
+    @Test
+    public void multipleTopicsWithExtraFieldsFirstTopic() {
+        new TestKit(actorSystemResource.getActorSystem()) {{
+            // Create a target with multiple FilteredTopics for the same streaming type with different extraFields
+            List <Target> targets = List.of(createTestTargetMultiTopics(Set.of(topic4(), topic3(), topic5())));
```
thjaeckle: Stray space between `List` and `<Target>` (also at lines 310, 341, 377, 402). Should be `List<Target>` to match the project style.

thjaeckle (commented):
Thanks for tackling this — the refactor direction looks right and the headline scenarios are well-covered. A few correctness concerns and polish items inline.

One overarching note on test coverage: **the four correctness concerns flagged below (shared-root trimming, cross-streaming-type early return, non-deterministic `findFirst` over a `Set`, and catch-all / extras precedence) are not yet exercised by the new tests.** Once the intended behavior for each is settled, please add regression tests that lock it in.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Removed the comment line.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Fixed.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Fixed.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActorTest.java
abalarev: Fixed.

connectivity/service/src/test/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActorTest.java
abalarev: Fixed.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Previously the signal produced enriched copy only in case of matching filter, but no signal at all if not. I agree now the signal matching the filter would behave differently too. 
Maybe it's worth to give priority to topics with extra as this would give more flexibility - it seems I have to just change the 'if'-statement to `if (allExtraFields.isEmpty()) {` in order to achieve this.
Or just documenting this change as there is no test covering the previous behaviour and therefore I would assume nobody use this in that way and relies on it. 

@thjaeckle  What is your opinion on this? Mine leans towards changing the 'if'-statement

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Agree, I've added sorting by extraFields, such as no-extra-fields topics have less priority.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: I had a feeling I've messed up something with the streamingType as I had few failing tests regarding it. I though I've fixed this in `applyFilter()` method, but I now realize it is not correct. 
I think now I must filter the topics returned by pairTargetsWithTopics to exclude topics with not the same streaming type, as you said LIVE_MESSAGES to be excluded. So my fix is to add a filter in pairTargetsWithTopics and to keep here the criterion as it is.

This is already tested by `AbstractConnectivityITestCases#publishEnrichedSignals` test method, which is failing, once the following incorrect 'fix' in applyFilter() method is removed:
```
        if (!topicMatchesTopicPath(topicPath, topic.getTopic())) {
            return Optional.empty();
        }
```

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
thjaeckle: @abalarev yes, I would also do the `if (allExtraFields.isEmpty()){`. 
The other one would be an unexpected behavior change - the fact that no unit test covers this does not mean that no adopters rely on this :)

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Trimming is now done by getting the neededFields from the JSON. 
Special handling of wildcard is implemented to work with '/features/*/xxx' extra fields.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
thjaeckle: @abalarev this is not yet addressed - did you intentionally leave it that way?

abalarev:
@thjaeckle Thank you for the review findings and comments! 
I've fixed the issues, added and updated the unit tests to cover the changes.
Please review again!

thjaeckle:
@abalarev one thread is still unresolved - the other changes look good 👍

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/OutboundMappingProcessorActor.java
abalarev: Aaah, sorry. I did it last week, but lost it somewhere on the way.
Which mean it's not covered by the tests yet.

abalarev: @thjaeckle Now it is fixed, a new unit test covers it.

thjaeckle (approved):
Thanks @abalarev — all the items from the previous round are addressed and the new tests do a good job of locking in the intended behavior.  
The refactor reads cleaner than the previous shape. 

LGTM.

PR #2429, opened 2026-05-04
#2354 Add policy-view=resolved for effective policy view

### LINKED ISSUE #2354: Add API support to return effective (resolved) policy after imports resolution
Today, Policy APIs return configured policy state (local entries + imports config), but not the final resolved policy view after imports are applied.

This issue proposes a follow-up API capability to retrieve the effective/resolved policy.

**Current behavior**

`GET /api/2/policies/{policyId} `returns stored policy config.
`GET /api/2/policies/{policyId}/entries` returns local entries only.
Imports config is visible, but effective merged entries are not directly visible.
**Problem**

- Difficult to debug and validate authorization behavior.
- Hard to explain effective access when imports and additions are involved.
- Operational/support troubleshooting requires internal knowledge.

**Proposal**
Add a read API/view that returns the effective resolved policy after:

- import resolution (implicit / explicit / never)
- entriesAdditions merge logic
- allowed additions constraints

**Possible shape (one option):** 

`GET /api/2/policies/{policyId}?view=resolved
`
**Alternative:**

`GET /api/2/policies/{policyId}/resolved
`
_Expected output (high level)_

- Fully resolved entries used for enforcement.
- Optional metadata per entry (e.g. local vs imported origin: policyId + entry label).
- Clear indication of what was ignored due to import rules (if feasible).

hu-ahmed:
Closes #2354. Adds a `policy-view` request hint on `GET /api/2/policies/{id}` that returns the merged "effective" policy.                                                              
   
  - `policy-view=original` (default) — policy as stored, unchanged behaviour.                                                                                                            
  - `policy-view=resolved` — own entries + declared imports + configured namespace-root policies, with `references` resolved.
                                                                                                                                                                                         
  Imported entries appear under `imported-<importedPolicyId>-<originalLabel>` (treat as opaque). The hint can also be sent as the `policy-view` Ditto header, so the resolved view is available uniformly via HTTP, WebSocket, SSE and Connectivity.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/headers/DittoHeaderDefinition.java
@@ -679,7 +679,32 @@ public enum DittoHeaderDefinition implements HeaderDefinition {
             JsonObject.class,
             true,
             false,
-            HeaderValueValidators.getJsonObjectValidator());
+            HeaderValueValidators.getJsonObjectValidator()),
+
+    /**
+     * Selects the policy representation returned by retrieval-style commands: {@code "original"}
+     * (default, policy as stored) or {@code "resolved"} (merged view including imports, references
+     * and namespace-root policies). Parsing/validation lives in the policies module.
+     * <p>
+     * Key: {@code "policy-view"}, Java type: {@link String}.
+     * </p>
+     *
+     * @since 3.9.0
+     */
+    POLICY_VIEW("policy-view", String.class, true, false, HeaderValueValidators.getNoOpValidator()),
+
+    /**
+     * Internal-only header threading the original {@code fields=} selector from the gateway route to
+     * {@code PolicyCommandEnforcement} so it can be re-applied to the merged JSON when
+     * {@code policy-view=resolved}.
+     * <p>
+     * Key: {@code "ditto-policy-view-fields-selector"}, Java type: {@link String}.
+     * </p>
+     *
+     * @since 3.9.0
+     */
+    POLICY_VIEW_FIELDS_SELECTOR("ditto-policy-view-fields-selector", String.class, false, false,
+            HeaderValueValidators.getNoOpValidator());
 
     /**
      * Map to speed up lookup of header definition by key.
--- a/base/model/src/test/java/org/eclipse/ditto/base/model/headers/ImmutableDittoHeadersTest.java
+++ b/base/model/src/test/java/org/eclipse/ditto/base/model/headers/ImmutableDittoHeadersTest.java
@@ -241,6 +241,8 @@ public final class ImmutableDittoHeadersTest {
                 .putHeader(DittoHeaderDefinition.DIVERT_RESPONSE_TO_CONNECTION.getKey(), KNOWN_DITTO_DIVERT_RESPONSE_TO)
                 .putHeader(DittoHeaderDefinition.DIVERTED_RESPONSE_FROM_CONNECTION.getKey(), KNOWN_DITTO_DIVERTED_RESPONSE_FROM)
                 .putHeader(DittoHeaderDefinition.DIVERT_EXPECTED_RESPONSE_TYPES.getKey(), KNOWN_DITTO_DIVERT_RESPONSE_TYPES)
+                .putHeader(DittoHeaderDefinition.POLICY_VIEW.getKey(), "resolved")
+                .putHeader(DittoHeaderDefinition.POLICY_VIEW_FIELDS_SELECTOR.getKey(), "policyId,entries")
                 .build();
 
         assertThat(underTest).isEqualTo(expectedHeaderMap);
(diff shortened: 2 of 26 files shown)


thjaeckle:
Thanks for the thorough work on this — the merge pipeline is well structured and the documentation page is clear. I did a security-focused review and ran a live reproduction; flagging one disclosure issue I think needs to be addressed before merge.

**Issue: `policy-view=resolved` exposes contents of imported / namespace-root policies to callers who have no READ on the source policy.**

A caller with broad `policy:/` READ on the importing policy can read full entries (subjects, grants, original labels) of any policy whose entries are merged in — both via declared `imports` and via namespace-root config — even when the caller has zero permission on those source policies.

**Why it leaks.** `PolicyCommandEnforcement.filterResponse` builds the merged JSON in `resolveAndRebuildResponse`, then runs `applyEnforcerJsonView` using the importing policy's enforcer. That enforcer evaluates READ on paths in the *importing* policy's resource tree only. There is no check that the caller has any permission on the source policy. A subject with `policy:/: GRANT READ` on the importing policy therefore matches every merged path. For namespace-root specifically, the import-time READ pre-enforcer is also bypassed (intentionally — it's operator-injected), so nothing along the entire chain validates the reader.

**Reproduction (live, against this PR's branch).**

Setup:
- `root:policy` — only `nginx:admin` has any access. Entry `AUDIT_TEAM` (`importable=implicit`) holds confidential subject `nginx:secret-auditor` granting `thing:/: READ`. `AUDIT_TEAM` does **not** grant any `policy:/` permissions.
- `importing:policy-1` (namespace `importing`) — entry `dave-full-read` grants `nginx:dave` `policy:/: GRANT READ`.
- `policies-dev.conf`: `ditto.namespace-policies { \"importing\" = [\"root:policy\"] }`. Restart.

Result:
- `GET /api/2/policies/root:policy` as dave → `404` (correct, dave has no permission on root).
- `GET /api/2/policies/importing:policy-1?policy-view=resolved` as dave → response includes the full `AUDIT_TEAM` entry, **including subject `nginx:secret-auditor` and its grants**.

Same outcome via declared imports: an `imports: { \"root:policy\": { \"entries\": [] } }` declaration on a separate importing policy yields `imported-root:policy-AUDIT_TEAM` with the same confidential contents to a similarly-scoped reader (`nginx:eve`). The `imported-<sourceId>-<label>` prefix additionally reveals the source policy ID and original label.

**Why this matters operationally.** Namespace-root is the worst case: operators bind a root policy in HOCON without any policy-side declaration. They reasonably expect its contents (audit/SRE/break-glass subjects, monitoring users, etc.) to remain confidential to tenants. After this PR, any tenant principal with broad `policy:/` READ on any policy in the bound namespace can dump those contents in cleartext via a single GET. Declared imports are less surprising (the importing policy author chose them) but the leak vector is the same.

**Suggested fix.** In `resolveAndRebuildResponse`, before including a merged entry in the rebuilt response, verify the caller has READ on the **source** policy entry's path. Concretely:

- For each entry contributed by import / namespace-root, look up the source policy via `policyResolver`, build (or reuse) the source's `PolicyEnforcer`, and check `sourceEnforcer.hasPartialPermissions(policy:/entries/<originalLabel>, ctx, READ)`.
- Drop the entry if the source-side check fails. The downstream `applyEnforcerJsonView` would then layer field-level filtering on top.

Lighter alternative if the cross-policy check is too heavy for now: skip namespace-root contributions in the resolved view altogether (declared imports could still surface, with the existing filter behavior — narrower blast radius since the importing-policy author had to opt in). Either way, this should be regression-protected with two cases in `PolicyCommandEnforcementTest`:

1. caller with broad `policy:/` READ on importing policy + zero permission on declared-import source → `imported-*` entry contents must be absent.
2. caller with broad `policy:/` READ on importing policy + zero permission on namespace-root source → namespace-root entries must be absent.

hu-ahmed:
system test passed 
https://github.com/eclipse-ditto/ditto/actions/runs/25489338187

thjaeckle (approved):
Re-reviewed after `8a21274f29`. The disclosure leak is closed:

- `PolicyCommandEnforcement.mergeAndDropUnreadable` runs source-side READ checks against each declared import and namespace-root before `applyEnforcerJsonView` filters per-caller — exactly the layered approach suggested.
- Both regression tests requested in the previous review are present in `PolicyCommandEnforcementResolvedViewTest`.
- Live-tested on a local cluster: declared imports (IMPLICIT and EXPLICIT importable), per-entry granular source-side READ correctly allows entries through, transitive imports produce the double-prefixed `imported-<B>-imported-<C>-<label>` form and are dropped without source-side READ, conditional headers and field selectors behave as designed.

Two minor follow-ups worth considering (none blocking):

1. The source enforcer is resolved twice per resolved-view request — once inside `withResolvedImportsAndNamespacePolicies` and again via `mergeAndDropUnreadable.resolveSourceEnforcer`. Negligible with a cache-backed resolver but could be deduplicated.
2. Comment at line 282 ("labels are unique per source via the imported-`<sourceId>`- prefix") could mention the double-prefix form for transitive imports.

LGTM.

PR #2433, opened 2026-05-07
#2431 prefix namespace-policy entries with nsimported- to prevent shadowing

### LINKED ISSUE #2431: namespace-policies entries can be silently shadowed by local labels — operator-intended access can be bypassed by tenants
## Summary

Namespace-policies merge their `importable=implicit` entries into the policy under their **original label** (`PolicyEnforcer.mergeImplicitEntries`). On any label collision the entry that was set *first* keeps the slot and every later contribution is silently dropped — with no operator-visible signal. This produces two related problems:

1. **Tenant shadow attack.** Local entries are merged before namespace-roots, so any tenant in a bound namespace can fully bypass a namespace-policy entry by name-clashing.
2. **Multi-root non-composition.** When several namespace-policies match the same namespace and contribute the same label, only the highest-precedence one survives. Operators cannot stack roots that share labels.

**Declared imports already solve both problems today.** They route every imported entry through `PolicyImporter.rewriteLabel`, which produces a unique label of the form `imported-<importedPolicyId>-<originalLabel>`, and the label validator rejects any local label that starts with `imported-` (HTTP 400 `policies:label.invalid`). As a result:

- Tenants cannot create a local entry under any imported label (validator rejects it).
- Two declared imports of the same `originalLabel` from different source policies become two distinct merged labels (because each carries its own source policy ID), so both compose additively.

Namespace-policies bypass that machinery and inherit neither property. The proposed fix is to bring them into line with the proven declared-imports approach.

## Impact

Operators configure `namespace-policies` for cross-tenant concerns where the implicit assumption is "this entry applies to every policy in the namespace, period":
- audit / governance READ access
- compliance / security monitoring
- break-glass / SRE accounts
- forced minimum logging or message subjects

### 1. Tenant shadow attack

A tenant in the bound namespace can:
1. Pick a label they suspect is in use (operator-configured labels are not secret — common names like `audit`, `compliance`, `monitoring`, `sre-break-glass` are easy to guess and can be enumerated by trial).
2. Add a local entry with that label (e.g. their own subject only) — silently nullifying the operator-configured access.
3. Do this at policy creation time or any time afterwards via `PUT` — no operator-visible signal is generated.

The effect is not just on the policy itself: any Thing whose `policyId` references a tenant-shadowed policy also loses the operator-configured access (e.g. an auditor cannot read the Thing).

### 2. Multi-root non-composition

When more than one namespace-root applies to a namespace (either multiple entries in one binding's array, or several pattern keys matching — `exact > more specific prefix wildcard > less specific prefix wildcard > "*"`), the merge iterates them in precedence order and skips any label that is already set. Concretely:

- **Within a single pattern's array:** array order = precedence order.
- **Across matching patterns:** exact-match pattern first, then increasingly broad prefix wildcards, then `*`.
- For each label, only the **first** root to contribute it ends up in the merged policy. Every subsequent root's same-label entry is silently ignored.

Operators commonly want the opposite: e.g. a global `audit-team` that applies cross-tenant *and* a tenant-specific `audit-team` that augments it. With the current design, the more-specific tenant root completely replaces the global one — the global audit-team subjects/grants disappear without warning.

## Reproduction

Test environment: local Ditto built from `master`; pre-authentication enabled.

### Reproduction A — tenant shadow attack

Config:
```hocon
ditto.namespace-policies {
  "tenant.acme" = ["acme:operator-root"]
}
```

`acme:operator-root`:
```json
{
  "entries": {
    "ROOT_OWNER": { "subjects": { "nginx:admin": {} }, "resources": {"policy:/": {"grant":["READ","WRITE"],"revoke":[]}}, "importable": "never" },
    "audit-team": {
      "subjects": { "nginx:auditor": {} },
      "resources": {
        "policy:/": { "grant": ["READ"], "revoke": [] },
        "thing:/":  { "grant": ["READ"], "revoke": [] }
      },
      "importable": "implicit"
    }
  }
}
```

Step 1 — tenant creates `tenant.acme:my-policy` *without* an `audit-team` entry. `nginx:auditor` correctly gets READ on the policy and on Things attached to it (HTTP 200).

Step 2 — tenant `PUT`s the policy with a local shadowing entry:
```json
{
  "entries": {
    "owner": { "subjects": { "nginx:admin": {} }, "resources": {"policy:/":{"grant":["READ","WRITE"],"revoke":[]},"thing:/":{"grant":["READ","WRITE"],"revoke":[]},"message:/":{"grant":["READ","WRITE"],"revoke":[]}}, "importable": "never" },
    "tenant-self": { "subjects": { "nginx:tenant-bob": {} }, "resources": {"policy:/":{"grant":["READ","WRITE"],"revoke":[]},"thing:/":{"grant":["READ","WRITE"],"revoke":[]}}, "importable": "never" },
    "audit-team": {
      "subjects": { "nginx:tenant-bob": {} },
      "resources": { "thing:/": { "grant": [], "revoke": ["READ"] } },
      "importable": "never"
    }
  }
}
```
→ `HTTP 204` (accepted, no warning).

Step 3 — the operator's audit access is gone:
- `GET /api/2/policies/tenant.acme:my-policy` as `nginx:auditor` → `HTTP 404`.
- `GET /api/2/things/tenant.acme:thing-1` as `nginx:auditor` → `HTTP 404`.

The same shadowing works at policy creation time — a brand-new policy can be created pre-shadowed so the namespace-policy never applies. Multiple namespace-root labels can be shadowed in the same `PUT`.

For comparison, the equivalent against a declared-import label fails fast:
```
PUT /api/2/policies/tenant.acme:try-import-prefix
{ "entries": { "imported-acme:operator-root-audit-team": { ... } } }

→ HTTP 400 policies:label.invalid
  "Label 'imported-acme:operator-root-audit-team' is not valid! It must not start with the blocklisted 'imported' prefix."
```

So the protection mechanism already exists in the codebase — it just isn't applied to namespace-policy entries.

### Reproduction B — multi-root non-composition

Config:
```hocon
ditto.namespace-policies {
  "tenant.acme" = ["acme:operator-root", "acme:operator-second-root"]
}
```

Both roots define `audit-team` with different subjects:

- `acme:operator-root` — `audit-team` (`importable=implicit`) granting `nginx:auditor` READ on `policy:/` and `thing:/`.
- `acme:operator-second-root` — `audit-team` (`importable=implicit`) granting `nginx:auditor-from-second-root` READ on `thing:/`. Also a non-colliding `second-root-only` entry.

`tenant.acme:my-policy` is created with no `audit-team` of its own.

Result:

| Subject                                                       | `GET /policies/tenant.acme:my-policy` |
|---------------------------------------------------------------|----------------------------------------|
| `nginx:auditor` (subject of *first* root's `audit-team`)      | **HTTP 200** ✓ merged in               |
| `nginx:auditor-from-second-root` (subject of *second* root's `audit-team`) | **HTTP 404** ✗ silently dropped       |
| `nginx:second-root-subject` (label only in second root, no collision) | merged in (verified by admin via the bulk merge)  |

The second root contributed its non-colliding `second-root-only` entry, but its `audit-team` was silently discarded. Operators stacking roots with overlapping label names cannot rely on them composing.

### Reproduction C — declared imports already compose correctly (control case)

`acme:source-A` and `acme:source-B`, both with an `importable=implicit` entry named `shared-label` but with different subjects. `acme:importing-both` declares imports of both. Resolved view of the importing policy:

```
owner                                       -> ['nginx:admin']
imported-acme:source-A-shared-label         -> ['nginx:userA-from-source-A']
imported-acme:source-B-shared-label         -> ['nginx:userB-from-source-B']
```

Both labels survive — `PolicyImporter.rewriteLabel` embedded the source policy ID, so the label collision never happened. Both `nginx:userA-from-source-A` and `nginx:userB-from-source-B` get HTTP 200 reading a Thing attached to the importing policy. This is the behavior we want for namespace-policies too.

## Suggested fix

**Apply the same approach that already works for declared imports** — route namespace-root entries through `PolicyImporter.rewriteLabel` with a distinct reserved prefix `nsimported`, so the merged label becomes `nsimported-<sourcePolicyId>-<originalLabel>`. Concretely:

1. In `PolicyEnforcer.mergeImplicitEntries`, replace the bare `result.setEntry(entry)` call with a `PolicyImporter.rewriteLabel`-based rewrite that uses the namespace-root's source policy ID to produce a unique label per source.
2. Add `nsimported` to the label validator's blocklist (the same one that already rejects `imported-`). Local labels starting with `nsimported-` are rejected at PUT time with `HTTP 400 policies:label.invalid`.
3. Adjust the resolved-view documentation to note the new label format, analogous to the existing note about `imported-<sourceId>-<originalLabel>`.

This single change addresses **both** problems with the proven mechanism:

- **Shadow attack:** local labels with the `nsimported-` prefix are rejected by the validator. Tenants cannot create an entry that collides with any namespace-root entry.
- **Multi-root composition:** because each namespace-root's contributions carry its source policy ID in the label, two roots with the same `originalLabel` produce two distinct merged labels (e.g. `nsimported-acme:operator-root-audit-team` and `nsimported-acme:operator-second-root-audit-team`). Both survive in the merged enforcer; both source's subjects get their permissions. Composition becomes additive by default.

The mechanism is already implemented, tested in production via declared imports, and verified live (Reproduction C above) — the fix is to extend it to the namespace-policies code path rather than invent a new scheme.

### Optional: audit signal on collision

As a complementary observability improvement (not a fix), emit a warning/event when:
- a tenant mutation creates or modifies a label that collides with an active namespace-policy label, or
- a namespace-policy load drops a label because a higher-precedence root already provided it.

Useful as a transitional safety net while existing deployments migrate to the rewritten label format.

thjaeckle:
Closes #2431.

## Summary

* Namespace-policies are now merged under rewritten labels of the form `nsimported-<sourcePolicyId>-<originalLabel>` (mirroring how declared imports already use `imported-<sourceId>-<originalLabel>`), and local labels starting with `nsimported-` are rejected by the validator with `HTTP 400 policies:label.invalid`.
* This closes both problems #2431 documented:
  - tenants can no longer silently nullify an operator-configured namespace-root entry by creating a local entry with the same label;
  - multiple namespace-roots that contribute entries with the same `originalLabel` now compose additively instead of dropping all but the highest-precedence one.
* The duplicated `mergeImplicitEntries` in `PolicyEnforcer` and `ResolvedPolicyCacheLoader` is consolidated into `PolicyImporter.mergeImplicitNamespaceRootEntries(rootPolicy, rootPolicyId, currentPolicy)`.
* Merged policies live in-memory only at enforcer-build time, so no migration is required.

## Test plan

- [x] `mvn test -pl policies/model` (1029 tests pass)
- [x] `mvn test -pl policies/enforcement` (261 tests pass)
- [x] `mvn test -pl thingsearch/service` (210 tests pass)
- [x] `mvn license:check` for affected modules
- [x] End-to-end verification on a local Ditto instance per the issue's reproduction:
  - admin's resolved view of a tenant policy in a namespace bound to two namespace-roots (`acme:operator-root` and `acme:operator-second-root`, both with an `audit-team` entry) shows both contributions under distinct `nsimported-<rootId>-audit-team` labels;
  - `nginx:auditor` and `nginx:auditor-from-second-root` retain their cross-tenant access via the merged enforcer;
  - a tenant adding a local `audit-team` entry no longer nullifies the operator's audit-team — both entries coexist and the operator-configured READ on the policy/Thing keeps working;
  - `PUT /policies/...` with a local `nsimported-foo` label is rejected `HTTP 400 policies:label.invalid` with the description mentioning both blocklisted prefixes.

## Tests added or updated

New regression tests for the issue's scenarios:
- `PolicyEnforcerNamespacePoliciesTest.localLabelDoesNotShadowNamespaceRootLabelOnConflict`
- `PolicyEnforcerNamespacePoliciesTest.multipleNamespaceRootsWithOverlappingLabelsBothCompose`
- `ResolvedPolicyCacheLoaderTest.localLabelDoesNotShadowNamespaceRootLabelInSearchIndexResolution`
- `ResolvedPolicyCacheLoaderTest.multipleNamespaceRootsWithOverlappingLabelsBothComposeInSearchIndexResolution`
- `ImmutableLabelTest.createNsimportedBlocklistedLabel`
- `ImmutableImportedLabelTest.nsImportedLabelFormatIncludesPolicyId` plus a few siblings

Existing tests updated where they encoded the buggy behavior (root entry dropped on collision) to match the new label form.

## Coordination note for #2429

PR #2429 (`policy-view=resolved`) hardcodes a check for the `imported-` prefix in `RetrievePolicyResponse.hasImportedPrefixedLabel` to choose between the strict and the lenient parser when round-tripping the resolved view. After this PR lands, the resolved view will also contain `nsimported-` labels — the prefix detector should be extended to also match `nsimported-` so the lenient parser kicks in for those.


hu-ahmed (commented):
Thanks for the fix, looks solid 👍 
just a small point we have 
**Stale Javadoc** on `PolicyEnforcer.withResolvedImportsAndNamespacePolicies` and the private `mergeNamespacePolicies`:                                                                
   
  > *"Local/imported entries always win: if a label already exists in resolvedPolicy, the corresponding namespace root entry is skipped."*                                               
                                                            
  That's no longer how the merge works — by construction, local and namespace-root labels can't collide. Both blocks should be updated to describe the prefix-based rewrite.

thjaeckle:
@hu-ahmed thanks for the hint - I updated it 👍

hu-ahmed (approved):

PR #2434, opened 2026-05-08
#2405 Add UI support for Ditto 3.9.0 policy and search features

### LINKED ISSUE #2405: Enhance Ditto UI by added Ditto 3.9.0 features
Ditto 3.9.0 got some features which are not yet reflected in the UI.

Enhance the Ditto UI to support those

## Search additions

* Support for "empty()" RQL predicate

## Policy additions

We added quite some policy model fields in 3.9.0:
* imports
  * entriesAdditions *new*
* importsAliases *new*
* entries
  * namespaces for policy entries *new*
  * allowedImportAdditions *new*

The Ditto UI "Policies" tab however does not yet visualize those and does not provide means to CRUD them.

thjaeckle:
Resolves: #2405

Policies tab:
- Imports: edit `transitiveImports` whitelist per import (Add input, ↗ navigate, × remove with confirm). Field omitted from PUT body when list is empty (absent ≡ []).
- Entries: new `namespaces` editor (side-by-side with `references`), new `references` editor with two-dropdown picker (local entries vs imported-policy entries, lazy-fetched and cached, free-text fallback when imported policy unreachable), inline `allowedAdditions` tri-state control (Default / Deny all / Allow only:) preserving the absent-vs-`[]` distinction.
- Layout restructure: Subjects + Resources side-by-side; Namespaces + References side-by-side underneath; "For the entry selected above:" hint to signal hierarchy.
- Resource JSON editor: arrays of primitives no longer wrap across lines (compact `["READ","WRITE"]`); editor capped to remaining column space.
- Subject crud-toolbar: tooltip on the label and `{"type":""}` JSON skeleton seeded with cursor inside the empty value when creating.
- New 3.9 features showcase template covering namespaces / allowedAdditions / references / transitiveImports.
- `<crud-toolbar>` gains optional `tooltip` attribute applied to its internal label.

Search:
- `empty()` RQL predicate: added to autocomplete examples, RQL filter tooltip, and the predicate-detection regex in `searchTriggered` (the latter previously routed `empty(...)` queries to a thingId lookup).

Misc:
- Edit-overlay z-indices (`editBackground` / `editForground`) lowered below Bootstrap modal layer so confirmation dialogs opened during editing render on top of the overlay.
- `Utils.confirm` renders messages via DOMPurify-sanitised innerHTML so the established `<br>` line-break convention works (was rendering literal `<br>`).

### DIFF
--- a/ui/main.scss
+++ b/ui/main.scss
@@ -231,10 +231,13 @@ hr {
   margin: 0.25rem;
 }
 
+// Edit-mode overlay z-indices sit BELOW Bootstrap's modal layer (modal-backdrop: 1050, modal: 1055)
+// so that confirmation/info modals opened from inside an editing crud-toolbar render on top of the
+// edit overlay rather than being faded out behind it.
 .editBackground {
   display: block;
   position: fixed;
-  z-index: 1100;
+  z-index: 1040;
   left: 0;
   top: 0;
   width: 100%;
@@ -245,13 +248,13 @@ hr {
 }
 
 .editForground {
-  z-index: 1200;
+  z-index: 1045;
   position: relative;
   background-color: white;
 }
 
 .editForgroundBig {
-  z-index: 1200;
+  z-index: 1045;
   position: fixed;
   background-color: white;
   top: 0;
--- a/ui/main.ts
+++ b/ui/main.ts
@@ -29,6 +29,8 @@ import * as PoliciesEffective from './modules/policies/policiesEffective.js';
 import * as PoliciesEntries from './modules/policies/policiesEntries.js';
 import * as PoliciesImports from './modules/policies/policiesImports.js';
 import * as PoliciesJSON from './modules/policies/policiesJSON.js';
+import * as PoliciesNamespaces from './modules/policies/policiesNamespaces.js';
+import * as PoliciesReferences from './modules/policies/policiesReferences.js';
 import * as PoliciesResources from './modules/policies/policiesResources';
 import * as PoliciesSubjects from './modules/policies/policiesSubjects';
 import * as Attributes from './modules/things/attributes.js';
@@ -74,6 +76,8 @@ document.addEventListener('DOMContentLoaded', async function() {
   PoliciesEntries.ready();
   PoliciesSubjects.ready();
   PoliciesResources.ready();
+  PoliciesNamespaces.ready();
+  PoliciesReferences.ready();
   PoliciesEffective.ready();
   Connections.ready();
   ConnectionsCRUD.ready();
--- a/ui/modules/policies/policies.html
+++ b/ui/modules/policies/policies.html
@@ -46,11 +46,11 @@
                         </a>
                     </li>
                 </ul>
-                <div class="tab-content" style="flex-grow: 1; overflow: hidden;">
+                <div class="tab-content" style="flex-grow: 1; overflow-y: auto; height: 80vh;">
                     <div class="tab-pane container active no-margin" id="tabPolicyEntries">
                         <h5>Policy Entries</h5>
                         <hr />
-                        <div class="split-h resizable_pane" style="height: 22vh;" data-split="horizontal" data-split-sizes="[50,50]" data-split-key="policies-entries">
+                        <div class="split-h resizable_pane" style="height: 16vh;" data-split="horizontal" data-split-sizes="[50,50]" data-split-key="policies-entries">
                             <div class="split-panel resizable_flex_column">
                                 <div class="input-group has-validation">
                                     <input class="form-control" id="tableValidationEntries" hidden="true"></input>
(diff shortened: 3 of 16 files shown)


hu-ahmed:
## Review of `55a34b2` — UI 3.9.0 features                                                                                                                                             
                                           
  Verified end-to-end against a local stack (round-trip of all four new fields, `tsc -noEmit` clean, manual UI walk-through). Two items I'd recommend addressing before merge.           
   
  ---                                                                                                                                                                                    
                                           
  ### 🟡 Medium

  #### M1 — Race condition in `loadEntriesForImport`                                                                                                                                     
   
  `ui/modules/policies/policiesReferences.ts:103-138`                                                                                                                                    
                                           
  `loadEntriesForImport` is async with no in-flight token. If a user clicks reference row A (`import=policyA`), then quickly clicks row B (`import=policyB`), and `policyA`'s fetch  resolves *after* `policyB`'s, `policyA`'s labels overwrite the dropdown the user is now looking at. Both `onPolicyReferencesClick` (L202) and `onImportSelectChange` (L91) fire-and-forget.                                                                                                                                                                       
                                           
  Suggested fix — monotonic token on resolve:                                                                                                                                            
   
  ```ts                                                                                                                                                                                  
  let loadToken = 0;                       
                                                                                                                                                                                         
  async function loadEntriesForImport(importValue: string, preselectEntry?: string) {
    const myToken = ++loadToken;                                                                                                                                                         
    // ...                                                                                                                                                                               
    const importedPolicy = await getImportedPolicy(importValue);                                                                                                                         
    if (myToken !== loadToken) return; // stale, a newer call superseded us                                                                                                              
    // ...                                                                                                                                                                               
  }                                                                                                                                                                                      
                                                                                                                                                                                         
  Repro: throttle Network to "Slow 3G", click rapidly between a local reference and an imported reference. The Entry dropdown momentarily shows wrong labels.                            
   
  ---                                                                                                                                                                                    
  M2 — In-place mutation of Policies.thePolicy.entries[selectedEntry]
                                                                     
  ui/modules/policies/policiesEntries.ts:164-170
                                                                                                                                                                                         
  function onUpdateEntryClick() {
    validateAllowedAdditionsSelection();                                                                                                                                                 
    const entry = Policies.thePolicy.entries[selectedEntry];   // reference, not copy
    entry.importable = dom.selectImportable.value as Policies.ImportableMode;                                                                                                            
    applyAllowedAdditions(entry);                              // mutates in place
    putOrDeletePolicyEntry(selectedEntry, entry, Policies.finishEditing(...));                                                                                                           
  }                                        
                                                                                                                                                                                         
  If the PUT fails, the in-memory cache holds the new importable / allowedAdditions but the server doesn't. A subsequent Cancel re-reads the corrupted state via updateEditors.          
   
  The pattern was preexisting for importable, but this commit compounds it via applyAllowedAdditions(entry) (which does delete entry.allowedAdditions or assigns the new array). The two new sibling files already do this correctly:
                                                                                                                                                                                         
  - policiesNamespaces.ts:78 → const entry = { ...Policies.thePolicy.entries[...] };
  - policiesReferences.ts:251 → same spread
                                                                                                                                                                                         
  Suggest applying the same pattern here:
                                                                                                                                                                                         
  function onUpdateEntryClick() {          
    validateAllowedAdditionsSelection();
    const entry = { ...Policies.thePolicy.entries[selectedEntry] };
    entry.importable = dom.selectImportable.value as Policies.ImportableMode;                                                                                                            
    applyAllowedAdditions(entry);
    putOrDeletePolicyEntry(selectedEntry, entry, Policies.finishEditing(...));                                                                                                           
  }                                                                                                                                                                                      
   
  ---                                                                                                                                                                                    
  Otherwise the change is in good shape — the absent-vs-[] discipline for allowedAdditions (the trickiest part) is handled correctly, the empty() regex change is clean, and  Utils.confirm's switch to DOMPurify-sanitised innerHTML is safe.                                                                                                                       
  ```

thjaeckle:
Thanks for the thorough review @hu-ahmed — both findings were spot on. Addressed in 67ca5c9cef:

- **M1**: `loadEntriesForImport` now uses a monotonic token (`++loadEntriesToken`) captured at entry and re-checked after every await; stale resolves drop their result. The synchronous `(local)` path skips the guard (no suspension point). The `finally` only re-enables the dropdown when still the latest call — otherwise the newer call owns the disabled-state transition.
- **M2**: `onUpdateEntryClick` spread-copies the entry before mutating `importable` / `allowedAdditions`, so a 4xx leaves `Policies.thePolicy.entries[...]` matching server-truth and a subsequent Cancel re-paints from clean state. Matches the pattern already in `policiesNamespaces.ts` and `policiesReferences.ts`.

hu-ahmed (approved):

Issue #2291, opened 2025-12-10
Provide timeseries facade for Ditto feature properties


thjaeckle:
Ditto is often used in IoT landscapes together with timeseries databases.  
* Ditto ingesting data into the TS database (e.g. whenever a `temperature` value is changed)
* Ditto being able with custom implementation for handling messages to fetch data and return it to the caller

The integration of Ditto and a timeseries database is however much manual effort an Ditto does not provide an "own" API to e.g. fetch timeseries data, which complicates things.

This issue describes a proposal how to improve the integration of Ditto as Digital Twin IoT middleware and an arbitrary timeseries database.

It shall cover:
* Defining which data should be ingested 
   * E.g. based on a WoT ThingModel extension to declare in a WoT model which properties to ingest
* Specifying an HTTP API in order to fetch timeseries data
* Integrating policies to define access control to access timeseries data
* A way to support different timeseries databases
   * E.g. as Ditto "connections" - or maybe even as separate (optional) Ditto service

First rough idea about how the HTTP API could look like to access timeseries data:
```
GET /things/{thingId}/features/{featureId}/timeseries/{propertyPath}
```

Example OpenAPI snippet:
```yaml
/things/{thingId}/features/{featureId}/timeseries
    get:
      summary: Retrieve historical data for all properties of the targeted feature
      description: >
        Fetches the historical data (timeseries) for all existing properties.
      parameters:
        - name: thingId
          in: path
          description: The ID of the Thing.
          required: true
          schema:
            type: string
        - name: featureId
          in: path
          description: The ID of the Feature.
          required: true
          schema:
            type: string

        - name: properties
          in: query
          description: Optional comma separated list of properties to fetch timeseries data for. If omitted, all will be fetched.
          required: false
          schema:
            type: string
          example: "temperature,humidity"

        # --- Standardized Time Query Parameters ---
        - name: from
          in: query
          description: Start timestamp (ISO 8601).
          required: true
          schema:
            type: string
            format: date-time
          example: "2023-10-27T10:00:00Z"
        - name: to
          in: query
          description: End timestamp (ISO 8601).
          required: true
          schema:
            type: string
            format: date-time
          example: "2023-10-28T10:00:00Z"
        - name: step
          in: query
          description: >
            Downsampling interval (e.g., '5m', '1h'). 
            Crucial for performance when querying long time ranges.
          required: false
          schema:
            type: string
          example: "1h"
        - name: agg
          in: query
[...]

PR #2438, opened 2026-05-11
#2291 Add timeseries feature: WoT-driven ingest + sharded persistent …

hu-ahmed:
…entity + read API

Introduces a new timeseries microservice that ingests WoT-annotated Thing property events and serves a time-bounded read API for retrieving the resulting points.

NEW MODULES

- timeseries/model — public domain (TimeseriesQuery, TimeseriesDataPoint, TimeseriesDataValue, TimeseriesQueryResult, TimeseriesResultMeta, WotTimeseriesAnnotation, Aggregation / FillStrategy / Ingest enums). Source conservatively Java-8-compatible per the public-API rule.
- timeseries/api — internal SPI (TimeseriesAdapter + AdapterConfig, HealthStatus, IngestDataPoints command + response, TimeseriesMessagingConstants, TimeseriesMappingStrategies — pulls in PoliciesMappingStrategies so PolicyTag cluster pub-sub invalidations deserialize on the timeseries node).
- timeseries/mongodb — default Ditto-shipped adapter using MongoDB Time Series collections; per-namespace bucketing, optional retention via TimeSeriesOptions.expireAfter, lazy collection creation on the WRITE path only (reads never create collections).
- timeseries/service — TimeseriesService entry point + TimeseriesRootActor + per-Thing sharded persistent TimeseriesIngestActor that journals each batch before writing it to the adapter, mirroring ThingPersistenceActor durability.

SIGNAL + PROTOCOL ADAPTER

- RetrieveTimeseries command + RetrieveTimeseriesResponse signal in timeseries-model; the feature-toggle gate lives in RetrieveTimeseries' constructor (à la MergeThing) so a disabled-toggle deployment rejects the command synchronously at the gateway route's try/catch — HTTP 400 signal.unsupported in ms instead of a 60 s ask timeout.
- New TopicPath.Criterion.TIMESERIES + TimeseriesTopicPathBuilder. Wire TimeseriesQueryCommandAdapter / TimeseriesQueryCommandResponseAdapter into DittoProtocolAdapter, DefaultAdapterResolver, AdapterResolverBySignal so the read path works over WebSocket + Connectivity, not just HTTP.

WRITE PATH (things-service side)

- TimeseriesIngestPublisher: non-persistent per-node publisher; subscribes to ThingEvent stream, resolves the post-event Thing snapshot, looks up the ditto:timeseries WoT annotation, asks the timeseries shard region with IngestDataPoints, retries with bounded MAX_ATTEMPTS, drops on configuration anomalies. Gated by FeatureToggle.isTimeseriesFeatureEnabled — when off, ThingsRootActor never starts the publisher and ThingPersistenceActor.publishEvent null-guards on the ref (zero per-event cost).
- ThingPersistenceActor.publishEvent now snapshots `entity` on the actor thread BEFORE registering the async enrichment continuation, so the publisher receives the post-event Thing instead of whatever the live field holds when the lambda fires.

READ PATH

- EdgeCommandForwarderActor routes RetrieveTimeseries via the timeseries shard region using askWithRetryCommandForwarder — same shape as forwardToThings.
- TimeseriesIngestActor handles RetrieveTimeseries on the same per-Thing entity that journals ingests: resolves the thing's policyId via SudoRetrieveThing, loads the (cached) PolicyEnforcer via PolicyEnforcerProvider, verifies the configured permission (READ_TS by default, READ in simplified mode) per requested path. All authorization failures surface as 404 ThingNotAccessibleException — never 403 — per the auth-authz 404-not-403 discipline.
- ThreadSafeDittoLogger LOGGER for callback-context log calls in the enforcement chain; inherited single-threaded log stays for on-actor-thread sites.

CONFIGURATION + DEPLOYMENT

- Permission.READ_TS added to things-api; intentionally NOT in DEFAULT_THING_PERMISSIONS so existing Things don't auto-grant timeseries read access. OpenAPI permission enum + JSON schema documented.
- DITTO_DEVOPS_FEATURE_TIMESERIES_ENABLED feature toggle in FeatureToggle.java + HOCON ditto.devops.feature.timeseries-enabled + System-property bridge in DittoService.injectSystemPropertiesLimits.
- timeseries.conf with full ditto.timeseries.adapter.mongodb subtree (URI / database / collection-prefix / granularity / retention) and a cleanup { ... } block matching things.conf for persistence-cleanup parity.
- timeseries-dev.conf for local cluster; new IntelliJ run config; root Ditto.run.xml + bom + BOM services-zone wiring.
- Helm: timeseries-deployment / timeseries-pdb / timeseries-podmonitor templates; mongodb-secret extension; values.yaml block with adapter + enforcement knobs.
- Logback config + Kamon config.
- pekko.cluster.roles for timeseries-service includes blocked-namespaces-aware (required so CachingPolicyEnforcerProvider's ddata replicator joins the cluster's blocked-namespaces group).

UI

- New Timeseries tab on the feature view: path / from (UTC) / to (UTC) / limit form + sparkline + table; clears on thing change.

DOCUMENTATION

- OpenAPI: api-2-index.yml, paths/timeseries/property.yml, schemas/timeseries/timeseriesQueryResult.yml; permission schema doc updated.
- jsonschema/policy.json updated for READ_TS enum value with both-mode docs.

TESTS

- TimeseriesIngestActorTest (persistent-actor lifecycle) + TimeseriesIngestActorEnforcementTest (read-path enforcement truth table: granted / denied / configured-READ / thing-not-found / no-policyId / enforcer-empty) + TimeseriesIngestActorMongoJournalIT (MongoDB journal IT).
- TimeseriesIngestPublisherTest (publisher contract).
- Parametrized protocol adapter test extended with TimeseriesQueryCommandAdapter.
- Per-class JSON round-trip / EqualsVerifier / null-param tests across the new model classes.


thjaeckle:
@hu-ahmed I created a new branch [timeseries-main](https://github.com/eclipse-ditto/ditto/tree/timeseries-main) - we can target the TS PRs to that branch instead of `master` and also have a separate nightly build for that to test integration.

timeseries/service/src/main/resources/timeseries.conf
```diff
@@ -0,0 +1,212 @@
...
+  journal.auto-start-journals = [
+    "pekko-contrib-mongodb-persistence-timeseries-journal"
+  ]
+  snapshot-store.auto-start-snapshot-stores = [
+    "pekko-contrib-mongodb-persistence-timeseries-snapshots"
+  ]
+}
+
+# MongoDB journal for the per-Thing TimeseriesIngestActor entities.
+pekko-contrib-mongodb-persistence-timeseries-journal {
```
thjaeckle: @hu-ahmed do we need Pekko Persistence here?  
This is a config (and also above and below) of Pekko Persistence, it surprises me to see it in the Timeseries service

hu-ahmed: good point, it was part of previous approach that must change, will modify it now

PR #2440, opened 2026-05-11
Service CPU optimisations from JFR profiling

thjaeckle:
## Summary

CPU optimisations identified by 60-second to 5-minute JFR recordings on the things, things-search, and connectivity dev pods. Four commits, all independent and behaviour-preserving:

- **things-service (`66ef7c6eb0`)** — five fixes targeting the JFR-observed hot paths:
  - O(k) forward-index in `PolicyEnforcerCache.deregisterImportMappings` (was O(N) full-map scan); null-PolicyId short-circuit in `AbstractEnforcerActor.loadPolicyEnforcer` plus WARN→DEBUG downgrade (was producing ~200 k log lines/day on dev).
  - Static `CBORFactory` reuse in `JacksonSerializationContext` (was allocated per serialisation; top byte[]/int[] allocator).
  - `DefaultDittoJsonHandler.parseToIntegerOrLong` parses long-then-downcast instead of try-`Integer.parseInt`-catch-`NumberFormatException`-fallback (JFR observed ~113 NFE/s from 64-bit pub/sub hashes).
  - Bulk-copy fast path in `JavaStringToEscapedJsonString.apply` (no-escape strings) + `append(CharSequence, start, end)` runs instead of per-char `replace`.
  - `ask-with-retry-dispatcher`: add the missing `thread-pool-executor` block with `allow-core-timeout = off` (Pekko's default-dispatcher fallback was killing idle core threads after 60 s); raise `wot-dispatcher` / `wot-dispatcher-cache-loader` `parallelism-min` 4 → 8.

- **things-search (`3578855ef6`)** — two structural fixes mirroring the things-service patterns:
  - Same `thread-pool-executor` template for `thing-cache-dispatcher` and `policy-enforcer-cache-dispatcher` (JFR observed ~0.95 thread-births/s on thing-cache-dispatcher pre-fix).
  - Append `-Dio.netty.leakDetection.level=disabled` to `global.jvmOptions` in Helm. Netty defaults to SIMPLE which captures a Throwable stack per sampled buffer; ~0.56/s on the Pekko-remote/Mongo paths.

- **base/model `validateValueTypes` (`5b9c4b7381`)** — invert the loop in `AbstractDittoHeadersBuilder.validateValueTypes`. The pre-existing variant iterates *all* known `HeaderDefinition`s (~50+) and `headers.get(definitionKey)` per definition; the new `@since 3.9.0` overload iterates the (typically small) headers map and looks up each entry in a `Map<String, HeaderDefinition>` — O(H) instead of O(D). Hits on every inbound Pekko cluster message; the legacy `Collection` overload is kept `@Deprecated` for binary compatibility with external subclasses. Defensive null-value skip preserves identical observable behaviour.

- **`signal-enrichment-cache-dispatcher` (`fd13a72e9b`)** — repo-wide audit found two remaining occurrences of the same defect (gateway + connectivity); fixed identically. After this commit, no instances of `type = Dispatcher` + `InstrumentedThreadPoolExecutorServiceConfigurator` + missing `thread-pool-executor` block remain in `*/main/resources/*.conf`.

All new HOCON tunables are `${?ENV_VAR}`-overridable; Helm wiring (`{things,thingsSearch,gateway,connectivity}.config.dispatchers.*` + deployment-template env-var bindings) is included.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/headers/AbstractDittoHeadersBuilder.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/headers/AbstractDittoHeadersBuilder.java
@@ -124,7 +124,8 @@ public abstract class AbstractDittoHeadersBuilder<S extends AbstractDittoHeaders
 
         checkNotNull(initialHeaders, "initial headers");
         checkNotNull(definitions, "header definitions");
-        validateValueTypes(initialHeaders, definitions); // this constructor does validate the known value types
+        checkNotNull(definitionsMap, "definitionsMap");
+        validateValueTypes(initialHeaders, definitionsMap); // this constructor does validate the known value types
         myself = (S) selfType.cast(this);
         headers = preserveCaseSensitivity(initialHeaders);
         metadataHeaders = MetadataHeaders.newInstance();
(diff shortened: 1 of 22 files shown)


hu-ahmed (approved):
LGTM!
Thanks for the well-targeted optimizations

PR #2441, opened 2026-05-11
Convert enforcement AskTimeoutException to 503 instead of 500

### LINKED ISSUE #2439: Convert enforcement AskTimeoutException to HTTP 503 instead of 500
## Problem

During rolling restarts, Eclipse Ditto produces thousands of `DittoInternalErrorException` (HTTP 500) responses. These are not actual internal errors — they are transient timeouts caused by shard rebalancing in the policies service.

The root cause is a timeout hierarchy mismatch in `AbstractPersistenceSupervisor`:

```
Supervisor
  └─ askEnforcerChild(signal)            ← 10s timeout (localEnforcerAskTimeout)
       └─ ThingEnforcerActor
            └─ PolicyCacheLoader
                 └─ AskWithRetry(policiesShardRegion)  ← 3s × 3 retries + backoff = ~42s worst case
```

The supervisor's 10s enforcer timeout fires while the enforcer is still legitimately retrying the policy shard (which can take up to 42s during rebalancing). The resulting `AskTimeoutException` is a Pekko-native exception, not a `DittoRuntimeException`, so `getEnforcementExceptionAsRuntimeException()` wraps it as `DittoInternalErrorException` (HTTP 500) — the generic fallback for unknown throwables.

This is incorrect: enforcement timeouts during shard rebalancing are transient and retryable. They should produce HTTP 503 (Service Unavailable), not HTTP 500 (Internal Server Error).

### Impact

- Clients receive HTTP 500, which signals a permanent server bug rather than a transient condition. Most HTTP clients and load balancers do not retry 500s.
- Each occurrence is logged at ERROR level (the special `DittoInternalErrorException` logging path in `handleSignalEnforcementResponse`), generating noise that masks genuine internal errors.
- The enforcer child continues retrying the policy shard uselessly after the supervisor has already returned 500 to the caller.

## Proposed Fix

Introduce `EnforcementTimeoutException` — a dedicated `DittoRuntimeException` subclass that maps enforcement `AskTimeoutException` to HTTP 503 instead of 500.

alstanchev:
During rolling restarts, AskTimeoutException from the enforcer child was wrapped as DittoInternalErrorException (HTTP 500). This is misleading — the timeout is transient and clients should retry.

Introduce EnforcementTimeoutException (HTTP 503) and convert AskTimeoutException to it in getEnforcementExceptionAsRuntimeException. Also unwrap CompletionException wrappers from async chains before checking the exception type.

Fixes #2439

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/exceptions/EnforcementTimeoutException.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/exceptions/EnforcementTimeoutException.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.ditto.base.model.exceptions;
+
+import java.net.URI;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+import org.eclipse.ditto.base.model.common.HttpStatus;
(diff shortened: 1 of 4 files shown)


thjaeckle (approved):
LGTM 👍 
Thx @alstanchev 

We can also add this to 3.9.0

PR #2442, opened 2026-05-11
Helm: Allow configuring `issuers` for OIDC

JCapucho:
Adds support for configuring the `issuers`[^1] option of OpenID Connect issuers in the helm chart values directly.

[^1]: https://eclipse.dev/ditto/installation-operating.html#openid-connect

### DIFF
--- a/deployment/helm/ditto/service-config/gateway-extension.conf.tpl
+++ b/deployment/helm/ditto/service-config/gateway-extension.conf.tpl
@@ -15,6 +15,14 @@ ditto {
         {{- range $key, $value := .Values.gateway.config.authentication.oauth.openidConnectIssuers }}
           {{$key}} = {
             issuer = "{{$value.issuer}}"
+            {{- with $value.issuers }}
+            issuers = [
+              "{{$value.issuer}}"
+            {{- range $index, $issuer := . }}
+              "{{$issuer}}"
+            {{- end }}
+            ]
+            {{- end }}
             auth-subjects = [
             {{- range $index, $subject := $value.authSubjects }}
               "{{$subject}}"
@@ -78,6 +86,14 @@ ditto {
           {{- range $key, $value := .Values.gateway.config.authentication.devops.oauth.openidConnectIssuers }}
             {{$key}} = {
               issuer = "{{$value.issuer}}"
+              {{- with $value.issuers }}
+              issuers = [
+                "{{$value.issuer}}"
+              {{- range $index, $issuer := . }}
+                "{{$issuer}}"
+              {{- end }}
+              ]
+              {{- end }}
               auth-subjects = [
               {{- range $index, $subject := $value.authSubjects }}
                 "{{$subject}}"
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -2317,10 +2317,17 @@ gateway:
         # allowedClockSkew configures the amount of clock skew in seconds to tolerate when verifying the local time against the exp and nbf claims
         allowedClockSkew: 20s
         # openidConnectIssuers holds a map of issuer-prefixes as key (e.g. "example")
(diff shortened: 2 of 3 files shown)


deployment/helm/ditto/values.yaml
```diff
@@ -2271,6 +2271,10 @@ gateway:
         openidConnectIssuers:
         #  example:
         #    issuer: "example.com"
+        #    #issuers = [
+        #    #  "localhost:9000/one"
+        #    #  "localhost:9000/two"
+        #    #]
```
thjaeckle: The example uses HOCON syntax inside a YAML comment. Users editing `values.yaml` write YAML, not HOCON — if someone uncomments these lines as-is, Helm cannot consume them. It should mirror the YAML list style used a few lines below for `authSubjects:`, e.g.:

```yaml
#    issuers:
#    - "localhost:9000/one"
#    - "localhost:9000/two"
```

Also, the leading `#issuers = [` / trailing `#]` use a double `#` (comment-within-comment) which looks unintended.

Same applies to the identical block in the devops section (lines 2326–2329).

While you're here, consider adding one line clarifying that `issuers` should contain **only additional** issuers — the primary `issuer` value is added automatically by the template (see my other comment on `gateway-extension.conf.tpl`).

deployment/helm/ditto/service-config/gateway-extension.conf.tpl
```diff
@@ -15,6 +15,14 @@ ditto {
         {{- range $key, $value := .Values.gateway.config.authentication.oauth.openidConnectIssuers }}
           {{$key}} = {
             issuer = "{{$value.issuer}}"
+            {{- with $value.issuers }}
+            issuers = [
+              "{{$value.issuer}}"
```
thjaeckle, line 20: The primary `$value.issuer` is always prepended to the rendered `issuers = [...]` list. This is necessary because `DefaultSubjectIssuerConfig` ignores the singular `issuer` field once `issuers` is non-empty — so without this line the primary issuer would silently drop out. Good.

The side-effect, though, is that if a user lists their primary issuer in `issuers:` too (a very natural thing to do), it appears twice in the rendered HOCON. Two ways to resolve:

1. Keep the current behavior and document in `values.yaml` that `issuers:` should hold **only additional** issuers (the primary is added automatically).
2. Drop the prepended `"{{$value.issuer}}"` and require users to repeat the primary issuer in `issuers:` themselves — this matches the existing `gateway.conf` example at lines 267–271.

Either is fine, but please pick one and make it explicit in the `values.yaml` comment so future users aren't surprised.

Same applies to the identical block in the devops section (around line 89).

thjaeckle (changes requested):
Thanks for wiring this through — the Helm change is small and backwards-compatible. Two changes requested before merging; see inline comments. The rest (example trailing-slash, comment phrasing) is just polish and not blocking.

One more thing: please bump the Helm chart `version` in `deployment/helm/ditto/Chart.yaml` — Helm chart linting requires a version bump when the chart content changes, otherwise the lint job fails.

JCapucho:
@thjaeckle Sorry for the HOCON totally forgot to translate it.

I've added a comment explaining that the `issuers` option is in addition to the existing `issuer` value. I think this option makes more sense, because having the `issuer` option be accepted and then ignored completely is an unexpected behavior in my opinion.

thjaeckle (approved):
LGTM, thanks @JCapucho for the contribution 👍 

Will be part of Helm chart version 4.1.0

thjaeckle:
@JCapucho I just noticed the Helm linting still failed.
Can you please bump the Helm chart version to `4.1.0` by yourself as I don't have permissions to make changes on your fork?

JCapucho:
@thjaeckle Done, not exactly sure why changes can't be made by maintainers, I'll check next time.

PR #2444, opened 2026-05-11
Fix `ssl-config` not being picked up

### LINKED ISSUE #2443: Ditto not picking up `ssl-config`
Configuring `ssl-config` for using a self-signed certificate with OpenID Connect as specified in the [documentation](https://eclipse.dev/ditto/installation-operating.html#openid-connect) doesn't work.

This seems to be a problem that was introduced in https://github.com/apache/pekko-http/commit/53365087aa1fe029fb4522c9d40045673cb23b6e where the default https context no longer uses `ssl-config`.

Using the (now deprecated) `createDefaultClientHttpsContext()` restores the old behaviour.

```patch
--- a/internal/utils/http/src/main/java/org/eclipse/ditto/internal/utils/http/DefaultHttpClientFacade.java
+++ b/internal/utils/http/src/main/java/org/eclipse/ditto/internal/utils/http/DefaultHttpClientFacade.java
@@ -72,8 +72,9 @@
 
     @Override
     public CompletionStage<HttpResponse> createSingleHttpRequest(final HttpRequest request) {
+        // https://github.com/apache/pekko-http/commit/53365087aa1fe029fb4522c9d40045673cb23b6e
         return Http.get(actorSystem)
-                .singleRequest(request, Http.get(actorSystem).defaultClientHttpsContext(),
+                .singleRequest(request, Http.get(actorSystem).createDefaultClientHttpsContext(),
                         connectionPoolSettings,
                         actorSystem.log()
                 );
```

JCapucho:
Fixes #2443

This fixes the problem but uses a deprecated method (`createDefaultClientHttpsContext()`)

### DIFF
--- a/internal/utils/http/src/main/java/org/eclipse/ditto/internal/utils/http/DefaultHttpClientFacade.java
+++ b/internal/utils/http/src/main/java/org/eclipse/ditto/internal/utils/http/DefaultHttpClientFacade.java
@@ -72,8 +72,9 @@ public final class DefaultHttpClientFacade implements HttpClientFacade {
 
     @Override
     public CompletionStage<HttpResponse> createSingleHttpRequest(final HttpRequest request) {
+        // https://github.com/apache/pekko-http/commit/53365087aa1fe029fb4522c9d40045673cb23b6e
         return Http.get(actorSystem)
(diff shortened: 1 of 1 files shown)


thjaeckle (approved):
LGTM, thanks @JCapucho .. unwanted and unexpected side-effect of library updates

PR #2445, opened 2026-05-11
Implement X509 authentication for MongoDB connection

JCapucho:
Adds support for using X509 authentication when connecting to MongoDB.

Also allows configuring the CA root certificate used in the TLS connection to MongoDB.

### DIFF
--- a/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
+++ b/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
@@ -14,8 +14,7 @@ package org.eclipse.ditto.internal.utils.persistence.mongo;
 
 import static org.eclipse.ditto.base.model.common.ConditionChecker.checkNotNull;
 
-import java.security.KeyManagementException;
-import java.security.NoSuchAlgorithmException;
+import java.io.File;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
@@ -23,7 +22,13 @@ import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
-import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLException;
+
+import com.mongodb.MongoCredential;
+
+import io.netty.handler.ssl.SslContext;
+
+import io.netty.handler.ssl.SslContextBuilder;
 
 import org.bson.Document;
 import org.bson.codecs.configuration.CodecRegistry;
@@ -349,10 +354,15 @@ public final class MongoClientWrapper implements DittoMongoClient {
         @Nullable private ConnectionString connectionString;
         private String defaultDatabaseName;
         private boolean sslEnabled;
+        private String sslCa;
         private boolean isUseAwsIamRole;
         private String awsRegion;
         private String awsRoleArn;
         private String awsSessionName;
+        private boolean isUseX509Authentication;
+        private String sslClientCert;
+        private String sslClientKey;
+        private String sslClientKeyPassword;
         @Nullable private EventLoopGroup eventLoopGroup;
 
         private MongoClientWrapperBuilder() {
@@ -401,6 +411,7 @@ public final class MongoClientWrapper implements DittoMongoClient {
 
             final MongoDbConfig.OptionsConfig optionsConfig = mongoDbConfig.getOptionsConfig();
             builder.enableSsl(optionsConfig.isSslEnabled());
+            builder.sslCaFile(optionsConfig.sslCaFile());
             builder.useAwsIamRole(optionsConfig.isUseAwsIamRole());
             builder.awsRegion(optionsConfig.awsRegion());
             builder.awsRoleArn(optionsConfig.awsRoleArn());
@@ -409,6 +420,10 @@ public final class MongoClientWrapper implements DittoMongoClient {
             builder.setReadConcern(optionsConfig.readConcern().getMongoReadConcern());
             builder.setWriteConcern(optionsConfig.writeConcern());
             builder.setRetryWrites(optionsConfig.isRetryWrites());
+            builder.useX509Authentication(optionsConfig.isUseX509Authentication());
+            builder.sslClientCertFile(optionsConfig.sslClientCertFile());
+            builder.sslClientKeyFile(optionsConfig.sslClientKeyFile());
+            builder.sslClientKeyPassword(optionsConfig.sslClientKeyPassword());
 
             return builder;
         }
@@ -446,6 +461,12 @@ public final class MongoClientWrapper implements DittoMongoClient {
         @Override
         public MongoClientWrapperBuilder enableSsl(final boolean enabled) {
             sslEnabled = enabled;
+            return this;
+        }
+
+        @Override
+        public MongoClientWrapperBuilder sslCaFile(final String sslCaFile) {
+            this.sslCa = sslCaFile;
             return this;
         }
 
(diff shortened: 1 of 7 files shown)


internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -563,40 +608,50 @@ public MongoClientWrapper build() {
                         AwsAuthenticationHelper.provideAwsIamBasedMongoCredential(awsRegion, awsRoleArn, awsSessionName)
                 );
             }
+
+            if (isUseX509Authentication) {
+                mongoClientSettingsBuilder.credential(MongoCredential.createMongoX509Credential());
+            }
```
thjaeckle, line 614: Two related issues here:

- **Precedence inconsistency vs `CustomizableScalaDriverPersistenceExtension`**: that file uses `if AWS else if X509 else null`, so AWS wins if both are enabled. Here, both `if` blocks run independently and X509 silently overwrites the AWS credential. Same Ditto deployment, two MongoDB code paths, two different credentials applied. Suggest either failing fast at config-load if both flags are set, or aligning the precedence rule in both files.
- **No prerequisite validation**: `createMongoX509Credential()` requires (a) `ssl=true` and (b) a configured client certificate. Neither is enforced. With `useX509Authentication=true` + `ssl=false`, or with empty cert paths, the failure surface is a cryptic MongoDB handshake error rather than a clear config error. Recommend validating in `DefaultOptionsConfig`'s constructor and throwing a `DittoConfigError`.

internal/utils/persistence/src/main/scala/org/eclipse/ditto/internal/utils/persistence/pekko/CustomizableScalaDriverPersistenceExtension.scala
```diff
@@ -41,16 +46,31 @@ class CustomizableScalaDriverPersistenceExtension(val actorSystem: ActorSystem)
...
           optionsConfig.awsRoleArn(),
           optionsConfig.awsSessionName()
         )
-        new CustomizableScalaMongoDriver(actorSystem, config, builder => builder.credential(mongoCredential))
+      } else if (optionsConfig.isUseX509Authentication) {
+        MongoCredential.createMongoX509Credential()
       } else {
-        new ScalaMongoDriver(actorSystem, config)
+        null
       }
```
thjaeckle, line 59: Precedence here differs from `MongoClientWrapper.build()`, which applies both `if (isUseAwsIamRole)` and `if (isUseX509Authentication)` as independent blocks (X509 wins). Here it's an `if/else if` — AWS wins. The two code paths should not diverge — pick one rule and apply it in both places, ideally backed by a config-time validation error when both flags are set.

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -563,40 +608,50 @@ public MongoClientWrapper build() {
...
+            if (sslCa != null && !sslCa.isEmpty()) {
+                final var sslCaFile = new File(sslCa);
+                builder.trustManager(sslCaFile);
+            }
+
+            if (sslClientCert != null && !sslClientCert.isEmpty()) {
+                final var sslClientCertFile = new File(sslClientCert);
+                final var sslClientKeyFile = new File(sslClientKey);
+                builder.keyManager(sslClientCertFile, sslClientKeyFile,
+                        sslClientKeyPassword.isEmpty() ? null : sslClientKeyPassword);
```
thjaeckle, line 651: NPE risk: `sslClientCert` is null/empty-guarded, but `sslClientKey` is used directly in `new File(sslClientKey)`, and `sslClientKeyPassword.isEmpty()` is called without a null guard. In practice `DefaultOptionsConfig` reads via `config.getString(...)` with `""` defaults so this survives the happy path, but a programmatic builder caller or a future refactor that allows nulls will NPE here. The same pattern is in `CustomizableScalaDriverPersistenceExtension.scala:99-104`. Either add explicit null/empty guards for the key + password, or fail fast at config load if `sslClientCertFile` is set without `sslClientKeyFile`.

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -563,40 +608,50 @@ public MongoClientWrapper build() {
...
-                                .enabled(sslEnabled));
-            }
+            eventLoopGroup = new NioEventLoopGroup();
+            mongoClientSettingsBuilder
+                    .transportSettings(TransportSettings.nettyBuilder()
+                            .eventLoopGroup(eventLoopGroup)
+                            .sslContext(tryToCreateAndInitSslContext(sslCa, sslClientCert, sslClientKey, sslClientKeyPassword))
+                            .build())
+                    .applyToSslSettings(builder -> builder.enabled(sslEnabled));
         }
```
thjaeckle, line 628: Behavior change worth noting: previously the SSL setup was guarded by `if (sslEnabled)` with a fallback `else if (null != connectionString)`. Now it runs unconditionally — every client allocates a `NioEventLoopGroup` and an `SslContext` even when SSL is disabled. Two consequences:
- Extra threads + potentially native resources for non-SSL deployments.
- The old fallback for builders constructed without a connection string is gone. Probably benign in production paths, but it's an unannounced behavior shift.

Suggest keeping the `if (sslEnabled)` guard around the Netty transport/SslContext setup, and outside it just call `applyToSslSettings(builder -> builder.enabled(false))`.

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/MongoClientWrapper.java
```diff
@@ -563,40 +608,50 @@ public MongoClientWrapper build() {
...
             }
         }
 
-        private static SSLContext createAndInitSslContext() throws NoSuchAlgorithmException, KeyManagementException {
-            final SSLContext result = SSLContext.getInstance("TLSv1.2");
-            result.init(null, null, null);
-            return result;
+        private static SslContext createAndInitSslContext(String sslCa, String sslClientCert, String sslClientKey, String sslClientKeyPassword)
+                throws SSLException {
+            final var builder = SslContextBuilder.forClient();
```
thjaeckle, line 640: Quiet protocol change: the previous code pinned `SSLContext.getInstance("TLSv1.2")` (TLS 1.2 only). Netty's default here enables TLS 1.2 + 1.3, which is a net improvement — but it's a behavior change worth calling out in the PR description, and worth being explicit about: consider `SslContextBuilder.forClient().protocols("TLSv1.3", "TLSv1.2")` so a future Netty default regression can't silently widen the allowed set.

internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/config/DefaultOptionsConfig.java
```diff
@@ -163,17 +190,23 @@ public boolean equals(final Object o) {
...
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(useAwsIamRole, awsRegion, awsArnRole, awsArnRole, awsSessionName, sslEnabled,
-                readPreference, readConcern, writeConcern, retryWrites, extraUriOptions);
+        return Objects.hash(useAwsIamRole, awsRegion, awsArnRole, awsArnRole, awsSessionName, sslEnabled, sslCaFile,
+                readPreference, readConcern, writeConcern, retryWrites, extraUriOptions, useX509Authentication,
+                sslClientCertFile, sslClientKeyFile, sslClientKeyPassword);
     }
```
thjaeckle, line 210: While you're touching `equals`/`hashCode` for the new fields: `toString()` (just below this block) wasn't updated and still only prints the original set. Accidentally safe for `sslClientKeyPassword` — which should NOT be logged — but the inconsistency is fragile. Suggest:
- Add the four non-sensitive new fields (`sslCaFile`, `useX509Authentication`, `sslClientCertFile`, `sslClientKeyFile`) to `toString()`.
- Deliberately omit `sslClientKeyPassword` and add a one-line comment explaining why, so a future contributor doesn't "fix" it.

thjaeckle (changes requested):
Hi @JCapucho — thanks for the contribution, this is a genuinely useful feature (X509 + custom CA covers both AWS DocumentDB and self-hosted MongoDB with private CAs). The TLS crypto plumbing itself looks correct (system trust store as default, custom CA additive, no insecure modes exposed). I've left inline comments on a handful of issues I'd like to see addressed before merge:

1. **Credential precedence diverges between the Java and Scala code paths** (`MongoClientWrapper` vs `CustomizableScalaDriverPersistenceExtension`) when both AWS IAM and X509 are configured. Same deployment, two different outcomes.
2. **NPE risk** if `sslClientCertFile` is set without `sslClientKeyFile` (or with a null password).
3. **No prerequisite validation** for X509 — should fail fast when `useX509Authentication=true` is set without `ssl=true` or without a client cert.
4. **`buildAndApplySslSettings` now runs unconditionally**, allocating a Netty event loop + SslContext even when SSL is disabled.
6. **TLS protocol set change** (was TLS 1.2 only, now Netty default = 1.2+1.3) is a net improvement but should be explicit, not implicit.
7. **`DefaultOptionsConfig#toString()`** missing the new fields (please omit the password deliberately and add a comment so it stays that way).

Happy to help review the next iteration.

PR #2446, opened 2026-05-12
Make 4xx DittoRuntimeException subclasses stackless by default

thjaeckle:
## Summary

`DittoRuntimeException` subclasses representing flow-control — HTTP status `< 500`, e.g. `ThingNotAccessibleException`, `PolicyNotAccessibleException`, `WotThingModelPayloadValidationException`, `MessageSendNotAllowedException` — now omit their stack trace and suppressed-exception list. This eliminates the per-throw `fillInStackTrace()` native call and `StackTraceElement[]` allocation, which JFR profiling on a gateway pod showed to be the dominant cost in the exception path (376 `ThingNotAccessibleException` / 309 s ≈ 1.22/s, ~30 frames captured each — roughly 720 kB/s of `StackTraceElement[]` garbage on the gateway alone).

Exceptions with HTTP status `>= 500` (`INTERNAL_SERVER_ERROR`, `BAD_GATEWAY`, `SERVICE_UNAVAILABLE`, `GATEWAY_TIMEOUT`, …) **always keep their stack trace** because they signal real bugs / infrastructure problems where the stack is diagnostic.

### Implementation

Centralised inside `DittoRuntimeException` itself — **one new private static helper** `computeWritableStackTrace(HttpStatus)` decides per construction based on `httpStatus.getCode() < 500 && FeatureToggle.isStacklessFlowControlExceptionsEnabled()`. All 232 subclasses inherit the new behaviour automatically without per-subclass code changes. A new `protected` 8-arg constructor takes an explicit `writableStackTrace` flag as an escape hatch for subclasses that need to override the default policy.

**Wire format unchanged** — `DittoRuntimeException.toJson` emits only `status` / `errorCode` / `message` / `description` / `href`; stack frames were never serialised to clients.

### Feature toggle

`ditto.devops.feature.stackless-flow-control-exceptions-enabled` (default `true`) restores the legacy behaviour — every Ditto exception captures a writable stack trace — when set to `false`. Plumbed end-to-end following the same pattern as every other toggle in `FeatureToggle.java`:

1. HOCON entry in `ditto-devops.conf` with `${?DITTO_DEVOPS_FEATURE_STACKLESS_FLOW_CONTROL_EXCEPTIONS_ENABLED}` env-var interpolation
2. HOCON→`System.setProperty` bridge in `DittoService.injectSystemPropertiesLimits`
3. Helm `global.featureFlags.stacklessFlowControlExceptionsEnabled` value + per-service deployment-template env-var binding (all 5 services)

### Side effects to be aware of

- SLF4J `log.X("...", e)` with a 4xx DRE as `Throwable` arg prints just `"ClassName: message"` instead of a full stack. Most Ditto logging uses the `"... <{}>", e` form which calls `toString()` and is unaffected.
- `Throwable.addSuppressed()` on a 4xx DRE becomes a no-op. The codebase has 3 production call sites (`JsonFieldsEncryptor`, `DocumentProcessor`, `ThingSupervisorActor`); all three use suppressed exceptions as diagnostic aids, not data.
- Operators who today rely on a 4xx DRE's stack trace for triage can flip the toggle off via `-Dditto.devops.feature.stackless-flow-control-exceptions-enabled=false` or the new Helm value.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/exceptions/DittoRuntimeException.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/exceptions/DittoRuntimeException.java
@@ -34,6 +34,7 @@ import org.eclipse.ditto.base.model.headers.WithManifest;
 import org.eclipse.ditto.base.model.json.FieldType;
 import org.eclipse.ditto.base.model.json.JsonSchemaVersion;
 import org.eclipse.ditto.base.model.json.Jsonifiable;
+import org.eclipse.ditto.base.model.signals.FeatureToggle;
 import org.eclipse.ditto.json.JsonFactory;
 import org.eclipse.ditto.json.JsonField;
 import org.eclipse.ditto.json.JsonFieldDefinition;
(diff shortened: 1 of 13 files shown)


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/25716670559

hu-ahmed (approved):
LGTM!
Thanks for the improvements

PR #2447, opened 2026-05-12
Make SSE publisher backpressure buffer size configurable

thjaeckle:
## Summary

- The buffer size of the SSE publisher source queue in `ThingsSseRouteBuilder` was hardcoded to `10`. Once a client consumes server-sent events slower than the upstream produces them, the buffer fills up and Pekko emits frequent `Backpressuring because buffer is full and overflowStrategy is: [Backpressure] in stream [queueSource]` WARN logs from `org.apache.pekko.stream.impl.QueueSource`. No data is lost (the upstream is backpressured), but the log noise is non-trivial and the threshold of 10 is much tighter than what's used for the WebSocket publisher (default 200).
- Make the SSE publisher buffer size configurable via:
  - HOCON: `ditto.gateway.streaming.sse.publisher.backpressure-buffer-size`
  - Env var: `GATEWAY_SSE_PUBLISHER_BACKPRESSURE`
  - Helm value: `gateway.config.sse.publisher.backpressureBufferSize`
- Raise the default from `10` → `100` (still well below the WebSocket default of 200; SSE messages are bigger and one event per `keepAlive` heartbeat already pushes through).

## Changes

| File                                                                          | Change                                                                                   |
| ----------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
| `gateway/.../streaming/SseConfig.java`                                        | Add `getPublisherBackpressureBufferSize()` + `PUBLISHER_BACKPRESSURE_BUFFER_SIZE` enum   |
| `gateway/.../streaming/DefaultSseConfig.java`                                 | Wire field + equals/hashCode/toString                                                    |
| `gateway/service/src/main/resources/gateway.conf`                             | New HOCON setting with env-var override                                                  |
| `gateway/.../sse/ThingsSseRouteBuilder.java`                                  | Both hardcoded `SupervisedStream.sourceQueue(10)` calls now use the config value         |
| `gateway/.../DefaultSseConfigTest.java` + `sse-test.conf`                     | New unit test (defaults, file values, equals/hashCode)                                   |
| `deployment/helm/ditto/values.yaml` + `gateway-deployment.yaml`               | Expose the new value via helm                                                            |

### DIFF
--- a/deployment/helm/ditto/Chart.yaml
+++ b/deployment/helm/ditto/Chart.yaml
@@ -16,7 +16,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 4.0.1  # chart version is effectively set by release-job
+version: 4.0.2  # chart version is effectively set by release-job
 appVersion: 3.9.0-M3
 keywords:
   - iot-chart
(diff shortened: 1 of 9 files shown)


hu-ahmed (approved):
LGTM!

PR #2449, opened 2026-05-13
Implement X509 authentication for MongoDB connection (#2445 follow-up)

thjaeckle:
## Summary

Adds support for **X509 client-certificate authentication** to MongoDB, and a configurable **CA root certificate** for the TLS connection. Useful for AWS DocumentDB and self-hosted MongoDB deployments using a private CA.

This PR is based on the work in **#2445 by @JCapucho** (commit `7ec10f1a03`, author preserved) and adds a follow-up commit (`544b54e8dc`) that addresses the review feedback on that PR. Opening this as a new PR because `maintainer_can_modify` was not set on #2445; #2445 is superseded by this one.

### Configuration (new `ditto.mongodb.options.*` keys)

| Key                    | Default | Purpose                                                                |
| ---------------------- | ------- | ---------------------------------------------------------------------- |
| `useX509Authentication`| `false` | Authenticate to MongoDB with an X509 client certificate (`MONGO-X509`) |
| `sslCAFile`            | `""`    | CA root certificate used for the TLS connection to MongoDB             |
| `sslClientCertFile`    | `""`    | Client certificate file (PEM) for X509 authentication                  |
| `sslClientKeyFile`     | `""`    | Client key file (PEM) for X509 authentication                          |
| `sslClientKeyPassword` | `""`    | Optional encryption password for the client key                        |

Helm values + env-var overrides are wired through accordingly.

### Review-feedback fixes applied on top of @JCapucho's commit

- **Config validation at load time** (`DefaultOptionsConfig`) — fails fast with a clear `DittoConfigError` when:
  - `useAwsIamRole` and `useX509Authentication` are both enabled (eliminates an AWS-vs-X509 precedence divergence between the Java and Scala code paths at the source).
  - `useX509Authentication=true` is set without `ssl=true`.
  - `useX509Authentication=true` is set without both `sslClientCertFile` and `sslClientKeyFile`.
- **SSL setup gated by `if (sslEnabled)`** in both `MongoClientWrapper` (Java) and `CustomizableScalaDriverPersistenceExtension` (Scala) — non-SSL deployments no longer allocate a `NioEventLoopGroup` + `SslContext`.
- **TLS protocols pinned explicitly** to `TLSv1.3`, `TLSv1.2` so a future Netty default cannot silently widen the allowed set. The pre-PR code pinned `TLSv1.2` only; the broadened set is intentional and now documented.
- **Defensive null/empty guards** for `sslClientKeyFile` / `sslClientKeyPassword` in both Java and Scala SslContext builders — programmatic callers passing nulls now fail with a clear `IllegalArgumentException` instead of an NPE.
- **`DefaultOptionsConfig.toString()`** updated to include `sslCaFile`, `useX509Authentication`, `sslClientCertFile`, `sslClientKeyFile`, and `sslClientKeyPassword=***` (deliberately masked, with an inline comment so a future contributor doesn't "fix" it to print the real value).
- **Tests** in `DefaultOptionsConfigTest`: extended `toString` assertion + three negative tests for the new config validations.

Targets milestone **3.9.0**.

Closes #2445 (superseded).

### DIFF
--- a/build-images.sh
+++ b/build-images.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/internal/utils/config/src/main/resources/ditto-mongo.conf
+++ b/internal/utils/config/src/main/resources/ditto-mongo.conf
@@ -35,6 +35,9 @@ ditto.mongodb {
     ssl = false
     ssl = ${?MONGO_DB_SSL_ENABLED}
 
+    sslCAFile = ""
+    sslCAFile = ${?MONGO_DB_CA_FILE}
+
     # read preference is one of: primary, primaryPreferred, secondary, secondaryPreferred, nearest
     readPreference = primary
     readPreference = ${?MONGO_DB_READ_PREFERENCE}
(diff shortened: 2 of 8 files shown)


thjaeckle:
@JCapucho as we want to perform the Ditto 3.9.0 release today and this is a feature which would have to wait otherwise to the next Ditto minor release (which would take months), I took the liberty to fix the findings of your PR #2445 in a new PR as I could not push changes to your fork.

PR #2453, opened 2026-05-18
optimize partial-access-paths event enrichment hot path

thjaeckle:
## Summary

JFR on production `things` pods running Ditto 3.9.0 showed roughly 25–30 % of total CPU spent in the partial-access-paths event enrichment path added in 3.9.0 (`ThingEventEnricher` → `PartialAccessPathCalculator` → `TreeBasedPolicyEnforcer.getAccessiblePathsForSubjects`). This PR reduces per-event cost and per-event allocation substantially without changing the emitted header.

## Changes

- Memoize the root-resource READ `SubjectClassification` on the immutable `PolicyEnforcer` instance via a new `getRootResourceReadClassification()`. The full policy-tree walk is now paid once per cached enforcer, not per emitted ThingEvent. `PartialAccessPathCalculator` additionally short-circuits when no subject has any READ grant at the root resource — the dominant production case.

- In `TreeBasedPolicyEnforcer`:
  - Skip the per-leaf prefix rebuild in `getAccessiblePaths`, `buildJsonView`, and `getAccessiblePathsForSubjects` when the resource path is empty (always true for the thing-root call path). Avoids one fresh `PointerAndValue` + `ImmutableJsonPointer` per Thing-JSON leaf.
  - Skip `getSubPointer` / `Optional` / `resourcePath.append` inside `extractAccessiblePaths` and `filterEntries` when the resource path is empty.
  - Replace the per-pointer prefix walk in `isAccessible` (which allocated a fresh `ImmutableJsonPointer` + `Optional` per depth level and did two `HashSet.contains` calls that hashed the full pointer hierarchy) with a `PathTrie` of granted+revoked paths built once per subject and walked alongside the pointer's `JsonKey` segments. Per-pointer check is now O(depth) with zero allocation.

- Rewrite `PartialAccessPathCalculator.collapseLeavesToAncestors` to walk a `LeafTrie` of accessible leaves alongside the Thing-JSON descent using a mutable `Deque<JsonKey>` path stack. A `JsonPointer` is materialized only at emission, and subtrees with no matching leaves short-circuit without descending. Drops the per-field `ImmutableJsonPointer` allocation that was the JFR no 2 allocation site at 6.73 %.

## Measured impact

JMH `AverageTime`, prod-matching memory-optimized `TreeBasedPolicyEnforcer`, Java 25, single fork, 3 warmup + 5 measurement iterations × 2 s. Lower is better.

| Scenario                  | Master            | This PR          | Speedup    |
| ------------------------- | ----------------- | ---------------- | ---------- |
| `noRestrictedSubjects`    | 1.95 µs/op        | **0.16 µs/op**   | **12.7×**  |
| `smallThingRestricted`    | 13.01 µs/op       | **5.20 µs/op**   | **2.5×**   |
| `largeThingRestricted`    | 310.89 µs/op      | **73.79 µs/op**  | **4.2×**   |
| `manyRestrictedSubjects`  | 1882.92 µs/op     | **364.39 µs/op** | **5.2×**   |

Per-operation allocation also drops sharply: 940 KB → 282 KB (−69 %) on `largeThingRestricted`, 5.4 MB → 711 KB (−87 %) on `manyRestrictedSubjects`. The only long-lived added state is the `SubjectClassification` memo on each cached `PolicyEnforcer` — three immutable `Set` references sharing the subject objects already alive in the wrapped policy, negligible against the policy-enforcer cache size.

## Compatibility

No behavior change: the `ditto-partial-access-paths` header format and contents are unchanged. Existing unit tests pass unmodified:

- `policies/model`: 1048 / 1048
- `things/service`: 619 / 619
- `policies/enforcement`: 261 / 261

A new JMH benchmark (`PartialAccessPathCalculatorBenchmark`, test scope) covers the four scenarios above as a regression net for future changes in this path.


hu-ahmed (approved):
LGTM!
Thanks for the nice improvement

thjaeckle:
System tests also passed: https://github.com/eclipse-ditto/ditto/actions/runs/26028359761

PR #2455, opened 2026-05-19
fix service account annotations and labels

kalinkostashki:
- added indent 4 for annotations -> if you had more than one it would fail to render
- lables moved to appropriate key


thjaeckle (approved):
LGTM, thanks @kalinkostashki

PR #2460, opened 2026-05-26
remove regex from ImmutableJsonPointer escape/parse hot path

thjaeckle:
## Summary

JFR profiling of production `things-service` pods on Ditto 3.9.1 showed `ImmutableJsonPointer.escapeTilde` as the single hottest Ditto method at **~3.8 % of CPU**, driven by `Matcher` allocation and regex execution on every `JsonKey` during `JsonPointer.toString()`. `decodeTilde` and the consecutive-slash check on the parse path used the same pattern-based approach.

Replace the three regex-based sites with manual single-pass char scans that fast-return the input string unchanged when no tilde is present — the dominant production case.

## Changes

- `escapeTilde(JsonKey)` &mdash; `indexOf('~')`-then-`StringBuilder` scan. Returns the original `String` unchanged when no `~` is present; otherwise walks the string once, replacing `~` with `~0`.
- `decodeTilde(CharSequence)` &mdash; manual scan recognising `~0` → `~`. Returns `keyString.toString()` when no `~` is present.
- `containsConsecutiveSlashes(CharSequence)` &mdash; replaces `DOUBLE_SLASH_PATTERN.matcher(...).find()`.
- Removed unused `Matcher` import and three `Pattern` constants. `SINGLE_SLASH_REGEX_PATTERN` (the lookahead-based split on line 86) is deliberately retained &mdash; out of scope for this PR.

Behaviour is character-for-character identical to the prior implementation. Verified by 14 new edge-case tests covering leading/trailing/lone/adjacent tildes, `~0~0`, `~1`, `~`-followed-by-non-digit, full round-trip, and leading/trailing/triple double-slash rejection.

## Measured impact

JMH `AverageTime`, single fork, 3 warmup + 5 measurement iterations × 2 s, Temurin 25. Lower is better.

| Scenario                  | Before (regex)  | After (manual)  | Speedup    |
| ------------------------- | --------------- | --------------- | ---------- |
| `toStringShallowNoTilde`  |  76.4 ns/op     |  64.1 ns/op     | **1.19×**  |
| `toStringDeepNoTilde`     | 229.2 ns/op     | 137.0 ns/op     | **1.67×**  |
| `toStringDeepWithTilde`   | 312.8 ns/op     | 173.4 ns/op     | **1.80×**  |
| `ofParsedShallowNoTilde`  | 195.4 ns/op     | 117.6 ns/op     | **1.66×**  |
| `ofParsedDeepNoTilde`     | 690.6 ns/op     | 306.3 ns/op     | **2.25×**  |
| `ofParsedDeepWithEscape`  | 952.3 ns/op     | 371.0 ns/op     | **2.57×**  |

The shallow `toString` floor (1.19×) reflects that only one key is escaped; the per-key win dominates on deeper paths.

## Compatibility

No public API change: `escapeTilde` and `decodeTilde` are private; `containsConsecutiveSlashes` is a private helper. No `@since` tag needed. All 931 tests in `ditto-json` pass unmodified.

A new JMH benchmark (`ImmutableJsonPointerBenchmark`, test scope) covers the six scenarios above and serves as a regression net for future changes in this path. JMH wiring in `json/pom.xml` mirrors the additions made in #2453.

### DIFF
--- a/json/pom.xml
+++ b/json/pom.xml
@@ -76,10 +76,34 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths combine.children="append">
+                        <path>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
+                            <version>${jmh.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>flatten-maven-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/26500263307

hu-ahmed (approved):
LGTM!
maybe just squash the commits before merging

PR #2465, opened 2026-06-11
fix partial-access filter stripping thingId and audit fields from SSE/WebSocket events

thjaeckle:
## Summary

The 3.9.0 partial-events feature (#2287) filters the emitted Thing JSON by
exact path match against the subscriber's policy grants. Top-level entity
fields that aren't a listed grant get stripped, including `thingId`. A
subscriber with READ on `thing:/attributes` therefore receives SSE events
without any way to identify which thing the change applies to — a behavior
regression versus 3.8.x.

The same code path also strips identity/audit fields from WebSocket payloads
of full-thing events (`ThingCreated`/`ThingModified`/`ThingMerged`) and from
user-requested `extraFields`.

## Fix

Always preserve `thingId`, `_namespace`, `_revision`, `_modified` and
`_created` when the subscriber holds at least partial READ access on the
thing:

* `ThingsSseRouteBuilder.filterJsonByPartialAccessPaths` injects the allowlist
  into `accessiblePaths` before `JsonPartialAccessFilter.filterJsonByPaths`.
* `AdaptablePartialAccessFilter` applies the same allowlist via a shared
  helper in both `filterAdaptablePayload` (full-thing event payloads) and
  `filterExtraFieldsIntoBuilder` (user-requested extras). The set of paths
  is exposed as `ALWAYS_INCLUDED_THING_FIELDS` so the SSE and connectivity
  filters share a single source of truth.
* `THING_QUERY_COMMAND_RESPONSE_ALLOWLIST` in `ThingCommandEnforcement` is
  widened from `thingId` alone to the same five-field set so `RetrieveThing`
  and the streaming filters stay consistent for any subscriber holding at
  least partial read access.

## What is NOT in the allowlist (intentional)

* `policyId` — reveals which Policy entity gates the thing; remains
  per-path policy-enforced.
* `_metadata` — currently per-path policy-enforced via `Enforcer.buildJsonView`
  in the standard query path (a READ grant on `thing:/attributes` does **not**
  propagate to `thing:/_metadata/attributes/...`). Auto-including it would
  expose data `RetrieveThing` does not, so the existing enforcement is
  retained.

## Affected files

| File | Change |
|------|--------|
| `gateway/service/.../ThingsSseRouteBuilder.java` | Inject allowlist before `JsonPartialAccessFilter` call. |
| `internal/utils/protocol/.../AdaptablePartialAccessFilter.java` | Expose `ALWAYS_INCLUDED_THING_FIELDS`; share `filterJsonByPathsWithAllowlist` helper across payload and extras filters. |
| `things/service/.../ThingCommandEnforcement.java` | Widen `THING_QUERY_COMMAND_RESPONSE_ALLOWLIST` from `thingId` to the five-field identity/audit set. |
| `*Test.java` (three test files) | Flip the buggy `thingId-is-stripped` assertion, add coverage for payload, extras, and `RetrieveThing` allowlist behavior. |

## Tests

* `internal/utils/protocol`: 45/45 (2 new tests covering payload + extras).
* `gateway/service` SSE: 22/22 (1 new test covering audit fields).
* `connectivity/service` `OutboundMappingProcessorTest`: 17/17.
* `things/service` `ThingCommandEnforcementTest`: 22/22 (existing field-order
  expectation updated to follow widened allowlist).

### DIFF
--- a/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/sse/ThingsSseRouteBuilder.java
+++ b/gateway/service/src/main/java/org/eclipse/ditto/gateway/service/endpoints/routes/sse/ThingsSseRouteBuilder.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -90,6 +91,7 @@ import org.eclipse.ditto.internal.models.signalenrichment.SignalEnrichmentFacade
 import org.eclipse.ditto.internal.utils.config.ScopedConfig;
 import org.eclipse.ditto.internal.utils.metrics.DittoMetrics;
 import org.eclipse.ditto.internal.utils.metrics.instruments.counter.Counter;
+import org.eclipse.ditto.internal.utils.protocol.AdaptablePartialAccessFilter;
 import org.eclipse.ditto.internal.utils.protocol.JsonPartialAccessFilter;
 import org.eclipse.ditto.internal.utils.protocol.PartialAccessPathResolver;
 import org.eclipse.ditto.internal.utils.pubsub.StreamingType;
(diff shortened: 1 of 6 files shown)


hu-ahmed (approved):
LGTM!

PR #2476, opened 2026-06-24
Add security hardening guide for operating Ditto

thjaeckle:
## Summary

Adds a new **Operating - Security Hardening** documentation page that collects the configuration and deployment steps required to run Eclipse Ditto securely in production. It frames these as shared-responsibility / deployment concerns (Ditto's security model intentionally delegates parts of the trust boundary to the operator) rather than code defects.

The page covers:

- **Trust boundary** — keep the gateway behind a reverse proxy; never expose it directly when pre-authentication is enabled.
- **Reverse proxy config** — enforce authentication and overwrite/strip the `x-ditto-pre-authenticated` header; replace the sample `nginx.htpasswd` credentials.
- **Network isolation** — Kubernetes `NetworkPolicy` guidance (no auth between services on the Pekko cluster).
- **Authentication & token lifetime** — prefer OIDC; short-lived JWTs; WebSocket session expiry behaviour (no pre-expiry warning — the client must refresh before expiry) and the stateless-revocation caveat.
- **DevOps & status endpoints** — change default credentials, prefer OAuth2, restrict from public ingress, keep secrets out of retrievable config.
- **Error responses** — clarifies that stack traces are never included in API responses.
- **Policy design & TLS.**

Includes a 10-point pre-production hardening checklist, and wires the page into the sidebar under *Operating Details*.

This is documentation only — no code or behaviour changes.

### DIFF
--- a/documentation/src/main/resources/_data/sidebars/ditto_sidebar.yml
+++ b/documentation/src/main/resources/_data/sidebars/ditto_sidebar.yml
@@ -298,6 +298,9 @@ entries:
                   - title: Authentication
                     url: /operating-authentication.html
                     output: web
+                  - title: Security Hardening
+                    url: /operating-security-hardening.html
+                    output: web
                   - title: MongoDB
                     url: /operating-mongodb.html
                     output: web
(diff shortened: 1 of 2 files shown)


alstanchev (approved):
LGTM 👍

PR #2478, opened 2026-06-28
fix swagger ui mount oidc js

kalinkostashki:
- added necessary volumes and volumeMounts


thjaeckle (approved):
Thanks @kalinkostashki - that seemed to have slipped through when adding the file 👍

PR #2479, opened 2026-06-30
fix http gaetway route segments strict matching

alstanchev:
This PR addresses a strange behavior where if for example 
/devops/logging/gaDeway 
is called it is correctly timing out as there is no such service but if we add a letter to the 'logging' segment
/devops/loggingA/gaDeway 
the logging part is matched as a prefix then the rest 'A/gaDeway' is passed along but since the next handle checks for leading slash the segment is dropped and the call is handled as it was 
/devops/logging

This behavior can potentially reach at any place where no strict segment matching is done.

System tests run:
https://github.com/eclipse-ditto/ditto/actions/runs/28428560314


thjaeckle (approved):
Oh, wow .. really strange behavior.

Thanks for catching it and the fix @alstanchev 

Approved 👍

PR #2480, opened 2026-07-02
Cache namespace-filtered policy enforcers to fix CPU hotspot

thjaeckle:
PolicyEnforcer.forNamespace(ns) rebuilt the entire TreeBasedPolicyEnforcer on every enforced signal when a policy had namespace-scoped entries (introduced with #2325 in 3.9.0). A prod JFR profile of ditto-things showed this path (ThingEnforcerActor -> forNamespace -> defaultEvaluator -> createInstance) accounting for ~18% of JVM user CPU.

The base PolicyEnforcer is cached per policy and replaced wholesale on every policy update, so forNamespace(ns) is a pure function of the namespace for a given instance. Memoize its result in a per-instance Caffeine cache so the enforcer tree is built at most once per distinct namespace instead of per signal. The memo is invalidated naturally when the instance is replaced. The "return this" fast-path (no entries filtered) is preserved.

forNamespace is only ever called by the things-service enforcer on the long-lived, provider-cached instances (from PolicyEnforcerCacheLoader), so only those get a bounded cache whose size is operator-configurable via ditto.policies-enforcer-cache.namespace-filtered-enforcer-max-size (default 100, env DITTO_POLICIES_ENFORCER_NAMESPACE_FILTERED_MAX_SIZE). All other construction paths (of/embed/withResolvedImports, filtered children, and the policies-service enforcers) never accumulate in the cache and use an unbounded (always-empty) one, so no hard-coded default size is needed.

Helm values and the policies/things/connectivity deployment templates expose the new setting.


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/28599981076

policies/service/src/main/java/org/eclipse/ditto/policies/service/enforcement/PolicyCommandEnforcement.java
```diff
@@ -70,6 +70,7 @@
 import org.eclipse.ditto.policies.model.signals.commands.query.PolicyQueryCommandResponse;
 import org.eclipse.ditto.policies.model.signals.commands.query.RetrievePolicy;
 import org.eclipse.ditto.policies.model.signals.commands.query.RetrievePolicyResponse;
+import org.jspecify.annotations.NonNull;
```
hu-ahmed: Not used import

policies/service/src/main/java/org/eclipse/ditto/policies/service/enforcement/PolicyCommandEnforcement.java
```diff
@@ -476,27 +481,38 @@ private JsonObject getJsonViewForPolicyQueryCommandResponse(final JsonObject res
...
+                        "could be determined: {}", policySignal);
+                return DittoInternalErrorException.newBuilder()
+                        .dittoHeaders(policySignal.getDittoHeaders())
+                        .build();
+            }
+        }
+    }
+
+    private static CommandToExceptionRegistry<PolicyCommand<?>, DittoRuntimeException>
+    getPolicyCommandDittoRuntimeExceptionCommandToExceptionRegistry(PolicyCommand<?> policyCommand) {
```
hu-ahmed: missing final on the parameter

hu-ahmed (commented):
LGTM!
just small minor stuff

PR #2481, opened 2026-07-02
Cache partial-access-paths enrichment per thing

thjaeckle:
## Problem

Partial-access-paths event enrichment (`ThingEventEnricher` → `PartialAccessPathCalculator`) recomputes, on **every** `ThingEvent` for a thing whose policy grants *restricted* READ to some subject, the set of accessible JSON paths per subject — walking the whole Thing JSON per restricted subject, collapsing to ancestor pointers and building an indexed header (~1.9 MB allocated per event for a 500-leaf thing). On a production `ditto-things` JFR profile this was the largest CPU consumer among the enrichment paths.

## Insight

The result is a pure function of `(policy revision, Thing structure)` — it does **not** depend on field values. IoT workloads are dominated by value-only telemetry updates (same structure, repeated events), so the identical result is recomputed on nearly every event.

## Change

`ThingEventEnricher` is created once per (cluster-sharded, sequentially-processing) `ThingPersistenceActor`, so the computed indexed `JsonObject` is memoized per thing, keyed on `(policyId, policyRevision, structureHash)`:

- `PartialAccessPathCalculator.structureHash(JsonObject)` — a 64-bit FNV-1a hash of the Thing's object-key tree with scalar **values excluded**, mirroring `TreeBasedPolicyEnforcer.collectFlatPointers` leaf semantics (recurse into non-empty objects; scalars, arrays and empty objects are leaves) with descend/ascend/leaf tokens so nesting boundaries cannot collide.
- The last result is held behind a single `volatile` immutable holder (enrichment may complete off the actor thread), so a cache hit always observes a consistent key+result tuple; races only cause a benign recompute. A policy-revision bump or any structural change misses naturally, and the cache is bypassed when the policy revision is unavailable.
- Safety valve: `event.partial-access-events.cache.enabled` (default `true`, env `THING_EVENT_PARTIAL_ACCESS_EVENTS_CACHE_ENABLED`), kept config-only to match the sibling `partial-access-events.enabled` flag.

Behaviour is unchanged — a cache hit returns exactly what a recompute would.

## Impact (JMH, JDK 25)

Repeated value-only events on a 500-leaf thing with one restricted subject; steady-state cache hit vs recompute:

| Metric | recompute | cache hit |
| ------ | --------- | --------- |
| time | 407 µs/op | 30 µs/op (**~13× faster**) |
| allocation | 1.89 MB/op | 0.12 MB/op (**~16× less**) |

## Tests

Added unit tests for `structureHash` (value-independence; add/remove/rename leaf; scalar-vs-object; nesting boundaries; empty object), enricher cache hit/miss-on-structure-change/miss-on-revision-change/toggle-off tests, and an enricher-level JMH benchmark.

### DIFF
--- a/things/service/src/main/java/org/eclipse/ditto/things/service/common/config/DefaultThingEventConfig.java
+++ b/things/service/src/main/java/org/eclipse/ditto/things/service/common/config/DefaultThingEventConfig.java
@@ -31,6 +31,7 @@ public final class DefaultThingEventConfig implements ThingEventConfig {
     private final DefaultEventConfig defaultEventConfigDelegated;
     private final List<PreDefinedExtraFieldsConfig> preDefinedExtraFieldsConfigs;
     private final boolean partialAccessEventsEnabled;
+    private final boolean partialAccessEventsCacheEnabled;
 
     private DefaultThingEventConfig(final DefaultEventConfig delegate, final ScopedConfig config) {
         this.defaultEventConfigDelegated = delegate;
@@ -42,6 +43,8 @@ public final class DefaultThingEventConfig implements ThingEventConfig {
                         .toList();
         partialAccessEventsEnabled =
                 config.getBoolean(ThingEventConfigValue.PARTIAL_ACCESS_EVENTS_ENABLED.getConfigPath());
+        partialAccessEventsCacheEnabled =
+                config.getBoolean(ThingEventConfigValue.PARTIAL_ACCESS_EVENTS_CACHE_ENABLED.getConfigPath());
     }
 
     /**
@@ -72,18 +75,25 @@ public final class DefaultThingEventConfig implements ThingEventConfig {
     }
 
     @Override
+    public boolean isPartialAccessEventsCacheEnabled() {
+        return partialAccessEventsCacheEnabled;
+    }
+
+    @Override
     public boolean equals(final Object o) {
         if (!(o instanceof final DefaultThingEventConfig that)) {
             return false;
         }
         return Objects.equals(defaultEventConfigDelegated, that.defaultEventConfigDelegated) &&
                 Objects.equals(preDefinedExtraFieldsConfigs, that.preDefinedExtraFieldsConfigs) &&
-                partialAccessEventsEnabled == that.partialAccessEventsEnabled;
+                partialAccessEventsEnabled == that.partialAccessEventsEnabled &&
+                partialAccessEventsCacheEnabled == that.partialAccessEventsCacheEnabled;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(defaultEventConfigDelegated, preDefinedExtraFieldsConfigs, partialAccessEventsEnabled);
+        return Objects.hash(defaultEventConfigDelegated, preDefinedExtraFieldsConfigs, partialAccessEventsEnabled,
+                partialAccessEventsCacheEnabled);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


hu-ahmed:
Nice optimization and the concurrency handling is sound, but I found one must-fix: the cache key is incomplete for policies whose grants come from imports or namespace-root policies, and that can leak field-level data to a restricted subject. I confirmed it with a live reproduction (not just reasoning). Details, repro, and a suggested fix below, plus a smaller doc-accuracy point and the things I verified are correct.                            
                                                                                                                                                                                                                                   
  ---                                                                                                                                                                                                                              
  🔴 Must-fix — stale cache leaks a revoked field for imported / namespace-root grants                                                                                                                                             
                                                                                                                                                                                                                                   
  Mechanism. The cache is keyed on (policyId, policyRevision, structureHash), where policyRevision comes from policyEnforcer.getPolicy().getRevision(). For an enforcer built via withResolvedImports / withResolvedImportsAndNamespacePolicies, getPolicy() returns the merged policy — but the merge is self.toBuilder().setAll(resolved).build() (Policy.java:494), which preserves the importing policy's own revision. The revisions of imported and namespace-root policies are not reflected in it.                                                                                                                                                                 
                                                                                                                                                                                                                                   
  When an imported / ns-root policy changes, PolicyEnforcerCache.invalidate (lines 144–168) cascades invalidation to every importer, so the enforcer is correctly rebuilt with the new grants — but the importing policy's revision  never moves (it's a separate entity). So matches() still sees the same policyId + revision + structureHash and serves the stale allowlist indefinitely, until the importing policy's own revision bumps or the thing's structure changes.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  This is exactly what cacheServesStoredResultForSamePolicyRevisionAndStructure asserts as correct behaviour ("same revision, narrower enforcer → reuse old result") — but "same revision, different grants" is reachable in production via imports, so that assertion encodes the leak.                                                                                                                                                                      
                                                                                                                                                                                                                                   
  How the reproduction works (plain language). Set up a restricted subject whose READ on public1 + public2 comes from an imported policy, with the thing owned by the importing policy. Then:                                      
                                                                                                                                                                                                                                   
  1. Warm the cache — subscribe to change events as the restricted subject and push value-only updates; it correctly receives public1 and public2.                                                                                 
  2. Revoke public2 in the imported policy only — leave the importing policy untouched.                                                                                                                                            
  3. Push another value-only update (structure unchanged) — the restricted subject still receives public2, a field it can no longer read. 🔴                                                                                       
  4. Control — change the thing's structure (add an attribute), update again — public2 is now correctly withheld.                                                                                                                  
                                                                                                                                                                                                                                   
  Two comparisons isolate the cache as the cause:                                                                                                                                                                                  
  - Read path vs. event path — a direct GET as the restricted subject correctly hides public2 (it doesn't use this cache), while the change-event stream leaks it.                                                                 
  - Value-only vs. structural update — the leak appears only when structure is unchanged and disappears the moment structureHash changes; that hash is the only key component that differs between the two cases.                  
                                                                                                                                                                                                                                   
  Each observation used a fresh event-stream subscription, ruling out the gateway's per-connection caching — what's stale is the things-side header this PR adds. Throughout, the importing policy's _revision stayed 1 while the effective grant changed. The same happens with namespace-root policies (operator trust boundary), and the cache ships enabled by default.                                                                                        
                                                                                                                                                                                                                                   
   | Observation channel | Value-only update | Structural change |                                                                                                                                                                  
  |---|---|---|                                                                                                                                                                                                                    
  | Direct `GET` (no enricher cache) | `public2` hidden ✅ | `public2` hidden ✅   |                                                                                                                                                 
  | Change-event stream (cached header) | **`public2` leaks** 🔴 | `public2` hidden ✅ |

thjaeckle:
@hu-ahmed indeed, very good catch - I often forget about the imports 🙄

thjaeckle:
@hu-ahmed fixed your identified gap - which even simplified the approach .. it now checks if the PolicyEnforcer changed (with object identity, so very cheap OP). This then hooks into the already established logic of when a PolicyEnforcer is rebuilt, including changing policy imports.

hu-ahmed (approved):
LGTM!
Thanks for the improvement

thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/28642803494

PR #2484, opened 2026-07-07
Memoize per-resource READ subject classification in PolicyEnforcer (per-event enforcement CPU hotspot)

thjaeckle:
## What

Adds a per-resource memo of `Enforcer.classifySubjects(resource, READ)` on `PolicyEnforcer`, removing a full policy-tree walk that ran on **every** emitted `ThingEvent`.

## Why — analysis

Profiling `ditto-things` on production (JFR) showed that, once the recent `PolicyEnforcer.forNamespace(...)` caching fix removed the per-signal enforcer **tree rebuild**, the per-event **read-subject evaluation** became the dominant things-service enforcement CPU consumer. Two hot paths re-walk the full `TreeBasedPolicyEnforcer` on every event:

- `ThingCommandEnforcement.addEffectedReadSubjectsToThingSignal` computes the `readGrantedSubjects` for every authorized signal (used for pub/sub routing). With `partial-access-events` enabled (the default) it walks the policy tree per event via `getSubjectsWithUnrestrictedPermission(eventResource)` and `getSubjectsWithPartialPermission(root)`.
- `ReadGrantCollector.collectSubjectsForPointer` calls `classifySubjects(...)` per pointer, per event.

`classifySubjects(resource, READ)` is a **pure function of the resolved policy grants and the resource path** — it does not depend on the Thing's field values. IoT workloads are dominated by value-only telemetry updates that repeatedly touch the same resource paths, so the identical classification was recomputed on every event.

## What changed

- **`PolicyEnforcer`**: generalize the existing single-slot root-read-classification memo into a **bounded per-`ResourceKey` cache**, exposed as `classifyReadSubjects(ResourceKey)`; `getRootResourceReadClassification()` now delegates to it. The memo lives on the `PolicyEnforcer` instance, which `CachingPolicyEnforcerProvider` replaces wholesale on any grant change (own policy, an imported policy, or a namespace-root cascade), so it is invalidated naturally — the same lifetime and invalidation story as the `forNamespace` memo. The bound is propagated to `forNamespace` filtered children, since that is the instance the things hot path enforces against.
- Route `addEffectedReadSubjectsToThingSignal` and `ReadGrantCollector.collectSubjectsForPointer` through the memo. The resulting `readGrantedSubjects` set is **identical** to the previous formula, because at the root `getSubjectsWithPartialPermission(root, READ) == partialOnly ∪ unrestricted` (see `ClassifySubjectsVisitor`).
- New config `ditto.policies-enforcer-cache.read-classification-cache-max-size` (default `1000`), plumbed through `PolicyEnforcerCacheLoader` and the Helm chart (things/policies/connectivity).

## Correctness

`classifyReadSubjects` returns the same `SubjectClassification` as a direct `Enforcer.classifySubjects(resource, READ)`; a cache miss/eviction simply recomputes an equal result, never a wrong one. Behaviour preservation is covered by unit tests asserting the emitted `readGrantedSubjects` are identical to the pre-change formula, both with and without partial-read subjects, plus tests for the memo itself (equality with the direct call, memoization, correctness under a bounded cache).

## Performance — JMH

`ClassifyReadSubjectsBenchmark` (added), `AverageTime`, µs/op:

| Benchmark                                    | Score (µs/op) | Note                                   |
| -------------------------------------------- | ------------- | -------------------------------------- |
| `baselineRecompute` (previous behaviour)     | 2.045 ± 0.108 | full policy-tree walk per call         |
| `memoizedHit` (repeated resource, prod shape)| 0.025 ± 0.002 | ~80× faster                            |
| `memoizedMiss` (no-repeat + tiny cache)      | 2.604 ± 0.910 | within noise of baseline — no regression|
| `endToEndMemoized` (full add-read-subjects)  | 1.311 ± 0.089 | now dominated by header building       |

The repeated-resource case is the production shape (value-only telemetry re-touching the same paths); the pathological no-repeat case (64 distinct keys against a size-4 cache) shows the memo does not regress the worst case.

## Memory

The memo adds a small, **bounded** retained set — `SubjectClassification` is a set of authorization-subject ids, capped at the configured bound per enforcer (and propagated to `forNamespace` children so it cannot grow unbounded). In exchange it removes the per-event transient allocation of the tree walk (temporary sets and visitor state), reducing young-gen allocation churn. The bound is operator-tunable.


thjaeckle:
System tests passed: https://github.com/eclipse-ditto/ditto/actions/runs/28870858118

hu-ahmed (approved):
LGTM!

PR #2485, opened 2026-07-09
Pre-serialize pub/sub signals once across fan-out (flag-gated, default off)

thjaeckle:
## What

Introduce an opt-in optimization that serializes a published signal **once** and reuses those bytes across all
remote fan-out destinations, instead of letting Pekko Artery re-serialize the identical payload once per
destination. Flag-gated behind `ditto.pubsub.pre-serialize-fanout-enabled` and **disabled by default**.

## Why

Under high fan-out load, cluster-internal serialization of published signals is a dominant CPU cost on the
`things` service (a top hotspot, ~25% of CPU, in production profiles). When a single published signal fans out to
N subscriber nodes, Artery serializes the identical payload N times.

## How

- New `PreSerializedPublishSignal` transport envelope carrying a shared, lazily-computed `SignalBytesHolder`. The
  first Artery encoder thread serializes the payload once; every subsequent remote destination reuses the cached
  bytes. Only the small per-destination routing information differs between envelopes.
- A dedicated `PreSerializedPublishSignalSerializer` (own serializer id) reconstructs a plain `PublishSignal` on
  the receiving side, so the `Subscriber` actor is unchanged.
- `Publisher` only uses the envelope when the fan-out reaches at least one **remote** node; purely node-local
  fan-out keeps using the in-memory signal (nothing is serialized there).
- Exposed via Helm on the publishing services (`things`, `policies`, `connectivity`). Subscriber-only services
  (`gateway`, `thingssearch`) do not read the flag but must run this version so they can *deserialize* incoming
  envelopes.

## Safety / rollout

The envelope relies on a new serializer; a node that does not have it registered would drop the fan-out message.
It is therefore only safe to enable **after the whole cluster runs this version**. A two-phase rollout is
documented in the (draft) 3.9.4 release notes: upgrade the entire fleet with the flag off, then enable it on the
publishing services. Because the default is off, upgrading is a no-op until the flag is explicitly turned on.

## Robustness

- All wire length prefixes are validated against the remaining bytes before allocating, so a truncated / corrupt /
  misrouted frame fails cleanly instead of raising `OutOfMemoryError` / `NegativeArraySizeException` on the Artery
  inbound thread.
- An unknown inner manifest (e.g. a newer signal type reaching an older node) propagates the inner serializer's
  `NotSerializableException` — consistent with the regular `PublishSignal` path — instead of throwing on the
  decoder thread.
- Strings are UTF-8-encoded exactly once on the serialization path.

## Validation

- **The full system-test suite passes with the flag enabled** — delivery correctness and cluster stability
  verified under real multi-node, multi-connection load (in addition to the default flag-off run).
- **JMH benchmark** (`SignalFanoutSerializationJmh`): the pre-serialized path stays flat as fan-out grows and is
  up to ~11× faster than per-destination serialization at higher fan-out (e.g. fan-out 10: ~23 µs vs ~263 µs per
  publication), with lower heap allocation. Deserialization cost is unchanged.

## Tests

Serializer round-trip (both entry points), holder thread-safety / serialize-once, serialization-extension
registration, end-to-end fan-out across two clustered actor systems, and malformed / truncated / unknown-manifest
failure cases.


thjaeckle:
System tests:
* With the optimisation **OFF** (default): https://github.com/eclipse-ditto/ditto/actions/runs/29007782302 ✅

thjaeckle:
System tests:
* With the optimisation **ON**: https://github.com/eclipse-ditto/ditto/actions/runs/29015024344 ✅

alstanchev (approved):

thjaeckle:
@alstanchev this has the potential to greatly improve throughput and reduce CPU resources when you have multiple subscribers of events (e.g. connections, websockets, SSEs) running in the cluster.  
However activation of it - as already added to the draft release notes - has to be carefully planned:  
It's a 2-step thing, first deploy 3.9.4 and afterwards activate the flag.

PR #2492, opened 2026-07-17
llms.txt release note fixing

kalinkostashki:
- added a check to make sure the latest release notes is appended in llms.txt
- added a check to compare llms.txt links actually exist -> not matching page in llms.txt forces error
- also checks for missing pages in llms.txt but logs warnings as some things might be omitted by design

### DIFF
--- a/documentation/src/main/resources/_plugins/llms_markdown_generator.rb
+++ b/documentation/src/main/resources/_plugins/llms_markdown_generator.rb
@@ -11,6 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+require 'set'
+
 # Jekyll plugin that generates clean .html.md versions of documentation pages
 # for LLM consumption, supporting the llms.txt specification (https://llmstxt.org).
 #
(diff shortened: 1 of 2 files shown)


kalinkostashki:
@thjaeckle 
If you run llms_markdown_generator.rb now you will get some protocol-* specific html files that are missing. 
Don't know if those have been omitted by design or it's something caught by the validation now.

thjaeckle:
> If you run llms_markdown_generator.rb now you will get some protocol-* specific html files that are missing.

@kalinkostashki no, this should not be by design.  
But this is not fixed by this PR, right?

kalinkostashki:
> > If you run llms_markdown_generator.rb now you will get some protocol-* specific html files that are missing.
> 
> @kalinkostashki no, this should not be by design. But this is not fixed by this PR, right?

This PR would just give you a warning that something is missing and error if there's a mismatch between llms.txt and a different name for a html.md file(or not existing)

thjaeckle (approved):
@kalinkostashki ok, cool - good thing to have this insight :)

PR #2493, opened 2026-07-19
fix: include resolved IP in blocked host error messages

### LINKED ISSUE #681: Error messages for connection domain blacklists are confusing
When creating an mqtt-connector:

In case the domain of e.g. an mqtt service for a connector is not only publicly available but also via an alias within a docker network directly reachable through the local network the connector will not be created and an error will be thrown that the domain `mqtt.public-and-valid.org` is blacklisted. This happens, because the domain is internally translated to local addresses, which is indeed an undesired target from the point of view of the implementation. The problem however is, that the error message does not provide any hint about this being the case and instead complains that the valid domain is blacklisted which is not the case.

The error messages in this case need to give more details about what is happening. This has to become more agnostic to the underlying networking behavior.

amanjain57-gif:
## Summary
  
  Fixes #681
  
  When a valid hostname resolves to a blocked address (loopback, site-local, multicast, or blocked subnet), the error
  message now includes the resolved IP address that triggered the block. This helps users understand why a publicly
  valid domain is being rejected — for example, when a domain resolves to a local Docker network alias internally.
  
  ## Changes
  
  - `DefaultHostValidator`: Each blocked-address check now includes the resolved IP in the error message (e.g., "the
  hostname resolved to a loopback address (127.0.0.1)")
  - `HostValidationResult`: Updated exception description to mention allowed hostnames configuration as a remediation
  path
  - Updated test assertion to match the new message format
  
  ## Before/After
  
  **Before:** `The configured host 'mqtt.example.org' may not be used for the connection because the host is blocked.`
  
  **After:** `The configured host 'mqtt.example.org' may not be used for the connection because the hostname resolved to
  a site local address (192.168.1.5).`
  
  Signed-off-by: Aman Jain <aman.jain57@gmail.com>

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/DefaultHostValidator.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/DefaultHostValidator.java
@@ -104,22 +104,28 @@ final class DefaultHostValidator implements HostValidator {
         try {
             final InetAddress[] inetAddresses = resolver.resolve(host);
             for (final InetAddress requestAddress : inetAddresses) {
+                final String resolvedAddress = requestAddress.getHostAddress();
                 if (requestAddress.isLoopbackAddress()) {
-                    return HostValidationResult.blocked(host, "the hostname resolved to a loopback address.");
+                    return HostValidationResult.blocked(host,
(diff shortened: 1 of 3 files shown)


thjaeckle (approved):
@amanjain57-gif thanks for the contribution.  
Did you run into that issue yourself or did you just want to tackle existing issues, providing a contribution?
Looks good to me either way 👍

PR #2494, opened 2026-07-21
Memoize JSON header-value parsing and skip re-validation of trusted DittoHeaders

thjaeckle:
## What

`DittoHeaders` stores every header value as a `String` and re-parses the JSON-typed values (e.g. `authorization-context`, `read-subjects`, `expected-response-types`) from text every time they are validated, serialized or accessed. Since `DittoHeaders` is immutable, parsing a value is a pure function of that string, so most of this work is redundant. This PR removes it in two places.

## Why

Profiling the Things service in production (10-minute JFR) showed roughly **11% of CPU** spent parsing header values through the JSON layer — split about evenly between build-time value-type validation and `toJson()`/accessor calls — and dominated by `DittoHeaders`. The dominant callers were `JsonObjectValueValidator.tryToParseJsonObject`, `AbstractDittoHeaders.toJson` and `getJsonArrayForDefinition`. This cost scales linearly with signal throughput (every signal carries headers), so it grows with load.

## Changes

- **Memoize the parsed value on `Header`.** `Header` now lazily caches the parsed `JsonValue` (idempotent, unsynchronized — the same benign-race pattern as `String.hashCode()`). `toJson()`, `getAuthorizationContext()` and `getJsonArrayForDefinition()` reuse it instead of re-parsing. Because the header map is shared across unmutated `DittoHeaders` copies, a value is parsed at most once per distinct `Header`, no matter how often the instance is serialized or read.
- **Skip re-validation of trusted headers on deserialization.** Cluster-internal deserialization rebuilt headers through the validating builder, re-parsing every JSON-typed value on each hop even though the sending node already validated them when the headers were first created. New `DittoHeaders.newFromTrustedJson(JsonObject)` (and `DefaultDittoHeadersBuilder.ofTrusted(...)`) build headers without re-running value-type validation, applying the same trust reasoning as the existing `DittoHeaders`-based builder constructor. `AbstractJsonifiableWithDittoHeadersSerializer` now uses it. External/untrusted ingress continues to validate as before.

Both changes are independent of the underlying JSON parser.

## Benchmark

New JMH benchmark `HeaderValueParsingBenchmark` in `base/model` (JDK 25, `AverageTime`, ns/op):

| Operation                                   | before  | after  | speedup |
|---------------------------------------------|---------|--------|---------|
| `getAuthorizationContext`                   | 703     | 93     | ~7.5×   |
| `getReadGrantedSubjects`                    | 411     | 99     | ~4.2×   |
| `toJson`                                    | ~1300   | 199    | ~6×     |
| parse → cache-hit (per JSON header set)     | 1087    | 3.5    | ~310×   |
| header deserialization (build from wire JSON) | 1584  | 594    | ~2.7×   |

## Notes

- New public API `DittoHeaders.newFromTrustedJson(JsonObject)` is tagged `@since 3.9.5`.
- No behaviour change. `mvn verify` is green on `base/model` (743 tests, japicmp API-compatibility clean) and `internal/utils/cluster` (73 tests, including serializer round-trips).

### DIFF
--- a/base/model/pom.xml
+++ b/base/model/pom.xml
@@ -51,10 +51,34 @@
             <artifactId>pekko-testkit_${scala.version}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths combine.children="append">
+                        <path>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
+                            <version>${jmh.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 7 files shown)


thjaeckle:
System tests are green: https://github.com/eclipse-ditto/ditto/actions/runs/29817176204

hu-ahmed (approved):
LGTM!
Thanks for the improvement

PR #2495, opened 2026-07-21
Require all doc pages in llms.txt, exclude only protocol-examples

thjaeckle:
## What

The `llms.txt` curated index only emitted an informational warning for doc
pages it didn't link, so 103 pages were silently missing from the index —
12 real content pages plus the auto-generated `protocol-examples-*` family.

This makes being listed the default:

- Adds the 12 missing content pages to `llms.txt`:
  - `glossary`
  - `operating-authentication`, `operating-configuration`, `operating-devops`,
    `operating-mongodb`, `operating-monitoring`, `operating-security-hardening`
  - `connectivity-response-diversion`
  - `protocol-specification-connections`
  - `sandbox`, `presentations`, `feedback`
- Escalates the "unlisted page" check in `llms_markdown_generator.rb` from a
  warning to a build failure.

Only two families remain permitted omissions:

- The auto-generated `protocol-examples-*` children (the curated
  `protocol-examples` overview stands in for the family).
- Historical `release_notes_*` pages, where only the latest is ever linked —
  already enforced separately by the existing `validate` step.

As a result the index can no longer silently drift out of sync: a newly added
doc page that isn't listed (and isn't one of the excluded families) now fails
the Jekyll build instead of printing an ignorable notice.

### DIFF
--- a/documentation/src/main/resources/_plugins/llms_markdown_generator.rb
+++ b/documentation/src/main/resources/_plugins/llms_markdown_generator.rb
@@ -116,7 +116,18 @@ end
 # shipping a stale link or description.
 module LlmsTxtIndex
   RELEASE_NOTES_PREFIX = 'release_notes_'
+  # Per-command protocol example pages (e.g. protocol-examples-creatething.html)
+  # are auto-generated boilerplate and intentionally not enumerated in llms.txt -
+  # the single curated "protocol-examples.html" overview stands in for the family.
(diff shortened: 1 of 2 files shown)


thjaeckle:
@kalinkostashki FYI - this fixes most of the new warnings - just the `protocol-example-*` are excluded and just a single curated `protocol-examples` overview is kept.

PR #2496, opened 2026-07-21
Fix ClusterSingletonManagerIsStuck causing status-route bind restarts on rolling updates

thjaeckle:
On every rolling update, service pods (first observed on connectivity) restarted 1-3x with `Bind failed ... Address already in use` on `<podIP>:8080` (the HTTP status route).

## Root cause

When a newly-oldest node cannot hand the cluster singleton over from the previous oldest before its retry budget is exhausted, the `ClusterSingletonManager` throws `ClusterSingletonManagerIsStuck`. During a rolling update the whole Ditto cluster (all services share one Pekko cluster) churns, and gossip dissemination of the previous oldest's `Leaving -> Exiting -> Removed` transition can take longer than the Pekko default budget of 15 retries (1s each). Prod logs showed the new oldest throwing stuck ~8s *after* the previous oldest had already been removed from the cluster.

`DittoRootActor`'s supervision decider escalated the exception, which restarted the root actor, which re-ran its constructor and re-bound the status route on the still-bound port -> `BindException` -> `system.terminate()` -> pod restart. This repeated until the old oldest was fully gone. Affects all services that extend `DittoRootActor`.

## Changes

- **`DittoRootActor` decider**: handle `ClusterSingletonManagerIsStuck` by `restart()`-ing the stuck singleton child (the recovery Pekko intends: start from a clean state and take over once the previous oldest is removed) instead of escalating and restarting the whole root actor.
- **`bindHttpStatusRoute`**: tolerate a `BindException` (port already held, e.g. by a prior binding that survived an actor restart) by logging a warning and keeping the actor system alive; only `system.terminate()` on genuinely fatal bind failures.
- **`pekko.cluster.singleton.min-number-of-hand-over-retries`**: raised from the Pekko default of 15 to 30 (env `PEKKO_CLUSTER_SINGLETON_MIN_NUMBER_OF_HAND_OVER_RETRIES`) so the new oldest waits for the previous oldest's removal to be gossiped and then takes over cleanly, avoiding the stuck exception in the first place. It is only a ceiling — hand-over still completes as soon as the previous oldest confirms or is removed.
- **Helm**: expose the setting as `global.cluster.singletonMinNumberOfHandOverRetries`, wired into all five service deployments; bump chart version.
- **Test**: `AbstractDittoRootActorTest` now asserts a stuck singleton restarts the child rather than terminating the root actor (runs for all five service root actors).

### DIFF
--- a/base/service/src/main/java/org/eclipse/ditto/base/service/actors/DittoRootActor.java
+++ b/base/service/src/main/java/org/eclipse/ditto/base/service/actors/DittoRootActor.java
@@ -17,6 +17,7 @@ import static org.apache.pekko.http.javadsl.server.Directives.logResult;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.net.BindException;
 import java.net.ConnectException;
 import java.util.NoSuchElementException;
 
@@ -39,6 +40,7 @@ import org.apache.pekko.actor.Props;
 import org.apache.pekko.actor.Status;
 import org.apache.pekko.actor.SupervisorStrategy;
 import org.apache.pekko.cluster.Cluster;
+import org.apache.pekko.cluster.singleton.ClusterSingletonManagerIsStuck;
 import org.apache.pekko.http.javadsl.Http;
 import org.apache.pekko.http.javadsl.server.Route;
 import org.apache.pekko.japi.pf.DeciderBuilder;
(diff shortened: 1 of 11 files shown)


hu-ahmed (approved):
LGTM!

PR #2499, opened 2026-07-23
Fix blank resource_type metric tag crashing cluster serialization

thjaeckle:
## Problem

Serializing certain signals over Pekko Artery throws on the encoder thread:

```
java.lang.IllegalArgumentException: The value must not be blank.
	at org.eclipse.ditto.internal.utils.metrics.instruments.tag.Tag.validateString(Tag.java:55)
	at org.eclipse.ditto.internal.utils.metrics.instruments.tag.Tag.of(Tag.java:51)
	...
	at org.eclipse.ditto.internal.utils.cluster.AbstractJsonifiableWithDittoHeadersSerializer.incrementCounter(...:176)
	at org.eclipse.ditto.internal.utils.cluster.AbstractJsonifiableWithDittoHeadersSerializer.toBinary(...:217)
	at org.apache.pekko.remote.MessageSerializer$.serializeForArtery(MessageSerializer.scala:97)
```

## Cause

Since `6d7e225130` ("Categorize cluster (de)serialization metric by category + resource type"), the `<serializer>_serializer_messages` counter is tagged with the signal's `resource_type`, derived from `WithResource#getResourceType()`.

Health signals (`RetrieveHealth`, `RetrieveHealthResponse`, `ResetHealthEvents`, `ResetHealthEventsResponse`) intentionally return a **blank** string (they have no resource type). `Tag.of(key, value)` rejects blank values, so building the tagged counter throws. Because the counter is built lazily via `computeIfAbsent`, the failure prevents caching and recurs for every such health message on the remoting hot path.

## Fix

In `AbstractJsonifiableWithDittoHeadersSerializer#resourceTypeOf`, fall back to the existing `"other"` bucket when the resource type is `null` or blank, so the metric tag value is always valid.

A regression test in `SharedJsonifiableSerializerTest.CategoryAndResourceTypeTagTest` round-trips a `RetrieveHealth` and asserts it is tagged `category=command, resource_type=other` (this test reproduces the exception before the fix).

### DIFF
--- a/internal/utils/cluster/src/main/java/org/eclipse/ditto/internal/utils/cluster/AbstractJsonifiableWithDittoHeadersSerializer.java
+++ b/internal/utils/cluster/src/main/java/org/eclipse/ditto/internal/utils/cluster/AbstractJsonifiableWithDittoHeadersSerializer.java
@@ -184,7 +184,10 @@ public abstract class AbstractJsonifiableWithDittoHeadersSerializer extends Seri
     private static String resourceTypeOf(final Object object) {
         final String result;
         if (object instanceof WithResource withResource) {
-            result = withResource.getResourceType();
+            final String resourceType = withResource.getResourceType();
+            // some signals (e.g. health signals) have no resource type and return a blank string; the metric tag value
+            // must not be blank, so fall back to the "other" bucket in that case:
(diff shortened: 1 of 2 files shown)


hu-ahmed (approved):
LGTM!

PR #2500, opened 2026-07-24
Make gateway devops/status passwords optional under oauth2 auth

thjaeckle:
## Problem

The gateway devops/status password `Secret` (`gateway-secret.yaml`) is always rendered, and when no
explicit password is configured the `randAlphaNum` fallback generates a **new** value on every template
render. Under `oauth2` (JWT) devops authentication the password is never consumed, yet this churns the
`Secret` on each render — which restarts the gateway on every ArgoCD sync.

## Changes

- **`gateway-secret.yaml`**: the `Secret` — and each individual `devops-password` / `status-password`
  key — is now only materialised when it is actually consumed, i.e. when the respective
  `devops.authMethod` / `devops.statusAuthMethod` is `basic`, or an explicit password is provided.
  Under `oauth2` no password is generated, so there is nothing left to churn.
- **`gateway-deployment.yaml`**: the `devops-password` / `status-password` env `secretKeyRef`s are marked
  `optional: true`, so the gateway starts cleanly when the Secret/key is absent under `oauth2`.
- **Helm chart**: bumped `version` to `4.5.0` and documented the change in the chart `CHANGELOG.md`
  (`Changed` + `Fixed`).

No behaviour change for `basic` devops auth or for explicitly-configured passwords.

### DIFF
--- a/deployment/helm/ditto/CHANGELOG.md
+++ b/deployment/helm/ditto/CHANGELOG.md
@@ -15,6 +15,19 @@ sections: `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`.
 
 ## [Unreleased]
 
+## [4.5.0]
+
+### Changed
+- The gateway devops and status `Secret` passwords are now only materialised when they are actually consumed,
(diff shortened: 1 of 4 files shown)


hu-ahmed (approved):
LGTM!

PR #2507, opened 2026-07-28
Avoid re-validating already-validated DittoHeaders on traced hops and journal recovery

thjaeckle:
## Problem

When distributed tracing is enabled (and especially at high trace sampling rates), several hot paths rebuild `DittoHeaders` through the **validating** constructor even though the source headers were already validated. Concretely, injecting span context via

```java
DittoHeaders.of(startedSpan.propagateContext(dittoHeaders))
```

goes through `DittoHeaders.of(Map)`, whose fast-path only applies when the argument is already a `DittoHeaders`. With an active span, `StartedSpan.propagateContext(Map)` returns a plain `HashMap` (the base headers plus the injected W3C trace context), so `DittoHeaders.of(...)` rebuilds through the validating constructor and **re-parses/re-validates every JSON-typed header value on each hop** — undoing the header-value memoization and trusted-deserialization optimizations introduced in 3.9.5. This shape is repeated at ~20 call sites across the services.

The same unnecessary re-validation happens on **event-journal recovery**: `AbstractMongoEventAdapter.fromJournal` and `AbstractPersistenceActor.mapJournalEntryToEvent` rebuild the persisted headers via `DittoHeaders.newBuilder(obj).build()`, re-parsing every JSON-typed header value even though those headers were validated when the event was originally persisted.

## Changes

1. **Add a trusted `StartedSpan.propagateContext(DittoHeaders)` overload.** It injects only the trace-context delta (`propagateContext(Map.of())`) onto the already-validated headers via `dittoHeaders.toBuilder()` (which uses the skip-validation constructor), so only the small trace-context delta is validated and the memoized parsed values are preserved. Existing call sites that pass a `DittoHeaders` now bind to this overload and skip the re-validation automatically; callers passing a plain `Map` (e.g. raw transport/HTTP headers) keep the existing `Map` overload. The cluster (de)serializer uses the overload directly.

2. **Deserialize persisted headers as trusted on journal recovery** via `DittoHeaders.newFromTrustedJson(...)` in the two journal-recovery paths.

## Notes

- Overload resolution is by static type, so this is source-compatible: only `DittoHeaders` arguments switch behavior, and none of those call sites mutate the returned map (transport consumers and the HTTP request-tracing directive pass plain `Map`s and are unaffected).
- Builds on the 3.9.5 header-value memoization / `newFromTrustedJson` work.

### DIFF
--- a/internal/utils/cluster/src/main/java/org/eclipse/ditto/internal/utils/cluster/AbstractJsonifiableWithDittoHeadersSerializer.java
+++ b/internal/utils/cluster/src/main/java/org/eclipse/ditto/internal/utils/cluster/AbstractJsonifiableWithDittoHeadersSerializer.java
@@ -269,8 +269,7 @@ public abstract class AbstractJsonifiableWithDittoHeadersSerializer extends Seri
             final DittoHeaders dittoHeaders,
             final StartedSpan startedSpan
     ) {
-        final var dittoHeadersWithSpanContext = DittoHeaders.of(startedSpan.propagateContext(dittoHeaders));
-        return dittoHeadersWithSpanContext.toJson();
+        return startedSpan.propagateContext(dittoHeaders).toJson();
     }
 
     @SuppressWarnings("java:S3740")
@@ -389,7 +388,7 @@ public abstract class AbstractJsonifiableWithDittoHeadersSerializer extends Seri
                 beforeDeserializeInstant
         );
         final var result =
-                deserializeJson(payload, manifest, DittoHeaders.of(startedSpan.propagateContext(dittoHeaders)));
+                deserializeJson(payload, manifest, startedSpan.propagateContext(dittoHeaders));
         try {
             return result;
         } finally {
--- a/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/AbstractMongoEventAdapter.java
+++ b/internal/utils/persistence/src/main/java/org/eclipse/ditto/internal/utils/persistence/mongo/AbstractMongoEventAdapter.java
@@ -96,7 +96,9 @@ public abstract class AbstractMongoEventAdapter<T extends Event<?>> implements E
                 final JsonObject jsonObject = jsonValue.asObject()
                         .setValue(EventsourcedEvent.JsonFields.REVISION.getPointer(), Event.DEFAULT_REVISION);
                 final DittoHeaders dittoHeaders = jsonObject.getValue(HISTORICAL_EVENT_HEADERS)
-                        .map(obj -> DittoHeaders.newBuilder(obj).build())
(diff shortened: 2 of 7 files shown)


hu-ahmed (commented):
`ditto.tracing.enabled = false` is the shipped default, so a stock deployment hits `EmptyStartedSpan`, whose `propagateContext(Map)` returns the argument unchanged.

- **Before:** `DittoHeaders.of(sameInstance)` hit the `instanceof DittoHeaders` fast path — a pointer return, zero work.
- **After:** `toBuilder()` is called *unconditionally* at `StartedSpan:132`, before the diff loop. So even when the loop finds zero changes we pay two `LinkedHashMap` copies (`preserveCaseSensitivity`, then `ImmutableDittoHeaders.fromBuilder`), the `entrySet()` copy that `Map#forEach` walks, and a per-entry `Optional`-allocating `get()`. Nothing in the method can return early.

Measured on a 10-header set with three JSON-typed values (crude wall-clock, 2M iterations — read the gaps, not the absolute figures):

| path | before | after |
|---|---|---|
| **no-op span (tracing off — default)** | **9.1 ns/op** | **785.3 ns/op** |
| traced span (tracing on) | 2659.5 ns/op | 1155.3 ns/op |

So tracing-on gains ~1500 ns/hop (~2.3×, the premise is sound) while tracing-off loses ~776 ns/hop across the ~18 `DittoHeaders.of(startedSpan.propagateContext(dittoHeaders))` sites. Sub-20 µs per request, so not a latency incident — but it is a performance PR making the default config slower, and it is avoidable.

Build only when something actually changed:

```java
default DittoHeaders propagateContext(final DittoHeaders dittoHeaders) {
    final Map<String, String> propagatedHeaders = propagateContext((Map<String, String>) dittoHeaders);
    DittoHeadersBuilder<?, ?> builder = null;
    for (final var entry : propagatedHeaders.entrySet()) {
        if (!Objects.equals(entry.getValue(), dittoHeaders.get(entry.getKey()))) {
            if (null == builder) {
                builder = dittoHeaders.toBuilder();
            }
            builder.putHeader(entry.getKey(), entry.getValue());
        }
    }
    return null == builder ? dittoHeaders : builder.build();
}
```

`Objects.equals` also removes an NPE if a propagation writer ever emits a `null` value — `KamonHttpContextPropagation` writes into a `HashMap` via `result::put`, which accepts nulls, and `value.equals(...)` would blow up in the lambda instead of hitting `putHeader`'s `checkNotNull(value, "value")`. Worth an explicit `EmptyStartedSpan` override returning `dittoHeaders` too, so the no-op path is obvious and free.

thjaeckle:
Great catch @hu-ahmed - indeed a performance improvement PR should not make performance worse for the default case (tracing off).  
I addressed your findings 👍

hu-ahmed (approved):
LGTM! 
Thanks for the improvement

PR #2510, opened 2026-07-29
Make operator metrics MongoDB read preference and concern configurable

### LINKED ISSUE #2163: Read concern for "aggregation" metrics is same as configured search read concern
I noticed that for the new [aggregation based metrics](https://eclipse.dev/ditto/installation-operating.html#operator-defined-custom-aggregation-based-metrics) the used MongoDB `readConcern` is the same as the configured one for all searches in Ditto.

This can be unwanted, e.g. when having configured a really "costly" readConcern like `linearizable` in order to guarantee a strong consistency of search results.

For aggregations, the readConcern should be `local` IMO - the cheapest possible.

Maybe making it configurable is the best way to fix that.

thjaeckle:
## What

Makes the MongoDB `readPreference` and `readConcern` used by the operator metrics configurable **independently** from the read settings used for normal searches (`query.persistence`).

Two optional, independent config blocks are added under `ditto.things-search.operator-metrics`:

| Block                                     | Applies to                                     |
|-------------------------------------------|------------------------------------------------|
| `custom-metrics-persistence`              | the count based `custom-metrics`               |
| `custom-aggregation-metrics-persistence`  | the `$group` `custom-aggregation-metrics`      |

Both are optional and **fall back to the general `query.persistence` config when absent**, so existing deployments are unaffected.

## Why

Closes #2163.

The operator (aggregation) metrics so far inherited the global search `readConcern` / `readPreference`. That is unwanted when searches are configured with an expensive `readConcern` (e.g. `linearizable`) for strong consistency: the periodic operator metric queries would rather run cheaply (`readConcern = local`) and/or be offloaded to a secondary node (`readPreference = secondaryPreferred`) without weakening the user-facing search guarantees.

## How

- **Aggregation metrics** use a dedicated `MongoThingsAggregationPersistence`, so it resolves the dedicated `custom-aggregation-metrics-persistence` config (else `query.persistence`) and binds read preference + read concern at construction.
- **Count metrics** share the `SearchActor` → `MongoThingsSearchPersistence.count/sudoCount` path with user-facing counts, so `SudoCountThings` gains optional `readPreference` / `readConcern` fields (`@since 3.9.7`, mirroring the existing `indexHint`), applied per-query via `collection.withReadPreference(...).withReadConcern(...)`. Unknown values are logged and ignored, not fatal.
- Helm: both blocks are rendered only when configured; chart version bumped to `4.6.1`.

## Out of scope

The `hint` sub-point from #2163 is intentionally left out: aggregation metrics already support a per-metric `index-hint`, so hints can already differ per metric. Decoupling the shared namespace-hint fallback can be tracked separately.

### DIFF
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -1729,6 +1729,19 @@ thingsSearch:
       #   # or as index key spec:
       #   # indexHint:
       #   #   "t.attributes.location": 1
+      # customMetricsPersistence optionally overrides the MongoDB read preference / read concern used for the
+      # count based "customMetrics" queries above. When unset (the default), the general query.persistence
+      # settings are used - so this can offload the periodic operator metric counts to a secondary node without
+      # affecting user-facing search/count requests.
+      # customMetricsPersistence:
+      #   readPreference: secondaryPreferred
+      #   readConcern: default
+      # customAggregationMetricsPersistence optionally overrides the MongoDB read preference / read concern used
+      # for the "customAggregationMetrics" ($group) queries above. When unset (the default), the general
+      # query.persistence settings are used. Can be configured independently from customMetricsPersistence.
+      # customAggregationMetricsPersistence:
+      #   readPreference: secondaryPreferred
+      #   readConcern: default
     # dispatchers contains tuning for the things-search service's Pekko dispatchers.
     # Defaults below match the in-jar HOCON defaults and only need overriding under sustained load.
     dispatchers:
--- a/deployment/helm/ditto/values.yaml
+++ b/deployment/helm/ditto/values.yaml
@@ -1730,18 +1730,19 @@ thingsSearch:
       #   # indexHint:
       #   #   "t.attributes.location": 1
       # customMetricsPersistence optionally overrides the MongoDB read preference / read concern used for the
-      # count based "customMetrics" queries above. When unset (the default), the general query.persistence
-      # settings are used - so this can offload the periodic operator metric counts to a secondary node without
-      # affecting user-facing search/count requests.
+      # count based "customMetrics" queries above - e.g. to offload the periodic operator metric counts to a
+      # secondary node without affecting user-facing search/count requests.
+      # Each setting is inherited from the general query.persistence config when it is not set here, so either
+      # key can be configured on its own: setting only readConcern keeps the query.persistence readPreference.
       # customMetricsPersistence:
       #   readPreference: secondaryPreferred
-      #   readConcern: default
+      #   readConcern: local
       # customAggregationMetricsPersistence optionally overrides the MongoDB read preference / read concern used
-      # for the "customAggregationMetrics" ($group) queries above. When unset (the default), the general
-      # query.persistence settings are used. Can be configured independently from customMetricsPersistence.
+      # for the "customAggregationMetrics" ($group) queries above, with the same per-setting inheritance from
+      # query.persistence. Can be configured independently from customMetricsPersistence.
       # customAggregationMetricsPersistence:
       #   readPreference: secondaryPreferred
-      #   readConcern: default
+      #   readConcern: local
     # dispatchers contains tuning for the things-search service's Pekko dispatchers.
     # Defaults below match the in-jar HOCON defaults and only need overriding under sustained load.
     dispatchers:
--- a/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultOperatorMetricsConfig.java
+++ b/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultOperatorMetricsConfig.java
@@ -32,6 +32,7 @@ import javax.annotation.concurrent.Immutable;
 
 import org.eclipse.ditto.internal.utils.config.ConfigWithFallback;
 import org.eclipse.ditto.internal.utils.config.KnownConfigValue;
+import org.eclipse.ditto.internal.utils.persistence.mongo.config.MongoDbConfig;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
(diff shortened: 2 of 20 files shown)


thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultOperatorMetricsConfig.java
```diff
@@ -59,6 +73,14 @@ private DefaultOperatorMetricsConfig(final ConfigWithFallback updaterScopedConfi
                 OperatorMetricsConfigValue.CUSTOM_METRICS);
         customAggregationMetricConfigs = loadCustomAggregatedMetricConfigurations(updaterScopedConfig,
                 OperatorMetricsConfigValue.CUSTOM_AGGREGATION_METRIC);
+        customMetricsPersistenceConfig = updaterScopedConfig.hasPath(CUSTOM_METRICS_PERSISTENCE_PATH)
```
hu-ahmed: This is where the partial-configuration trap comes from. `DefaultSearchPersistenceConfig.of(...)` applies its own `ConfigValue` defaults (`primaryPreferred` / `default`) for anything the block omits, so `hasPath` gives block-level optionality but no per-field fallback to `query.persistence`.

`DittoSearchConfig` already builds `queryPersistenceConfig` (line 105) before `operatorMetricsConfig` (line 110), so the fallback is available right here at construction:

```java
// DittoSearchConfig
operatorMetricsConfig = DefaultOperatorMetricsConfig.of(configWithFallback, queryPersistenceConfig);
```

and inside, for each block, take the configured value when that specific path is present and the `query.persistence` value otherwise, instead of routing the whole block through `DefaultSearchPersistenceConfig`.

If you'd rather keep the block as an all-or-nothing replacement, I'd validate that both `readPreference` and `readConcern` are present when the block exists and fail with a `DittoConfigError` otherwise. Silent divergence is the only outcome I think we should rule out.

thingsearch/service/src/main/resources/search.conf
```diff
@@ -402,6 +402,28 @@ ditto {
         #   # index-hint { "t.attributes.location": 1 }
         # }
       }
+
+      # Optional dedicated MongoDB read settings for the count based "custom-metrics" queries above.
```
hu-ahmed: Two notes on the commented examples:

1. The `${?THINGS_SEARCH_...}` lines below are inert as shipped, since the assignments they belong to are commented out. They're the only commented-out `${?ENV}` lines in any Ditto service reference conf — every other one is live — so someone grepping for env var names will reasonably expect them to work. Fine if the intent is "copy this block into your own conf", but worth making that explicit in the comment, or dropping those two lines per block.
2. `readConcern = "default"` is a slightly unfortunate value to show in the example, since `ReadConcern.DEFAULT` maps to an *empty* read concern, which actively clears a client-level `majority` rather than inheriting it. `"local"` matches the stated use case better.

deployment/helm/ditto/values.yaml
```diff
@@ -1729,6 +1729,19 @@ thingsSearch:
       #   # or as index key spec:
       #   # indexHint:
       #   #   "t.attributes.location": 1
+      # customMetricsPersistence optionally overrides the MongoDB read preference / read concern used for the
```
hu-ahmed: "When unset (the default), the general `query.persistence` settings are used" is only true when the whole block is omitted — see the main review comment. Worth stating explicitly whether this block is a partial override or a complete replacement, since operators will read it as per-setting.

thingsearch/service/src/test/java/org/eclipse/ditto/thingsearch/service/common/config/DefaultOperatorMetricsConfigTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Test;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/**
+ * Unit test for {@link DefaultOperatorMetricsConfig}, focusing on the optional per-metric-type persistence
+ * (read preference / read concern) configuration.
+ */
+public final class DefaultOperatorMetricsConfigTest {
```
hu-ahmed: Thanks for adding this — it's the first test for this class. Two additions I'd suggest, since the PR also (correctly) changes `equals`/`hashCode` to include both metric maps:

- a `testHashCodeAndEquals()` using `EqualsVerifier`, matching every sibling (`DefaultSearchPersistenceConfigTest:45`, `DefaultCustomMetricConfigTest`, …)
- a case covering partial configuration, i.e. what a block with only one of the two fields resolves to — that's the behaviour under discussion in the main comment and it's currently unpinned

The new `DefaultSearchPersistenceConfig.of(config, configPath)` overload is also untested.

thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/read/ThingsSearchPersistence.java
```diff
@@ -90,6 +90,27 @@ default Source<Long, NotUsed> sudoCount(Query query, DittoHeaders dittoHeaders,
...
+     * @param indexHint the optional index hint (string for index name, object for index key spec).
+     * @param readPreferenceOverride the optional MongoDB read preference to use for this query (e.g.
+     * {@code "secondaryPreferred"}); when {@code null} the persistence default read preference is used.
+     * @param readConcernOverride the optional MongoDB read concern to use for this query (e.g. {@code "local"});
+     * when {@code null} the persistence default read concern is used.
+     * @return an {@link Source} which emits the count.
+     * @throws NullPointerException if {@code query} is {@code null}.
+     * @since 3.9.7
+     */
+    default Source<Long, NotUsed> sudoCount(Query query, DittoHeaders dittoHeaders,
```
hu-ahmed: The `default` implementation silently discards both overrides, which is the right call for compatibility, but the Javadoc reads as though they're applied. Worth a line saying the default ignores them and that `MongoThingsSearchPersistence` is the implementation honouring them.

Also, that's three `sudoCount` overloads now. Not worth churning in this PR, but if a fourth knob appears I'd bundle these into an options holder rather than adding a sixth positional parameter.

hu-ahmed (commented):
Tightly-scoped change and the aggregation half is exactly right — `MongoThingsAggregationPersistence` is used only by `OperatorAggregateMetricsProviderActor`, so binding the read settings once at construction can't leak into user-facing search. The new `SudoCountThings` fields being optional and additive also makes this safe across a rolling upgrade. I verified locally that both touched modules compile clean and the new tests pass (`SudoCountThingsTest` 12/12, `DefaultOperatorMetricsConfigTest` 2/2). The `equals`/`hashCode` fix in `DefaultOperatorMetricsConfig` is a nice drive-by — both metric maps were previously ignored.

One substantive issue, and it's about how the "optional" part behaves.

## The config block is all-or-nothing, but it's documented as if it were per-key

`DefaultOperatorMetricsConfig` (line 76) decides via `hasPath` at the *block* level. Once the block exists, `DefaultSearchPersistenceConfig` fills any key you left out from its own `ConfigValue` defaults — `primaryPreferred` / `default` — and not from `query.persistence`. So a half-filled block doesn't inherit the missing half; it silently replaces it.

I ran this against the branch's config classes, simulating a deployment with `ditto.mongodb.options.readPreference = secondaryPreferred` and `readConcern = majority`:

```
A) no block configured            -> inherits query.persistence            (correct)
B) block sets readPreference only -> readConcern  silently becomes "default"
C) block sets readConcern only    -> readPreference silently becomes "primaryPreferred"
D) block sets both                -> exactly as configured                (correct)
```

Case C is the one that bites: an operator asking only for the cheap read concern gets the periodic metric queries pulled **back onto the primary** — the opposite of this PR's purpose — with no warning and no log line.

I originally suggested per-key inheritance, but I'm happy to take the whole-block reading instead: *if you configure this block, you configure all of it.* That's a perfectly coherent design. It just needs three things to actually hold, because right now nothing enforces it and the docs say the opposite:

**1. The Helm chart currently generates half-filled blocks itself.** In `search-extension.conf.tpl` (line 151) `readPreference` is the gate and `readConcern` is conditional, so the chart's minimum-viable path renders exactly the block the design forbids:

```
$ helm template t . --set ...customAggregationMetricsPersistence.readPreference=nearest
custom-aggregation-metrics-persistence {
  readPreference = "nearest"          # no readConcern -> silently "default"
}
```

Worth dropping the `{{- if .readConcern }}` and giving `readConcern` a value in `values.yaml`, so a rendered block is never partial. Related: `readConcern` can't currently be set on its own at all through the chart, which is awkward given "searches stay `linearizable`, metrics run `local`" is the motivating example in the PR description.

**2. Nothing enforces completeness in code.** If the block is meant as a full replacement, I'd require both keys when it's present and fail with a `DittoConfigError` otherwise. At minimum, log the effective values at startup — there's precedent, `MongoThingsSearchPersistence` already logs `Query readConcern=<...> readPreference=<...>` on construction, and the metrics path deserves the same visibility.

**3. The wording says the opposite of the behaviour.** `values.yaml` (line 1732) and `search.conf` (line 406) both say "When unset (the default), the general `query.persistence` settings are used", which reads per-key. It should say: when the block is *omitted entirely* you inherit `query.persistence`; if you configure it, set **both** keys, because an omitted key falls back to `primaryPreferred` / `default`.

This third point is the one I'd want fixed whichever design wins — docs promising per-key inheritance over code doing block-level replacement is what turns a design choice into a trap.

thjaeckle:
Thanks — you're right on all three, and case C was a real trap. Went with per-key inheritance rather than enforced completeness: each block is now resolved against `query.persistence`, so a block only overrides the settings it names and `readConcern = local` alone keeps your `secondaryPreferred`. That also let the chart keep both keys optional — it now renders the block when either is set, so a readConcern-only block is reachable (it wasn't before). Both metric paths log their effective read settings and their source at startup, mirroring `MongoThingsSearchPersistence`'s existing line, and the `values.yaml` / `search.conf` / javadoc wording now matches the behaviour. `DefaultSearchPersistenceConfig.of(Config, String)` became `ofScopedConfig(Config)` since the inherited values have to be merged in before the block is parsed. Config test is now an explicit A/B/C/D matrix, and I re-checked all four cases through `helm template`.

hu-ahmed (approved):
LGTM! 
Thanks for the improvements

PR #2511, opened 2026-07-31
support fn: pipeline expressions in connection target topic filters

alstanchev:
Connection target topics get a second filter parameter next to the RQL `filter`: **`fn-filter`**, a placeholder
pipeline which is evaluated per outbound signal.

```
_/_/things/twin/events?fn-filter=topic:action|fn:filter('ne','deleted')
_/_/things/live/messages?fn-filter=header:content-type|fn:lower()|fn:filter('like','application/json*')
_/_/things/twin/events?filter=gt(attributes/counter,42)&fn-filter=header:ditto-originator|fn:filter('ne','some:subject')
```

**What it does**
- Filters on any placeholder available for an outbound signal — `header:`, `topic:`, `thing:`, `entity:`,
  `feature:`, `resource:`, `request:`, `time:`, `connection:id`, and `thing-json:` for data carried by a thing
  event — optionally passed through value-transforming `fn:` stages. It works for every signal type a target topic
  can carry (twin events, live events, live commands, live messages), also where an RQL filter has no thing data to
  match.
- Shape: `<placeholder>[|fn:<value stage>...]|fn:filter('<rqlFunction>',<comparedValue>)` — it starts with the
  placeholder to filter and ends with its only `fn:filter` stage (`eq`, `ne`, `like`, `exists`).
- The topic is published when the pipeline resolves. A placeholder which does not resolve (e.g. an absent header)
  suppresses the topic, also for `ne`; publishing in that case is opted into with an `fn:default(...)` stage before
  the filter.
- `fn-filter` is repeatable: all `fn-filter`s and the RQL `filter` must match (AND). OR is expressed by several
  `topics` entries, as before. `filter` is always RQL, `fn-filter` is always a pipeline — told apart by name.

**Validation** (create / modify / test / open, HTTP 400): an expression starting with a function, a pipeline placed
in `filter` or RQL placed in `fn-filter`, a last stage other than `fn:filter` (incl. a bare placeholder), a second
`fn:filter`, the 3-parameter `fn:filter` form, `fn:delete()` anywhere, a stage without the `fn:` prefix, an unknown
`rqlFunction` literal, `exists` with anything but the constant `'true'`, the constant compared value `'exists'`, a
leading placeholder without a name or with whitespace at its colon, more than 10 `fn:` stages. A repeated `filter`,
`namespaces` or `extraFields` parameter is rejected with `TopicParseException` (400) instead of a 500. A rejection
message suggests a corrected expression only if that expression is itself valid.

**Runtime**: `SignalFilter` evaluates the `fn-filter`s before the RQL filter and before enrichment;
`OutboundMappingProcessorActor` evaluates them again per topic for topics with `extraFields` (they never see
enriched data). A `RuntimeException` during evaluation is logged, recorded as a failure in the connection logs and
treated as a non-match.

**API / docs**: `FilteredTopic#getFnFilters()`, `FilteredTopicBuilder#withFnFilters(...)` (`@since 4.0.0`).
`basic-connections.md` (syntax, unresolved placeholders, restrictions, characters which cannot be carried in a stored
topic, rolling-upgrade note), `basic-placeholders.md`, `basic-changenotifications.md`, pointers from the messages,
acknowledgements and enrichment pages.

System tests (ditto-testing [feature/target-topic-pipeline-filter](https://github.com/eclipse-ditto/ditto/actions/runs/35599728081/job/106332719150))

### DIFF
--- a/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/ImmutableFilteredTopic.java
+++ b/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/ImmutableFilteredTopic.java
@@ -21,11 +21,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
@@ -45,13 +45,15 @@ final class ImmutableFilteredTopic implements FilteredTopic {
     private static final String QUERY_ARG_VALUE_DELIMITER = "=";
 
     private static final String FILTER_ARG = "filter";
+    private static final String FN_FILTER_ARG = "fn-filter";
     private static final String NAMESPACES_ARG = "namespaces";
     private static final String EXTRA_FIELDS_ARG = "extraFields";
 
 
     private final Topic topic;
     private final List<String> namespaces;
-    @Nullable private final String filterString;
+    @Nullable private final String filter;
+    @Nullable private final String fnFilter;
     @Nullable private final ThingFieldSelector extraFields;
 
     private ImmutableFilteredTopic(final ImmutableFilteredTopicBuilder builder) {
@@ -60,7 +62,8 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         namespaces = null != namespacesFromBuilder
                 ? Collections.unmodifiableList(new ArrayList<>(namespacesFromBuilder))
                 : Collections.emptyList();
-        filterString = Objects.toString(builder.filter, null);
+        filter = Objects.toString(builder.filter, null);
+        fnFilter = Objects.toString(builder.fnFilter, null);
         extraFields = builder.extraFields;
     }
 
@@ -102,7 +105,12 @@ final class ImmutableFilteredTopic implements FilteredTopic {
 
     @Override
     public Optional<String> getFilter() {
-        return Optional.ofNullable(filterString);
+        return Optional.ofNullable(filter);
+    }
+
+    @Override
+    public Optional<String> getFnFilter() {
+        return Optional.ofNullable(fnFilter);
     }
 
     @Override
@@ -131,8 +139,10 @@ final class ImmutableFilteredTopic implements FilteredTopic {
     }
 
     private String getQueryParametersAsString() {
-        return join(QUERY_ARG_DELIMITER, getQueryParameterString(NAMESPACES_ARG, String.join(",", namespaces)),
-                getQueryParameterString(FILTER_ARG, filterString),
+        return join(QUERY_ARG_DELIMITER,
+                getQueryParameterString(NAMESPACES_ARG, String.join(",", namespaces)),
+                getQueryParameterString(FILTER_ARG, filter),
+                getQueryParameterString(FN_FILTER_ARG, fnFilter),
                 getQueryParameterString(EXTRA_FIELDS_ARG, extraFields));
     }
 
@@ -169,13 +179,14 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         final ImmutableFilteredTopic that = (ImmutableFilteredTopic) o;
         return topic == that.topic &&
                 namespaces.equals(that.namespaces) &&
-                Objects.equals(filterString, that.filterString) &&
+                Objects.equals(filter, that.filter) &&
+                Objects.equals(fnFilter, that.fnFilter) &&
                 Objects.equals(extraFields, that.extraFields);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(topic, namespaces, filterString, extraFields);
+        return Objects.hash(topic, namespaces, filter, fnFilter, extraFields);
     }
 
     /**
@@ -187,12 +198,14 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         private final Topic topic;
         @Nullable private Collection<String> namespaces;
         @Nullable private CharSequence filter;
+        @Nullable private CharSequence fnFilter;
         @Nullable private ThingFieldSelector extraFields;
 
         private ImmutableFilteredTopicBuilder(final Topic topic) {
             this.topic = checkNotNull(topic, "topic");
             namespaces = null;
             filter = null;
+            fnFilter = null;
             extraFields = null;
         }
 
@@ -208,6 +221,14 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         public ImmutableFilteredTopicBuilder withFilter(@Nullable final CharSequence filter) {
             if (supportsFilters()) {
                 this.filter = filter;
+            }
+            return this;
+        }
+
+        @Override
+        public ImmutableFilteredTopicBuilder withFnFilter(@Nullable final CharSequence fnFilter) {
+            if (supportsFilters()) {
+                this.fnFilter = fnFilter;
             }
             return this;
         }
@@ -261,6 +282,7 @@ final class ImmutableFilteredTopic implements FilteredTopic {
             return getBuilder(parseTopic(topicName))
                     .withNamespaces(parseNamespaces(queryParameters.get(NAMESPACES_ARG)))
                     .withFilter(queryParameters.get(FILTER_ARG))
+                    .withFnFilter(queryParameters.get(FN_FILTER_ARG))
                     .withExtraFields(parseExtraFields(queryParameters.get(EXTRA_FIELDS_ARG)))
                     .build();
         }
@@ -275,10 +297,19 @@ final class ImmutableFilteredTopic implements FilteredTopic {
             if (null == queryParamsString || queryParamsString.isEmpty()) {
                 return Collections.emptyMap();
             }
-            return Arrays.stream(queryParamsString.split(QUERY_ARG_DELIMITER))
-                    .map(paramString -> paramString.split(QUERY_ARG_VALUE_DELIMITER, 2))
-                    .filter(queryParamPair -> 2 == queryParamPair.length)
-                    .collect(Collectors.toMap(queryParamPair -> urlDecode(queryParamPair[0]), av -> urlDecode(av[1])));
+            final Map<String, String> queryParameters = new HashMap<>(4);
+            for (final String paramString : queryParamsString.split(QUERY_ARG_DELIMITER)) {
+                final String[] queryParamPair = paramString.split(QUERY_ARG_VALUE_DELIMITER, 2);
+                if (2 != queryParamPair.length) {
+                    continue;
+                }
+                final String name = urlDecode(queryParamPair[0]);
+                if (null != queryParameters.putIfAbsent(name, urlDecode(queryParamPair[1]))) {
+                    // no query parameter is repeatable - a duplicated one is rejected like before
+                    throw new IllegalStateException("Duplicate key " + name);
+                }
+            }
+            return queryParameters;
         }
 
         private static String urlDecode(final String value) {
--- a/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/ImmutableFilteredTopic.java
+++ b/connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/ImmutableFilteredTopic.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -51,7 +52,7 @@ final class ImmutableFilteredTopic implements FilteredTopic {
 
     private final Topic topic;
     private final List<String> namespaces;
-    @Nullable private final String filterString;
+    private final List<String> filters;
     @Nullable private final ThingFieldSelector extraFields;
 
     private ImmutableFilteredTopic(final ImmutableFilteredTopicBuilder builder) {
@@ -60,7 +61,10 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         namespaces = null != namespacesFromBuilder
                 ? Collections.unmodifiableList(new ArrayList<>(namespacesFromBuilder))
                 : Collections.emptyList();
-        filterString = Objects.toString(builder.filter, null);
+        final Collection<String> filtersFromBuilder = builder.filters;
+        filters = null != filtersFromBuilder
+                ? Collections.unmodifiableList(new ArrayList<>(filtersFromBuilder))
+                : Collections.emptyList();
         extraFields = builder.extraFields;
     }
 
@@ -102,7 +106,12 @@ final class ImmutableFilteredTopic implements FilteredTopic {
 
     @Override
     public Optional<String> getFilter() {
-        return Optional.ofNullable(filterString);
+        return filters.isEmpty() ? Optional.empty() : Optional.of(filters.get(0));
+    }
+
+    @Override
+    public List<String> getFilters() {
+        return filters;
     }
 
     @Override
@@ -131,9 +140,13 @@ final class ImmutableFilteredTopic implements FilteredTopic {
     }
 
     private String getQueryParametersAsString() {
-        return join(QUERY_ARG_DELIMITER, getQueryParameterString(NAMESPACES_ARG, String.join(",", namespaces)),
-                getQueryParameterString(FILTER_ARG, filterString),
-                getQueryParameterString(EXTRA_FIELDS_ARG, extraFields));
+        final List<String> queryParameterStrings = new ArrayList<>(2 + filters.size());
+        queryParameterStrings.add(getQueryParameterString(NAMESPACES_ARG, String.join(",", namespaces)));
+        for (final String filter : filters) {
+            queryParameterStrings.add(getQueryParameterString(FILTER_ARG, filter));
+        }
+        queryParameterStrings.add(getQueryParameterString(EXTRA_FIELDS_ARG, extraFields));
+        return join(QUERY_ARG_DELIMITER, queryParameterStrings.toArray(new String[0]));
     }
 
     private static String getQueryParameterString(final String parameterName, @Nullable final Object parameterValue) {
@@ -169,13 +182,13 @@ final class ImmutableFilteredTopic implements FilteredTopic {
         final ImmutableFilteredTopic that = (ImmutableFilteredTopic) o;
         return topic == that.topic &&
                 namespaces.equals(that.namespaces) &&
-                Objects.equals(filterString, that.filterString) &&
+                filters.equals(that.filters) &&
                 Objects.equals(extraFields, that.extraFields);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(topic, namespaces, filterString, extraFields);
+        return Objects.hash(topic, namespaces, filters, extraFields);
     }
 
     /**
@@ -186,13 +199,13 @@ final class ImmutableFilteredTopic implements FilteredTopic {
 
         private final Topic topic;
         @Nullable private Collection<String> namespaces;
-        @Nullable private CharSequence filter;
+        @Nullable private List<String> filters;
         @Nullable private ThingFieldSelector extraFields;
 
         private ImmutableFilteredTopicBuilder(final Topic topic) {
             this.topic = checkNotNull(topic, "topic");
             namespaces = null;
-            filter = null;
+            filters = null;
             extraFields = null;
         }
 
@@ -206,8 +219,15 @@ final class ImmutableFilteredTopic implements FilteredTopic {
 
         @Override
         public ImmutableFilteredTopicBuilder withFilter(@Nullable final CharSequence filter) {
+            return withFilters(null != filter ? Collections.singletonList(filter) : null);
+        }
+
+        @Override
+        public ImmutableFilteredTopicBuilder withFilters(@Nullable final Collection<? extends CharSequence> filters) {
             if (supportsFilters()) {
-                this.filter = filter;
+                this.filters = null != filters
+                        ? filters.stream().map(CharSequence::toString).collect(Collectors.toList())
+                        : null;
             }
             return this;
         }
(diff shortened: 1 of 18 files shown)


documentation/src/main/resources/pages/ditto/basic-connections.md
```diff
@@ -307,6 +309,125 @@ filter values before using them:
...
+* `header:ditto-origin` resolves to the ID of the connection that originally caused the signal, e.g.
+  `filter=fn:filter(header:ditto-origin,'ne','some-other-connection-id')`.
+  Ditto already suppresses signals a connection caused itself by default; filtering on
+  `ditto-origin` is only needed to additionally exclude signals caused by *other* connections.
+
+Several `fn:` stages can be chained with `|`. Each stage only runs if the previous one resolved
+(matched), so chaining is **AND**: every stage must match for the pipeline to resolve, for example:
+
+```text
+filter=fn:filter(header:ditto-originator,'ne','some:subject')|fn:filter(header:ditto-origin,'ne','some-connection-id')
```
thjaeckle: @alstanchev to be honest, I think using `|` to declare **AND** operation is very confusing. I would have expected it to be **OR** (as `|` is the "or" logical expression in programming languages).  

I understand the need for AND .. But maybe we can do that e.g. as we would in HTTP APIs:  
In HTTP APIs, you could just define several `filter` query parameters - and those would be combined with AND semantics.

E.g. instead:
```
?filter=fn:filter(header:ditto-originator,'ne','some:subject')
&filter=fn:filter(header:ditto-origin,'ne','some-connection-id')
```

We could then even keep the `|`, however as **OR** semantics

documentation/src/main/resources/pages/ditto/basic-connections.md
```diff
@@ -307,6 +309,125 @@ filter values before using them:
...
+```
+
+A pipeline can also be combined with a leading RQL expression, joined by an unquoted `|` immediately
+followed by `fn:`, again with **AND** semantics -- both the RQL part and every pipeline stage must
+match:
+
+```text
+filter=gt(attributes/counter,42)|fn:filter(header:ditto-originator,'ne','some:subject')
```
thjaeckle: This would also be "nicer" by just defining several `filter` params:

```
?filter=gt(attributes/counter,42)
&filter=fn:filter(header:ditto-originator,'ne','some:subject')
```

documentation/src/main/resources/pages/ditto/basic-connections.md
alstanchev: In my head when i was doing it, i was tinking of the bash pipe and dito pipeline syntax but you are right it makes more sense like you suggest. 
Will change it

alstanchev:
https://github.com/eclipse-ditto/ditto/actions/runs/30714370336/job/91407324869

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/ConnectionValidator.java
```diff
@@ -329,10 +330,44 @@ private void validateSourceAndTargetAddressesAreNonempty(final Connection connec
...
+                                .build();
+                    }
+                    if (partitionedFilters.getPipelineExpressions().size() > 1) {
+                        throw ConnectionConfigurationInvalidException
+                                .newBuilder("The topic '" + topic + "' of the target with address '" +
+                                        target.getAddress() + "' declares " +
+                                        partitionedFilters.getPipelineExpressions().size() +
+                                        " pipeline 'filter' parameters - at most one pipeline filter is allowed " +
+                                        "per topic.")
+                                .description("Combine several pipeline conditions by chaining 'fn:' stages with " +
```
hu-ahmed: Thomas's |-as-AND concern is only half resolved
The repeated-param syntax landed, but the validator still caps pipeline filters at one per topic, and its error message tells users to "Combine several pipeline conditions by chaining 'fn:' stages with '|'". Docs repeat it  (basic-connections.md:358-364, 374-377). So the only way to express two pipeline conditions is still |-as-AND — the exact construct that was flagged as confusing.

alstanchev: This was my intention. To not use pipe as a symbol to join RQL and functions but to keep the already established way of writing fn expressions. I am still on vacation so maybe we could discuss this on the next community meeting.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/TargetTopicFilter.java
```diff
@@ -0,0 +1,185 @@
...
+     * without a name (e.g. {@code header:}) is rejected, because it passes the grammar but cannot be evaluated.
+     *
+     * @param fnFilter the raw {@code fn-filter} parameter value to validate.
+     * @param dittoHeaders the headers of the command which triggered the validation, stamped onto the thrown
+     * exception for correlation.
+     * @throws ConnectionConfigurationInvalidException if the expression is invalid, e.g. because it is empty, is an
+     * RQL expression, references an unknown placeholder or function, starts with a placeholder without a name, has
+     * an invalid function signature or exceeds the maximum number of pipeline stages.
+     */
+    public static void validateFnFilter(final String fnFilter, final DittoHeaders dittoHeaders) {
```
thjaeckle, line 147: **Finding 1** — `validateFnFilter` is thorough about *syntax* but accepts expressions that are semantically constant, i.e. that publish everything or nothing regardless of the signal.

Measured against two signals (`ditto-originator` = `someone:else` vs `excluded:subject`):

| `fn-filter` | validation | other subject | excluded subject | |
|---|---|---|---|---|
| `fn:filter(header:ditto-originator,'ne','excluded:subject')` | accepted | publish | suppress | correct baseline |
| `fn:filter(header:ditto-originator,'NE','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'neq','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `header:ditto-originator\|fn:filter('NE','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'eq')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'ne')` | accepted | suppress | suppress | **always silent** |
| `…'ne','excluded:subject')\|fn:default('x')` | accepted | publish | publish | **filter defeated** |
| `…'ne','excluded:subject')\|fn:delete()` | accepted | suppress | suppress | **always silent** |

Mechanically: an unknown `rqlFunction` name leaves `Parameters#getRqlFunction()` empty, `applyRqlFunction` returns `false`, and the pipeline stays unresolved — so a single typo makes the target permanently silent, with no error at create time and no connection-log entry at runtime (suppression is intentionally silent). A trailing `fn:default(...)` resolves every unresolved pipeline and therefore publishes exactly the events the operator wanted excluded.

The docs acknowledge both ("Double-check spelling", "this is not enforced"), but for a feature whose primary use case is *not* leaking events to a given subject, all three checks are cheap to add right here:

* reject a literal `rqlFunction` outside `{eq, ne, like, exists}`
* reject `eq`/`ne`/`like` given without a compared value (the 2-param form is only meaningful for `exists`)
* reject a trailing `fn:default(...)` / `fn:delete()` stage

The first two are literal-only checks, so a placeholder-valued `rqlFunction` would simply be skipped. Note that rejecting a trailing `fn:filter`-less pipeline outright would break the documented bare-placeholder form (`fn-filter=header:ditto-originator`), so the terminal-stage check should target `fn:default`/`fn:delete` specifically rather than requiring `fn:filter`.

For context on how solid the rest is: I fuzzed 30 malformed expressions through this method, including `like` patterns full of regex metacharacters, and none escaped as anything other than `ConnectionConfigurationInvalidException` — no 500s from the validation path.

connectivity/model/src/main/java/org/eclipse/ditto/connectivity/model/ImmutableFilteredTopic.java
```diff
@@ -275,10 +297,19 @@ private static Map<String, String> parseQueryParameters(@Nullable final String q
...
+            final Map<String, String> queryParameters = new HashMap<>(4);
+            for (final String paramString : queryParamsString.split(QUERY_ARG_DELIMITER)) {
+                final String[] queryParamPair = paramString.split(QUERY_ARG_VALUE_DELIMITER, 2);
+                if (2 != queryParamPair.length) {
+                    continue;
+                }
+                final String name = urlDecode(queryParamPair[0]);
+                if (null != queryParameters.putIfAbsent(name, urlDecode(queryParamPair[1]))) {
+                    // no query parameter is repeatable - a duplicated one is rejected like before
+                    throw new IllegalStateException("Duplicate key " + name);
```
thjaeckle, line 309: **Finding 3** (low, pre-existing) — this preserves the `Collectors.toMap` behaviour faithfully, but the resulting exception is a raw `IllegalStateException`, whereas the sibling error in the same parser (`parseTopic`) throws `TopicParseException`, which is a `DittoRuntimeException` mapped to HTTP 400.

An `IllegalStateException` escaping here reaches `RootRouteExceptionHandler.matchAny`, which completes with `StatusCodes.INTERNAL_SERVER_ERROR` and no body — so a user who writes `?filter=a&filter=b` (or duplicates `fn-filter` or `namespaces`) gets a bare 500 with no hint about what is wrong.

Since this loop is being rewritten by hand anyway, the outcome is now documented in `basic-connections.md` ("a repeated query parameter makes the topic string unparseable") and three new tests assert `IllegalStateException`, this seems like the moment to switch to `TopicParseException` with a message naming the duplicated parameter. That also keeps the parser's error contract uniform.

documentation/src/main/resources/pages/ditto/basic-placeholders.md
```diff
@@ -325,6 +325,39 @@ _org.eclipse.ditto/device-123/things/live/messages/hello.world_ these placeholde
...
+In a connection's [target topic filter](basic-connections.html#filtering-with-placeholder-functions),
+a placeholder function pipeline whose last stage is [`fn:filter()`](#function-library) may be used as the
+`fn-filter` query parameter, alongside an optional [RQL expression](basic-rql.html) in the `filter` parameter
+(at most one of each; if both are given, both must match). As with
+[RQL expressions when filtering for Ditto Protocol messages](#scope-rql-expressions-when-filtering-for-ditto-protocol-messages),
+such a pipeline is a bare expression and placeholders must not be surrounded by curly braces. The pipeline may
+start with a placeholder or directly with a function, e.g.
+`header:ditto-originator|fn:filter('ne','some:subject')` or
+`fn:filter(header:ditto-originator,'ne','some:subject')`; the topic is published exactly when the pipeline
+resolves to a value, which is why the last stage must be `fn:filter()` -- the stage that yields the boolean
```
thjaeckle, line 339: **Finding 4** — this says the last stage "**must** be `fn:filter()`", while `basic-connections.md` says "must ... this is not enforced, a pipeline that ends with a value-producing stage is accepted but pointless" and additionally documents a bare `header:ditto-originator` (no `fn:filter` stage at all) as a valid exists-style filter.

So "must" overstates it in two directions: it is not enforced, and it is not actually required. Suggest softening to "should" here and keeping `basic-connections.md` as the authoritative explanation — or enforcing the terminal stage as sketched in finding 1, in which case this page becomes correct as written and the `basic-connections.md` caveat can go.

thjaeckle (commented):
## Review @ `8d06aa9f89`

**State**: open, not draft, mergeable/clean against `master`. CI green on the head commit (`build`, `check-license-header-year`, `eclipsefdn/eca`).

**Verification I ran locally** on the head commit: full build of `connectivity/service` and its dependencies, then the touched suites — 304 tests green (`TargetTopicFilterTest` 42, `ConnectionValidatorTest` 52, `SignalFilterWithFilterTest` 19, `SignalFilterTest` 125, `OutboundMappingProcessorActorTest` 27, `ImmutableFilteredTopicTest` 33, `ImmutableTargetTest` 6). `japicmp` runs on `connectivity/model` and passes despite `FilteredTopic.getFnFilter()` being added as a non-default interface method. I also spot-checked the documentation claims against the implementation: the absent-header outcome table matches `PipelineFunctionFilter`/`FilterFunctions` on all four rows (including `ne`-on-absent → published), and the "10 stages" limit matches `MAX_COUNT_PIPELINE_FUNCTIONS + 1`.

### Earlier review round

@thjaeckle's two threads (`|`-as-AND is confusing; prefer several parameters as in an HTTP API) are resolved and outdated — the design was reworked into separate `filter` / `fn-filter` parameters, which addresses them.

@hu-ahmed's thread is **still open**. Their point was that the concern was only half resolved, since `|` remained the only way to combine two pipeline conditions. That is still true in the current design, but the construct that was actually flagged — using `|` to join an *RQL expression* with a *pipeline* — is gone, and what remains is Ditto's pre-existing pipeline `|` semantics, consistent with `fn:` usage everywhere else. I think @alstanchev's position is right; the thread just needs an explicit decision recorded (they asked to discuss it in a community meeting).

### On the design

The separate-parameter design is the right call and landed cleanly. `filter` is always RQL, `fn-filter` is always a pipeline, told apart by **name, never by content**; an `fn:` value in `filter` is rejected with an error naming `fn-filter` and echoing the value ready to paste, and vice versa. The quote-aware `|`-splitting heuristic is deleted, removing the risk of re-interpreting existing RQL filters containing an unquoted `|` in a property path (with a regression test for exactly that). Extensibility is good: a further filter dialect is one more named parameter, AND-combined; OR is expressed by listing several `topics` entries. `fn-filter` also works on `_/_/things/live/commands`, where an RQL data filter cannot meaningfully match — a real capability gain. The `fn:default('true')|` seed for function-first pipelines is a clever workaround for the pipeline API lacking a resolved entry value, and unusually well explained in the Javadoc, including why placeholder-first pipelines must *not* be seeded.

### Findings

**1. Misconfigured `fn-filter`s pass validation and silently become constants** — the one I'd like addressed before merge (inline comment on `TargetTopicFilter.validateFnFilter`). Measured at runtime against two signals (`ditto-originator` = `someone:else` vs `excluded:subject`):

| `fn-filter` | validation | other subject | excluded subject | |
|---|---|---|---|---|
| `fn:filter(header:ditto-originator,'ne','excluded:subject')` | accepted | publish | suppress | correct baseline |
| `fn:filter(header:ditto-originator,'NE','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'neq','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `header:ditto-originator\|fn:filter('NE','excluded:subject')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'eq')` | accepted | suppress | suppress | **always silent** |
| `fn:filter(header:ditto-originator,'ne')` | accepted | suppress | suppress | **always silent** |
| `…'ne','excluded:subject')\|fn:default('x')` | accepted | publish | publish | **filter defeated** |
| `…'ne','excluded:subject')\|fn:delete()` | accepted | suppress | suppress | **always silent** |

A typo'd `rqlFunction` makes the target publish nothing at all, with no error and no connection-log entry (suppression is intentionally silent). A trailing `fn:default(...)` does the opposite and publishes everything — i.e. exactly the events the operator wanted excluded reach the target. The PR documents both as restrictions ("Double-check spelling", "this is not enforced"), but for an ops-facing feature whose purpose is not leaking events to a subject, documentation feels thin when all three checks are cheap in `validateFnFilter`: reject a literal `rqlFunction` outside `{eq, ne, like, exists}`, reject `eq`/`ne`/`like` without a compared value, and reject a trailing `fn:default(...)`/`fn:delete()`.

Worth stressing that validation is otherwise impressively robust — I fuzzed 30 malformed expressions, including `like` patterns full of regex metacharacters, and **none** escaped as anything other than `ConnectionConfigurationInvalidException`, so no 500s from the validation path.

**2. The PR description no longer describes the PR.** It still documents "combined filters: `filter=<rql>|fn:...`" and the splitting of a filter string at the first unquoted `|`, a mechanism that no longer exists. Both system-test links also predate the 2026-09-04 redesign, so system tests have not been run against the current design — worth a fresh run given the model and parser changed.

**3. Duplicate query parameter yields a bare HTTP 500** (low, pre-existing; inline comment on `ImmutableFilteredTopic`). `?filter=a&filter=b`, a duplicated `fn-filter`, or a duplicated `namespaces` throws a raw `IllegalStateException("Duplicate key …")` from `FilteredTopicStringParser.parseQueryParameters`, which `RootRouteExceptionHandler.matchAny` turns into `INTERNAL_SERVER_ERROR` with no body — while the sibling error in the same method correctly throws `TopicParseException` → 400. Behaviour is unchanged from the previous `Collectors.toMap`, but since this PR rewrote that loop by hand, documented the outcome and added tests asserting `IllegalStateException`, switching to `TopicParseException` is a two-line improvement.

**4. Documentation inconsistency on the terminal stage** (inline comment on `basic-placeholders.md`). `basic-placeholders.md` says the last stage "**must** be `fn:filter()`"; `basic-connections.md` says "must … this is not enforced" and also documents a bare `header:ditto-originator` with no `fn:filter` at all as a valid exists-style filter. Align on "should", or enforce it per finding 1.

Nothing here is a correctness bug in the happy path. The pre-enrichment gate plus post-enrichment re-evaluation split is sound — the per-topic re-evaluation in `applyFilter` is genuinely necessary because `SignalFilter` only does `anyMatch` across a target's topics — `pairTargetsWithTopics` correctly restricts to matching streaming types, and the runtime failure policy (catch `RuntimeException` → warn + connection-log failure → treat as non-match) is applied consistently at both call sites.

connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/TargetTopicFilter.java
```diff
@@ -0,0 +1,384 @@
...
+     */
+    private static void validateFilterStage(final String fnFilter, final String stage, final List<String> params,
+            final boolean carrierIsSeed, final DittoHeaders dittoHeaders) {
+        final boolean constantsOnly = params.stream().allMatch(TargetTopicFilter::isQuotedConstant);
+        if (params.size() == 3) {
+            if (constantsOnly) {
+                throw filterStageNeverLooksAtTheSignal(fnFilter, stage, "all three parameters are constants",
+                        dittoHeaders);
+            }
+            rejectUnknownRqlFunctionLiteral(fnFilter, stage, params.get(1), dittoHeaders);
```
thjaeckle: Non-blocking, but this is the last surviving member of the class of misconfiguration finding 1 was about, and the check fits right here.

`ExistsFunction` resolves its second parameter with `Boolean.parseBoolean`, so any literal that is not a case-insensitive `"true"` silently means `false` — which does not make the filter constant, it **inverts** it:

| `fn-filter` | validation | header present | header absent |
| --- | --- | --- | --- |
| `fn:filter(header:ditto-originator,'exists','true')` | accepted | publish | suppress |
| `fn:filter(header:ditto-originator,'exists','false')` | accepted | suppress | publish |
| `fn:filter(header:ditto-originator,'exists','ture')` | accepted | **suppress** | **publish** |
| `fn:filter(header:ditto-originator,'exists','yes')` | accepted | **suppress** | **publish** |
| `fn:filter(header:ditto-originator,'exists','1')` | accepted | **suppress** | **publish** |

For an exclusion filter, publishing exactly the inverse set is arguably worse than publishing nothing: a typo in the third parameter turns "only signals that carry the header" into "only signals that do not", with no error at create time and no connection-log entry at runtime. `basic-placeholders.md` already specifies this parameter as `String true|false`, so the accepted values are pinned down.

The check is the same literal-only shape as `rejectUnknownRqlFunctionLiteral` right here: in the 3-parameter form, when the literal `rqlFunction` is `exists` and the third parameter is a quoted constant, require it to be `true` or `false` (case-insensitively). A placeholder-valued third parameter would be skipped, as elsewhere.

thjaeckle (approved):
Approving. Findings 1, 3 and 4 from the previous round are addressed, and 1 went further than I asked: `fn:delete()` at any position, `fn:default` after any `fn:filter`, `fn:filter` stages with no placeholder at all, and the terminal-stage rule actually enforced rather than softened in the docs.

**Verified on `e0b5aa2671`:** full build of `connectivity/service` and its dependencies, then the touched suites — 323 tests green (`TargetTopicFilterTest` 60, `ConnectionValidatorTest` 52, `SignalFilterTest` 125, `OutboundMappingProcessorActorTest` 27, `SignalFilterWithFilterTest` 19, `ImmutableFilteredTopicTest` 34, `ImmutableTargetTest` 6); `japicmp` on `connectivity/model` clean. All eight rows of my finding-1 table are now rejected at connection create/modify time, each with a pinned test. I also threw 35 degenerate inputs at `validateFnFilter` (`""`, `" "`, `"|"`, `"||"`, `"'"`, `"'''"`, unbalanced parens, 4-parameter forms, `FN:FILTER`) and none produced anything other than a clean accept or a `ConnectionConfigurationInvalidException` — in particular the unguarded `stages.get(stages.size() - 1)` is safe, because the grammar check upstream guarantees at least one stage. The hand-rolled quote-aware split agrees with the resolver's grammar on `,`, `|`, `)` and escaped quotes inside constants, and the `carrierIsSeed` tracking holds up on the subtle cases: `fn:default(header:x)|fn:filter('ne','y')` is correctly rejected (the seed makes that `fn:default` a no-op, so the carrier really is constant), while `header:x|fn:default('none')|fn:filter('ne','y')` is correctly accepted.

**One thing before merging: the description needs another pass.**

- It still lists only the three checks I originally suggested. The commit that landed also rejects a bare placeholder as the last stage — which removes a form the docs previously described as valid — plus `fn:delete()` anywhere, `fn:default` after any `fn:filter`, and placeholder-free `fn:filter` stages. The bare-placeholder removal is the one user-visible behaviour change in this round and is worth calling out explicitly.
- The **Run Link** points at run `34968649515` (2026-09-15 12:24 UTC), which predates `e0b5aa2671` (2026-09-16 18:54 UTC). The run that actually covers this head is [`35142495106`](https://github.com/eclipse-ditto/ditto/actions/runs/35142495106) (2026-09-16 19:45 UTC, green) — worth linking that one instead. Note that `35137598557`, two minutes after the commit, failed, and the ditto branch did not move afterwards, so something on the `ditto-testing` side was adjusted in between; its logs have expired, so I could not tell which cases.

Left one non-blocking inline comment on `validateFnFilter`.

The `%`-in-a-filter-value behaviour that surfaced while reviewing this is pre-existing and unrelated to the feature, so it is tracked separately in #2563 — nothing to do here.

PR #2513, opened 2026-07-31
Optimize trace-context propagation to avoid per-hop full header-map copy

thjaeckle:
## What

Optimizes trace-context propagation so a traced hop no longer copies the entire `DittoHeaders` map, and no longer re-validates headers when re-wrapping them.

## Why

A production JFR profile (10 min, profile config) showed trace-context propagation as a significant CPU cost across the fleet — roughly 7% of overall microservice JVM-user CPU, and the single biggest hotspot in the gateway (~14.5%, driven by per-signal propagation on the streaming/WebSocket paths). It is paid on effectively every signal and cluster hop because tracing runs with an always-on sampler.

Two O(N) costs were incurred per hop:

1. `KamonHttpContextPropagation.propagateContextToHeaders(...)` did `new HashMap<>(headers)` — copying **all** incoming headers — only to add the ~2 W3C trace-context entries (`traceparent`/`tracestate`) that the propagation actually writes.
2. The vast majority of call sites then re-wrapped the result via `DittoHeaders.of(startedSpan.propagateContext(x.getDittoHeaders()))`, sending the map back through the **validating** `DittoHeaders` constructor and re-parsing every JSON-typed header value on each hop.

## What changed

- **`KamonHttpContextPropagation`**: added `propagateContextToNewMap(Context)`, which writes only the propagated trace-context entries into a fresh small map instead of copying all incoming headers.
- **`StartedKamonSpan`**: overrides `propagateContext(DittoHeaders)` to apply only the propagated delta onto the already-validated (trusted) headers via `toBuilder()`. This removes both the full-map copy and the full `entrySet()` re-scan that the `StartedSpan` interface default performed to recover the delta.
- **Call-site migration** (20 sites): replaced `DittoHeaders.of(startedSpan.propagateContext(x.getDittoHeaders()))` with the trusted `startedSpan.propagateContext(x.getDittoHeaders())` overload, dropping the redundant re-validating re-wrap. Raw-`Map<String,String>` callers (Kafka/AMQP/RabbitMQ consumers, `RequestTracingDirective`, `ExternalMessage` header paths) were intentionally left on the `Map` overload.

## Behavior & compatibility

- Behavior is unchanged: the same context is propagated and the same trace-context headers are written. The pre-existing `StartedKamonSpanTest` equivalence test asserts the trusted overload is byte-for-byte equal to `DittoHeaders.of(propagateContext((Map) ...))`, and new unit tests cover the delta path.
- API change is additive only (a new public method on `KamonHttpContextPropagation` plus an override); no existing signature changed.
- `internal/utils/tracing` test suite passes (114 tests).

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/MappingTimer.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/MappingTimer.java
@@ -192,7 +192,7 @@ final class MappingTimer {
     }
 
     private Adaptable propagateContextToAdaptable(final Adaptable adaptable) {
-        return adaptable.setDittoHeaders(DittoHeaders.of(startedSpan.propagateContext(adaptable.getDittoHeaders())));
+        return adaptable.setDittoHeaders(startedSpan.propagateContext(adaptable.getDittoHeaders()));
     }
 
     /**
@@ -209,7 +209,7 @@ final class MappingTimer {
     }
 
     private Signal<?> propagateContextToSignalDittoHeaders(final Signal<?> signal) {
-        return signal.setDittoHeaders(DittoHeaders.of(startedSpan.propagateContext(signal.getDittoHeaders())));
+        return signal.setDittoHeaders(startedSpan.propagateContext(signal.getDittoHeaders()));
     }
 
     /**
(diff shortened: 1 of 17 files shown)


hu-ahmed (approved):
LGTM!

PR #2515, opened 2026-07-31
Compute IndexLengthRestrictionEnforcer pointer byte-length without allocating

thjaeckle:
## What

`IndexLengthRestrictionEnforcer.jsonPointerBytes(...)` computed the pointer's UTF-8 byte length via `jsonPointer.toString().getBytes(UTF_8).length`. This runs **per string/number leaf of every indexed thing** on the search-write path, and each call allocates two throwaway objects:

- a full pointer `String` (level join + per-level `escapeTilde`), and
- a `byte[]` from the UTF-8 encoder — created solely to read `.length`, then discarded.

A production JFR profile showed this as a notable thingssearch CPU and allocation hotspot.

## Change

Replace it with a direct, allocation-free scan over the pointer's `JsonKey`s that sums the UTF-8 byte length of each escaped segment plus the `/` separators — no intermediate `String`, no `byte[]`.

The computed count is **byte-for-byte identical** to `toString().getBytes(UTF_8).length`:
- RFC 6901 escaping of `~` → `~0` (mirroring `ImmutableJsonPointer.escapeTilde`, which does not escape `/`), and
- the single `?` replacement byte the UTF-8 encoder emits for an unpaired surrogate.

The change is contained to `IndexLengthRestrictionEnforcer`; the `json` module is untouched.

## Correctness

An equivalence test asserts the new computation equals `pointer.toString().getBytes(UTF_8).length` for root/empty, deep, tilde-escaped (`~`, `~~`, `~0`/`~1`, leading/trailing), multi-byte unicode (2-byte é, 3-byte €, 4-byte 😀) and unpaired-surrogate pointers, plus a 5000-iteration seeded randomized property check over random BMP and supplementary code points. Index-length enforcement behaviour is unchanged. `IndexLengthRestrictionEnforcerTest` passes (10 tests).

### DIFF
--- a/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/IndexLengthRestrictionEnforcer.java
+++ b/thingsearch/service/src/main/java/org/eclipse/ditto/thingsearch/service/persistence/write/IndexLengthRestrictionEnforcer.java
@@ -20,6 +20,7 @@ import java.nio.CharBuffer;
 import java.nio.charset.CoderResult;
 import java.util.Optional;
 
+import org.eclipse.ditto.json.JsonKey;
 import org.eclipse.ditto.json.JsonPointer;
 import org.eclipse.ditto.json.JsonValue;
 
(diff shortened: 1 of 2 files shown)


hu-ahmed (approved):
LGTM!

PR #2527, opened 2026-08-26
Memoize read-subject parsing and ditto-read-subjects rendering

thjaeckle:
## What

Removes two per-signal costs around the `ditto-read-subjects` header that scale with the number of READ-granted subjects of a policy: the repeated *parsing* of the header value into `Set<AuthorizationSubject>`, and the repeated *rendering* of that value for every emitted `ThingEvent`.

## Why

After a policy migration that roughly doubled the number of READ-granted subjects per Thing (and lengthened each subject ID), things-service CPU rose by ~25% permanently. A production JFR profile attributed the increase to paths that scale with the read-subject payload rather than with request volume: `DittoHeaders` construction and copies (~17%), enforcer tree walks plus per-event subject-set merging (~10%), and serialization including header-value escaping (~19%).

Both costs are recomputations of something that only changes when the policy does:

- `DittoHeaders#getReadGrantedSubjects()` / `#getReadRevokedSubjects()` allocate N `AuthorizationSubject` instances plus a `HashSet` on *every* call, although the result is a pure function of the immutable header value. On the hot path this happens once per publish and then again on every subscriber node per received message (`ReadSubjectExtractor`), per connection target (`SignalFilter`) and per streaming session (`StreamingSessionActor`) — so the cost is (number of granted subjects) × (local fan-out).
- `ThingCommandEnforcement#addEffectedReadSubjectsToThingSignal` merges the READ classifications into a fresh `HashSet` and re-serializes the subject IDs for every emitted `ThingEvent`. Rendering means allocating a `JsonValue` per subject, building a `JsonArray` (which eagerly builds its CBOR representation) and escaping the whole array into a header string — repeated per event, all of it scaling with the number of granted subjects.

This follows the memoization precedents already in the code base: `Header.parsedValue` and `PolicyEnforcer`'s read-classification memo.

## What changed

**Memoized parsed read-subject sets (`base/model`)**

- `Header` gets a second lazily populated memo slot next to `parsedValue`, and `AbstractDittoHeaders#getAuthorizationSubjectSet` derives the subject set through it. `Header` instances are shared by reference across header copies, so the memo also survives `DittoHeaders.newBuilder(existing).build()`.
- The derived value is built with `Set.copyOf`, not `Collections.unmodifiableSet(new HashSet<>(…))`: the memo field is deliberately non-volatile — the same benign-data-race pattern as `String#hashCode()` — so only final-field freeze semantics (JLS 17.5) keep a racily published set from being observed partially constructed. Both read-subject header keys share a single derivation function, which keeps the untyped memo slot free of type confusion. The constraints are documented on the accessor.
- No new field on `AbstractDittoHeaders`, so the existing `EqualsVerifier` tests stay valid; the memo does not participate in `equals`/`hashCode`.

**Memoized rendered header value (`policies/enforcement`, `things/service`)**

- New `PolicyEnforcer#getReadGrantedSubjectsHeaderValue(ResourceKey, boolean)` performs the merge and the rendering, memoized per resource key alongside the existing read-classification memo and invalidated the same way — `CachingPolicyEnforcerProvider` replaces the instance wholesale on every grant change. The include-partial variant is keyed by resource and bounded by the existing read-classification cache size (entries are 1:1 with classification entries, so no new configuration knob); the root-only variant does not depend on the resource and gets a single slot.
- Subject IDs are sorted and de-duplicated, so the rendered value is stable across nodes and restarts. Consumers parse it back into a `Set`, so ordering is not semantically relevant.
- New `DittoHeadersBuilder#readGrantedSubjects(JsonArray)` lets the enforcement hand the memoized array to the builder without re-rendering or re-parsing it. It keeps the element-type check the parse-based validator would have done, so an invalid value still cannot reach a reader that would fail on `asString()`. Empty arrays are put as `[]`, exactly as the collection variant does. `JsonArray` is not a `Collection`, so there is no overload ambiguity.
- `putJsonValue` now also seeds the new `Header`'s JSON memo with the value it just rendered: that `Header` is fresh per built signal, so otherwise the next reader — typically the pub/sub topic extractor on the very same node — would parse the string straight back into the object already at hand.
- `ThingCommandEnforcement#addEffectedReadSubjectsToThingSignal` keeps its signature, so `LiveSignalEnforcement` and `StreamRequestingCommandEnforcement` benefit unchanged.

## Behavior & compatibility

- The rendered header value is set-equal to the previous one; only element order changes (sorted instead of `HashSet` iteration order). During a rolling upgrade, mixed nodes therefore emit differently ordered values — harmless, since every consumer parses the header into a `Set`.
- The `Set` returned by `getReadGrantedSubjects()` / `getReadRevokedSubjects()` is now unmodifiable and may be a shared instance. The javadoc already stated that changes to it are not reflected back into the headers, and no in-repo caller mutates it. `Set.copyOf` tolerates duplicate subject IDs and collapses them, matching what the previous `Collectors.toSet()` did — nothing on the wire enforces uniqueness, so a test pins that behaviour.
- Only an interface method is added (`METHOD_ADDED_TO_INTERFACE`); japicmp is clean on both `base/model` and `policies/enforcement`. `readGrantedSubjects(null)` written as a bare literal becomes an ambiguous overload at compile time (source-only, no binary impact).
- Memory: the new memo holds at most one small `JsonArray` per cached resource classification, whose strings are already retained by the classification memo.
- New tests cover the memo identity and immutability, survival across header copies, the `JsonArray` builder overload (round-trip, non-string element rejection, empty array, equality with the collection variant), the memoized render formula (sorted, de-duplicated, per-resource, `forNamespace` children independent, empty-grant policy) and the empty-value end-to-end case. The existing JMH benchmarks were extended with a production-shaped policy — 22 granted subjects, realistic ID lengths — because the previous two-subject fixture understates the cost in exactly the dimension that caused the regression.

### DIFF
--- a/base/model/src/main/java/org/eclipse/ditto/base/model/headers/AbstractDittoHeaders.java
+++ b/base/model/src/main/java/org/eclipse/ditto/base/model/headers/AbstractDittoHeaders.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
@@ -67,6 +68,19 @@ public abstract class AbstractDittoHeaders implements DittoHeaders {
 
     private static final String REDACTED_VALUE = "***";
     private static final Set<String> REDACTED_HEADER_KEYS = loadRedactedHeaderKeys();
+
+    /*
+     * Single derivation function shared by the "ditto-read-subjects" and "ditto-read-revoked-subjects" headers, so
+     * that the untyped memo slot of Header is only ever populated with a Set<AuthorizationSubject> for those keys.
+     * Set.copyOf is required (not Collections.unmodifiableSet): only its final fields make the result safe to
+     * publish through Header's deliberately non-volatile memo field.
+     */
+    private static final Function<JsonValue, Set<AuthorizationSubject>> AUTHORIZATION_SUBJECT_SET_DERIVATION =
+            jsonValue -> Set.copyOf(jsonValue.asArray()
+                    .stream()
+                    .map(JsonValue::asString)
+                    .map(AuthorizationSubject::newInstance)
+                    .collect(Collectors.toList()));
 
     final Map<String, Header> headers;
 
@@ -214,11 +228,11 @@ public abstract class AbstractDittoHeaders implements DittoHeaders {
     }
 
     private Set<AuthorizationSubject> getAuthorizationSubjectSet(final HeaderDefinition definition) {
-        final JsonArray jsonValueArray = getJsonArrayForDefinition(definition);
-        return jsonValueArray.stream()
-                .map(JsonValue::asString)
-                .map(AuthorizationSubject::newInstance)
-                .collect(Collectors.toSet());
+        @Nullable final Header header = headers.get(definition.getKey());
+        if (null == header) {
+            return Set.of();
+        }
+        return header.getDerivedValue(AUTHORIZATION_SUBJECT_SET_DERIVATION);
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


thjaeckle:
@hu-ahmed could you have a brief look on this optimisation?

hu-ahmed (approved):
LGTM!

PR #2530, opened 2026-08-28
Fix devops-managed WoT validation dynamicConfigs never reaching the validator

thjaeckle:
Dynamic WoT validation config sections written through the devops API are persisted, replicated and returned by every read endpoint, but are never honoured by the validator. Sections from the static config file keep working.

On a production-like cluster the validator ran WoT validation config **revision 503 for over a week** while the entity had advanced to 530 — so every dynamic section added after revision 503 was silently ignored, and things whose payload violated their Thing Model were rejected with `wot:payload.validation.error` despite a matching scope with `log-warning-instead-of-failing-api-calls: true`.

## Root cause

The config is replicated to all things nodes through an `ORSet` in `WotValidationConfigDData`. Its update function returned a freshly built set, ignoring the value it was handed:

```java
public CompletionStage<Void> add(final JsonObject config) {
    final ORSet<JsonObject> newSet = ORSet.<JsonObject>empty().add(selfUniqueAddress, config);
    return update(getKey(0), writeAll(), orSet -> newSet);   // current value ignored
}
```

The replicator *merges* that result into the locally stored value rather than replacing it, and `ORSet` merge resolves per writing node. Such a write therefore only supersedes the element written by its own node — elements written by other nodes, in particular by pods that have since left the cluster, stay in the set forever.

`ThingsRootActor` then read the set with `getElements().iterator().next()`, commented *"we only expect one"*. With several elements present it picked whichever came first in iteration order; all replicas having converged on the same value, that was the same element on every node. Once it was an outdated one, every node stayed on the outdated config.

This also explains why the problem is invisible from the outside: `GET /devops/wot/config` and `…/merged` are answered from the persistence actor's in-memory entity, and `ditto:static` sections never travel through the distributed data at all. Only the validator reads the replicated copy.

## Changes

**Replication (commit 1)**

- Derive the `ORSet` update functions of `add()` and `clear()` from the set they are handed, so a write supersedes the elements of *all* nodes. This repairs an already corrupted set on the next write — no manual cleanup or cluster restart needed on deploy.
- Select the config with the highest revision in `ThingsRootActor` instead of an arbitrary one, and log a warning if the set ever holds more than one element.
- Publish the config from the persistence actor's `onEntityModified()` hook instead of from five individual command strategies, following the pattern `PolicyPersistenceActor` already uses. The hook runs after the event was persisted and applied, so the published config reflects the persisted state. Previously `MergeDynamicConfigSection` published the entity with its *pre-command* revision, and all strategies published *before* the event was persisted. `ddata` is consequently no longer a dependency of any command strategy.

**Timestamps (commit 2)**

`_created` and `_modified` never advanced while `_revision` kept increasing. Two independent defects:

- `ImmutableWotValidationConfig.of(…, revision, created, modified, …)` forwarded positionally to a constructor declaring `(…, revision, modified, created, …)`, exchanging the two timestamps on construction. `WotValidationConfigUtils.mergeWithEntity()` passed them pre-swapped and cancelled the defect out for the merged view, hiding it, while every other call site got them wrong. The constructor is aligned with the factory and the compensating swap removed — these must change together.
- The event strategies for merged and deleted dynamic config sections bumped the revision but copied both timestamps over unchanged. `modified` is now taken from the event timestamp, as `AbstractThingEventStrategy` does, so recovery replays deterministically.

## Notes

- No public API signature changes; `ImmutableWotValidationConfig` is package-private and japicmp passes.
- No new configuration, so no Helm changes. No feature toggle — this restores intended behaviour rather than changing it.
- Verified on a dev cluster: after the fix the validator loads revision 530 with all 20 devops scopes instead of revision 503 with 3, and the two affected Thing Models now log a warning instead of returning HTTP 400.

### DIFF
--- a/things/model/src/main/java/org/eclipse/ditto/things/model/devops/ImmutableWotValidationConfig.java
+++ b/things/model/src/main/java/org/eclipse/ditto/things/model/devops/ImmutableWotValidationConfig.java
@@ -68,8 +68,8 @@ final class ImmutableWotValidationConfig implements WotValidationConfig {
     @Nullable private final FeatureValidationConfig featureConfig;
     private final List<DynamicValidationConfig> dynamicConfig;
     @Nullable private final WotValidationConfigRevision revision;
-    @Nullable private final Instant modified;
     @Nullable private final Instant created;
+    @Nullable private final Instant modified;
     @Nullable private final Boolean deleted;
     @Nullable private final Metadata metadata;
 
@@ -81,8 +81,8 @@ final class ImmutableWotValidationConfig implements WotValidationConfig {
             @Nullable final FeatureValidationConfig featureConfig,
             final List<DynamicValidationConfig> dynamicConfig,
             @Nullable final WotValidationConfigRevision revision,
-            @Nullable final Instant modified,
             @Nullable final Instant created,
+            @Nullable final Instant modified,
             @Nullable final Boolean deleted,
             @Nullable final Metadata metadata) {
         this.configId = Objects.requireNonNull(configId, "configId");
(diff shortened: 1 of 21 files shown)


hu-ahmed (approved):
LGTM!
Thanks for the fix

PR #2532, opened 2026-08-31
#2531 fix partial-access filter dropping non-root object payloads

### LINKED ISSUE #2531: Partial read access: events with an object payload at a non-root path are silently dropped
## Summary

A subscriber holding **partial** READ permission on a thing (grants on specific subtrees rather than on `thing:/`) does not receive twin events whose payload is a JSON object at a non-root path — even when the event's path lies entirely inside a granted subtree.

Scalar payloads and root-path (`/`) events are unaffected, which makes the behaviour look inconsistent from the outside.

## Steps to reproduce

Given a policy that grants subject `X` READ on `thing:/features/status` only, and a subscriber authorized as `X` (connection target, SSE or WebSocket) on twin events:

| # | Request | Expected | Actual |
|---|---------|----------|--------|
| 1 | `PATCH /things/{id}/features/status` with `{"properties":{"value":42}}` | event delivered | **no event** |
| 2 | `PATCH /things/{id}/features/status/properties/value` with `42` | event delivered | event delivered |

Same permission, same subtree — only the shape of the payload differs.

## Cause

`AdaptablePartialAccessFilter.filterAdaptableWithResult` derives `eventPath` from `payload.path`, but only uses it in the non-object branch. Object payloads go to `filterAdaptablePayload(...)` → `JsonPartialAccessFilter.filterJsonByPaths(payload, accessiblePaths)`, which matches **thing-absolute** accessible-path pointers against a payload that is **relative to `payload.path`**.

For an event at `/features/status` with value `{"properties":{"value":42}}`, the accessible path `/features/status` matches nothing inside that payload, so the filtered payload becomes `{}` and `OutboundMappingProcessor.processWithPartialAccessPaths` drops the signal.

This is a regression: before partial-access event emitting was introduced, `addEffectedReadSubjectsToThingSignal` used `getSubjectsWithUnrestrictedPermission(<event resource path>)`, so such an event was delivered in full and never filtered.

## Secondary issue: the drop is invisible

Signals dropped on this path log at `DEBUG` only and increment no counter: `filtered` counts the signal, while `mapped`, `dropped` and `published` do not, and no connection-log entry is written. The symptom is therefore a large, unexplained gap between `filtered` and `published` with zero failures recorded anywhere. This case should increment the existing `dropped` counter.

## Suggested fix

In the object-payload branch:

1. if `eventPath` is at or below an accessible path, the whole payload is readable — pass it through unchanged;
2. otherwise, rebase the accessible paths relative to `eventPath` and filter with those (so an event at `/features` still yields the granted features);
3. if nothing remains, empty payload → drop, as today.

The prefix walk required by (1) already exists as `isPathAccessibleForNonObjectPayload`.

hu-ahmed:
fixes #2531

### DIFF
--- a/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
+++ b/internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
@@ -152,14 +152,65 @@ public final class AdaptablePartialAccessFilter {
         final JsonPointer eventPath = JsonPointer.of(adaptable.getPayload().getPath().toString());
 
         if (!isPayloadObject) {
-            final boolean isPathAccessible = isPathAccessibleForNonObjectPayload(eventPath, accessiblePaths);
+            final boolean isPathAccessible = isPathWhollyAccessible(eventPath, accessiblePaths);
             if (!isPathAccessible) {
                 return createEmptyPayloadAdaptable(adaptable);
             }
             return filterExtraFields(adaptable, accessiblePaths);
         }
 
-        return filterAdaptablePayload(adaptable, accessiblePaths);
+        // An event payload is relative to the event's own path, whereas the accessible paths are
+        // thing-absolute. Both must be expressed in the same frame of reference before filtering,
+        // otherwise a payload below a granted path matches nothing and is wrongly emptied.
+        if (isPathWhollyAccessible(eventPath, accessiblePaths)) {
+            // The event is rooted at or below a granted path: the entire payload is readable.
+            return filterExtraFields(adaptable, accessiblePaths);
+        }
+        final Set<JsonPointer> payloadRelativePaths = rebaseOnEventPath(eventPath, accessiblePaths);
+        if (payloadRelativePaths.isEmpty()) {
+            return createEmptyPayloadAdaptable(adaptable);
+        }
+        return filterAdaptablePayload(adaptable, payloadRelativePaths, accessiblePaths, eventPath.isEmpty());
+    }
+
+    /**
+     * Re-expresses the thing-absolute {@code accessiblePaths} relative to {@code eventPath}, keeping only those
+     * which can occur at all within a payload rooted at {@code eventPath}.
+     *
+     * @param eventPath the path the payload is rooted at.
+     * @param accessiblePaths the thing-absolute accessible paths.
+     * @return the accessible paths relative to {@code eventPath}, empty if none of them lie below it.
+     */
+    private static Set<JsonPointer> rebaseOnEventPath(final JsonPointer eventPath,
+            final Set<JsonPointer> accessiblePaths) {
+
+        if (eventPath.isEmpty()) {
+            return accessiblePaths;
+        }
+        final Set<JsonPointer> rebased = new HashSet<>();
+        for (final JsonPointer accessiblePath : accessiblePaths) {
+            if (isDescendantOrSelf(accessiblePath, eventPath)) {
+                rebased.add(accessiblePath.getSubPointer(eventPath.getLevelCount()).orElse(JsonPointer.empty()));
+            }
+        }
+        return rebased;
+    }
+
+    /**
+     * @return whether {@code path} is {@code base} itself or lies below it. Compared level by level rather than by
+     * string prefix, so {@code /features/foo} is not treated as lying below {@code /features/fo}.
+     */
+    private static boolean isDescendantOrSelf(final JsonPointer path, final JsonPointer base) {
+        final int baseLevelCount = base.getLevelCount();
+        if (path.getLevelCount() < baseLevelCount) {
+            return false;
+        }
+        for (int i = 0; i < baseLevelCount; i++) {
+            if (!path.get(i).equals(base.get(i))) {
+                return false;
+            }
+        }
+        return true;
     }
 
     private static Adaptable createEmptyPayloadAdaptable(final Adaptable adaptable) {
@@ -170,20 +221,29 @@ public final class AdaptablePartialAccessFilter {
                 .build();
     }
 
+    /**
+     * @param payloadRelativePaths accessible paths expressed relative to the event path, for filtering the value.
+     * @param accessiblePaths thing-absolute accessible paths, for filtering the (thing-rooted) extra fields.
+     * @param payloadIsThingRooted whether the payload is the thing itself, i.e. the event path is {@code "/"}.
+     */
     private static Adaptable filterAdaptablePayload(
             final Adaptable adaptable,
-            final Set<JsonPointer> accessiblePaths) {
+            final Set<JsonPointer> payloadRelativePaths,
+            final Set<JsonPointer> accessiblePaths,
+            final boolean payloadIsThingRooted) {
 
         final JsonObject originalPayload = adaptable.getPayload().getValue()
                 .filter(JsonValue::isObject)
                 .map(JsonValue::asObject)
                 .orElse(JsonFactory.newObject());
 
-        // ThingCreated/ThingModified/ThingMerged events at path "/" carry the full thing JSON
-        // here; stripping the identity/audit fields would make the protocol message hard to use
-        // for the partial reader. For events with non-thing-rooted object payloads the allowlist
-        // pointers simply don't exist in the payload and the addition is a no-op.
-        final JsonObject filteredPayload = filterJsonByPathsWithAllowlist(originalPayload, accessiblePaths);
+        // ThingCreated/ThingModified/ThingMerged events at path "/" carry the full thing JSON here;
+        // stripping the identity/audit fields would make the protocol message hard to use for the
+        // partial reader. The allowlist is thing-rooted, so it must only be applied to a thing-rooted
+        // payload — against a payload rooted deeper it would match same-named nested fields instead.
+        final JsonObject filteredPayload = payloadIsThingRooted
+                ? filterJsonByPathsWithAllowlist(originalPayload, payloadRelativePaths)
+                : JsonPartialAccessFilter.filterJsonByPaths(originalPayload, payloadRelativePaths);
 
         final PayloadBuilder payloadBuilder = Payload.newBuilder(adaptable.getPayload())
                 .withValue(filteredPayload);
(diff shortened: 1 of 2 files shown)


hu-ahmed:
[system test ](https://github.com/eclipse-ditto/ditto/actions/runs/33391167863)passed

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
```diff
@@ -152,14 +152,65 @@ public static Adaptable filterAdaptableWithResult(
...
+        // thing-absolute. Both must be expressed in the same frame of reference before filtering,
+        // otherwise a payload below a granted path matches nothing and is wrongly emptied.
+        if (isPathWhollyAccessible(eventPath, accessiblePaths)) {
+            // The event is rooted at or below a granted path: the entire payload is readable.
+            return filterExtraFields(adaptable, accessiblePaths);
+        }
+        final Set<JsonPointer> payloadRelativePaths = rebaseOnEventPath(eventPath, accessiblePaths);
+        if (payloadRelativePaths.isEmpty()) {
+            return createEmptyPayloadAdaptable(adaptable);
+        }
```
thjaeckle, line 172: **Extra-field leak, asymmetric with this PR's own fix (low–medium).**

`createEmptyPayloadAdaptable` seeds the builder from the original payload, so it keeps the **unfiltered** `extra` and only empties the value. The downstream drop guards (connectivity `OutboundMappingProcessor` and gateway `WebSocketRoute`) fire on `isEmptyPayload && !isEmptyOriginalPayload` — so when the *original* payload value is already an empty object, the event is **not** dropped and the full unfiltered `extra` (which can contain non-granted data) is delivered to a partial-access subscriber.

This PR routes object payloads through this branch for the first time, and the new test `extraFieldsAreClearedWhenNothingInThemIsAccessible` closes exactly this leak on the *pass-through* path — but not here. Filtering the extra here restores the symmetry:

```suggestion
        if (payloadRelativePaths.isEmpty()) {
            // Empty the value AND filter the extra: createEmptyPayloadAdaptable alone would hand the
            // partial reader the original, unfiltered extra whenever the drop guard does not fire
            // (i.e. the original payload value was already an empty object).
            return filterExtraFields(createEmptyPayloadAdaptable(adaptable), accessiblePaths);
        }
```

(The same latent leak exists at the two pre-existing `createEmptyPayloadAdaptable` call sites — the `!shouldFilter()` and non-object-not-accessible branches — for the empty-original-value case; worth a look if you want to close it holistically, though those are out of this PR's scope.)

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
```diff
@@ -152,14 +152,65 @@ public static Adaptable filterAdaptableWithResult(
...
+     *
+     * @param eventPath the path the payload is rooted at.
+     * @param accessiblePaths the thing-absolute accessible paths.
+     * @return the accessible paths relative to {@code eventPath}, empty if none of them lie below it.
+     */
+    private static Set<JsonPointer> rebaseOnEventPath(final JsonPointer eventPath,
+            final Set<JsonPointer> accessiblePaths) {
+
+        if (eventPath.isEmpty()) {
+            return accessiblePaths;
```
thjaeckle: **Minor: inconsistent set ownership (low).** Every other branch of `rebaseOnEventPath` returns a fresh `HashSet`; this one returns the caller's shared, resolver-owned `accessiblePaths`. No live bug (the sole consumer copies defensively before mutating), but the mixed contract is a latent aliasing hazard.

```suggestion
            return new HashSet<>(accessiblePaths);
```

thjaeckle (commented):
Thanks for the thorough fix and the reproducing tests — the root-cause analysis in #2531 is spot on. The core change is correct: expressing the thing-absolute grants and the payload-relative value in the same frame of reference (pass-through when the event is at/below a grant, `rebaseOnEventPath` when it is above them) is the right model, and it stays consistent with `filterJsonRecursive`'s existing ancestor-grant semantics. I confirmed thing-rooted (`"/"`) events are unchanged, the public API is untouched, and all 7 new + 31 existing related tests pass with no regression.

One finding worth acting on (extra-field leak, inline below) plus two optional cleanups. Not a blocker.

**Also from #2531, not addressed here (note only):** the secondary observability gap — genuinely-dropped events log at `DEBUG` and increment no `dropped` counter. The PR title only claims the drop fix, so this is fine to leave; flagging so it isn't lost.

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
```diff
@@ -152,14 +152,65 @@ public static Adaptable filterAdaptableWithResult(
...
+            }
+        }
+        return rebased;
+    }
+
+    /**
+     * @return whether {@code path} is {@code base} itself or lies below it. Compared level by level rather than by
+     * string prefix, so {@code /features/foo} is not treated as lying below {@code /features/fo}.
+     */
+    private static boolean isDescendantOrSelf(final JsonPointer path, final JsonPointer base) {
```
thjaeckle, line 207: **Minor: duplicated prefix-walk (low, optional).** `isDescendantOrSelf` hand-rolls a level-by-level pointer comparison, while `isPathWhollyAccessible` walks `getPrefixPointer` for essentially the inverse check. Two ad-hoc prefix checks in one class are two independent places for a future off-by-one / empty-pointer bug. A single shared `JsonPointer` prefix helper would keep them provably consistent. Non-blocking.

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
hu-ahmed: Good catch — confirmed, and I traced it end to end before changing anything. Both guards are `... && isEmptyPayload && !isEmptyOriginalPayload` (`OutboundMappingProcessor.java:352`, `WebSocketRoute.java:805`), and `createEmptyPayloadAdaptable` seeds the builder from the original payload, so an event whose value was *already* an empty object is not dropped and carries the unfiltered `extra` with it.

The asymmetry you point at is the crux: the `filterExtraFieldsIntoBuilder` change in this PR only helps paths that actually go through it, and this branch bypasses it entirely.

Applied your suggestion verbatim, plus a regression test (`extraFieldsAreClearedWhenPayloadPathIsOutsideGrants`) that fails without it.

On the two pre-existing call sites — I left them deliberately, and I'd value your view. Making `createEmptyPayloadAdaptable` filter the `extra` itself would close all three at once, but the `!shouldFilter()` site is the *no-access* case, and filtering there runs through `ALWAYS_INCLUDED_THING_FIELDS`, so a subject with no read access at all would still receive `thingId` / `_revision` / `_modified`. That looks like a separate semantic question (does the identity allowlist apply to no-access subjects, or only to partial readers as its javadoc states?) rather than a mechanical fix, so I'd rather do it in its own PR with its own tests. Happy to open an issue for it.

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
hu-ahmed: Applied — `new HashSet<>(accessiblePaths)`, so every branch returns a set this method owns. Agreed on the reasoning: a defensive copy in the current sole consumer shouldn't be the thing keeping this correct.

internal/utils/protocol/src/main/java/org/eclipse/ditto/internal/utils/protocol/AdaptablePartialAccessFilter.java
hu-ahmed: Went slightly further than deduplicating. `isPathWhollyAccessible` *is* "is the event path at or below some granted path", so it now delegates to `isDescendantOrSelf` and the `getPrefixPointer` walk is gone entirely — one pointer-prefix predicate left in the class.

I checked equivalence case by case before swapping it in: exact match, root granted (an empty `base` is an ancestor of everything), intermediate prefixes, and the empty-set case (`anyMatch` returns false, matching the old early return). The existing `AdaptablePartialAccessFilterTest` cases cover all four and still pass.

hu-ahmed:
Thanks for the careful review @thjaeckle — all three points addressed in c712564693, replies inline.

Summary:
- **Extra-field leak** — confirmed and fixed as you suggested, with a regression test. I left the two pre-existing `createEmptyPayloadAdaptable` sites alone on purpose; reasoning in the inline reply, and I'd like your read on it since it touches whether `ALWAYS_INCLUDED_THING_FIELDS` should apply to *no-access* subjects.
- **Set ownership** — applied.
- **Duplicated prefix-walk** — collapsed rather than deduplicated: `isPathWhollyAccessible` now delegates to `isDescendantOrSelf`, leaving one prefix predicate in the class.

`internal/utils/protocol` is green: 53 tests, including 8 in the new class and the 21 existing `AdaptablePartialAccessFilterTest` cases, no regressions.

On the observability gap from #2531 — agreed it's out of scope here, and thanks for flagging it so it isn't lost. I'll raise it separately; the `dropped` counter already exists, so it should be a small change, and it's the reason this took hours to find rather than minutes.

Two things worth stating plainly about verification: the unit tests are the only automated proof so far. Beyond them I verified the behaviour by hand against a local cluster — a partial reader granted READ on `/features/deviceStatus` only, over WebSocket, before and after the fix:

| merge | before | after |
|---|---|---|
| object at `/features/deviceStatus` (granted) | not delivered | delivered in full |
| scalar at `…/status/severity` (control) | delivered | delivered |
| object at `/features/grid` (not granted) | not delivered | not delivered |
| object at `/features` (mixed) | not delivered | delivered, `grid` stripped |

The control rows matter: the scalar write to the same granted subtree came through on both builds, and `grid` stayed hidden on both, so the fix restores delivery without loosening enforcement.

There are also three ITs on the ditto-testing side (`WebsocketIT`, branch `bugfix/2531-partial-access-non-root-object-payload`) asserting delivery *and* payload content for these cases. Existing partial-access ITs assert shape and non-leakage but never that a permitted event arrives — `partialAccessEventsFilteredForNestedAttributeUpdates` performs exactly the merge that regressed and passes either way, which is why this went unnoticed. Those ITs have not been executed yet; a system-test run is in flight and I'll report back.

thjaeckle (approved):
Reviewed - thanks a lot for the fix 👍

PR #2538, opened 2026-09-01
Harden host validation of connection targets

thjaeckle:
Tightens the outbound host validation applied to connection targets (`DefaultHostValidator`) in three ways:

* **Case-insensitive matching** of the allow-list and of the blocked-host regex. Hostnames are case-insensitive (RFC 4343), so a configured block rule such as `.*\.internal` previously did not match a differently cased host, and allow-list entries could be missed for the same reason.
* **Link-local addresses are blocked** (`169.254.0.0/16`, `fe80::/10`). These cover the cloud instance-metadata endpoint `169.254.169.254`, which was previously only blocked when an operator happened to configure the corresponding subnet.
* **IPv6 unique-local addresses are blocked** (`fc00::/7`, RFC 4193). `InetAddress#isSiteLocalAddress` only matches the deprecated `fec0::/10` prefix, so unique-local addresses such as `fd00:ec2::254` passed every built-in address-class check.

The default configuration is deliberately left unchanged: connectivity ships `blocked-hostnames = ""`, which short-circuits validation entirely, so these checks only take effect for deployments that have host validation enabled.

`HostValidatorTest` is extended accordingly: link-local and IPv6 unique-local blocking (both halves of `fc00::/7`, plus a public IPv6 address as control), and case-insensitivity of both the allow-list and the blocked-host regex.

### DIFF
--- a/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/DefaultHostValidator.java
+++ b/connectivity/service/src/main/java/org/eclipse/ditto/connectivity/service/messaging/validation/DefaultHostValidator.java
@@ -12,9 +12,11 @@
  */
 package org.eclipse.ditto.connectivity.service.messaging.validation;
 
+import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Collection;
+import java.util.Locale;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -24,8 +26,8 @@ import org.eclipse.ditto.connectivity.service.config.ConnectivityConfig;
 import org.apache.pekko.event.LoggingAdapter;
 
 /**
- * Validates a given hostname against a set of fixed blocked addresses (e.g. loopback, multicast, ...), a set of
- * blocked/allowed hostnames and a set of blocked subnets from configuration.
+ * Validates a given hostname against a set of fixed blocked addresses (e.g. loopback, link-local, multicast, ...), a
+ * set of blocked/allowed hostnames and a set of blocked subnets from configuration.
  * <p>
  * The allowed hostnames override the blocked hostnames e.g. if a host would be blocked because it resolves to a blocked
  * address (localhost, site-local, ...), the host can be allowed by adding it to the list allowed hostnames.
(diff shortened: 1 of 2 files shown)


hu-ahmed:
The `fc00::/7` mask and the branch ordering look correct to me, and I confirmed the defaults are untouched (`blocked-hostnames = ""` in both `connectivity.conf` and the Helm values). I checked out the branch and ran `HostValidatorTest` against both this implementation and `master`'s. Three things:

### 1. `expectBlockedHostRegexIsCaseInsensitive` passes on `master` too

It doesn't prove the regex fix. The mixed-case host is unresolvable, so on `master` it falls past the regex into `validateInetAddressesAndSubnets` and comes back `invalid` — which is also `isValid() == false`:

```
master: gateway.things.SVC.Cluster.LOCAL -> valid=false, "... is invalid: nodename nor servname provided"
this PR: gateway.things.SVC.Cluster.LOCAL -> valid=false, "... the host is blocked"
```

It's DNS-dependent too: with a search domain or NXDOMAIN hijacking, `master` returns `valid` and the test flips to failing. The other three new tests do fail on `master`, so only this one is affected.

A deterministic version — I verified this one gives `valid=true` on `master` and blocked here:

```java
@Test
public void expectBlockedHostRegexIsCaseInsensitive() {
    // resolve to a public address so only the regex can block the host:
    when(connectionConfig.getBlockedHostnames()).thenReturn(List.of("dummy.org"));
    when(connectionConfig.getBlockedSubnets()).thenReturn(emptyList());
    when(connectionConfig.getBlockedHostRegex()).thenReturn("^.*\\.svc\\.cluster\\.local$");
    final DefaultHostValidator.AddressResolver resolver = host -> resolveHost(host,
            "dummy.org".equals(host) ? new byte[]{1, 2, 3, 4} : new byte[]{8, 8, 8, 8});

    final HostValidator underTest = getHostValidatorWithCustomResolver(resolver);

    assertFalse(underTest.validateHost("gateway.things.SVC.Cluster.LOCAL").isValid());
}
```

The pre-existing `expectKubernetesClusterDNSIsBlocked` has the same weakness if you want to fix both.

### 2. Three IPv6 encodings of `169.254.169.254` still pass

Probed against this branch:

```
::ffff:169.254.169.254   blocked   (Java folds it to Inet4Address)
::169.254.169.254        allowed   IPv4-compatible, ::/96 (RFC 4291 2.5.5.1)
64:ff9b::a9fe:a9fe       allowed   NAT64 well-known prefix (RFC 6052)
2002:a9fe:a9fe::         allowed   6to4 (RFC 3056)
```

The NAT64 one is the one that matters in practice — on an IPv6-only cluster with NAT64/DNS64 that address routes to the v4 metadata endpoint. The other two are mostly historical. Same shape of check as `isUniqueLocalAddress`: pull out the embedded v4 bytes and re-run the address-class checks on them. Fine as a follow-up rather than in this PR, but probably worth an issue so it isn't lost.

### 3. `blockedAddresses.isEmpty()` gates more than the block-list

You call this out in the description, so just naming the root cause: that early return conflates "no blocked *hostnames* configured" with "disable *all* validation" — `blocked-subnets` and `blocked-host-regex` are silently ignored too unless a blocked hostname happens to be set. A separate `block-private-addresses = true` would let the address-class checks be on by default without operators having to configure a hostname block-list they don't otherwise want. Not for this PR.

Only 1 blocks merge from my side; 2 and 3 are follow-ups.

thjaeckle:
Thanks — you're right on 1, and I reproduced it the same way before changing anything: running this PR's test file against `master`'s `DefaultHostValidator`, `expectBlockedHostRegexIsCaseInsensitive` passed while the other three new tests failed. The assertion could not distinguish "blocked by the regex" from "the host did not resolve", and as you say the outcome depends on the resolver of the machine running it.

Fixed in 2b5faed66d, using your resolver approach, with two additions:

* It lives in the shared `getHostValidatorWithBlockedRegexPattern()` helper, so the pre-existing `expectKubernetesClusterDNSIsBlocked` gets the same treatment rather than just the new test.
* The helper now also carries a positive control — `gateway.things.example.com` has to come back valid — so if the setup ever stops letting non-matching hosts through, the test fails instead of quietly passing for the wrong reason.

I also escaped the remaining dots in the pattern (`^.*\.svc\.cluster\.local$`); previously only the first one was escaped.

Verified in both directions, with the same test file:

* against this branch's validator: 11/11 pass
* against `master`'s validator: all four new tests fail, the regex one now included

On 2 — agreed, and one addition: `::ffff:169.254.169.254` is already covered, because Java folds it to an `Inet4Address` and it hits the link-local check. `::169.254.169.254`, `64:ff9b::a9fe:a9fe` and `2002:a9fe:a9fe::` are not, and I agree NAT64 is the one with real-world reach — an IPv6-only cluster with DNS64 is a realistic deployment, the other two much less so. Agreed it belongs in a follow-up rather than here.

On 3 — agreed. The early return conflating "no blocked hostnames configured" with "validation off" also makes `blocked-subnets` and `blocked-host-regex` silently inert, which is easy to get wrong operationally, since neither of those keys hints that it depends on a third one. A separate `block-private-addresses` flag would let the address-class checks stand on their own. Also a follow-up.

I'll leave the follow-up issues to be filed separately rather than opening them from here.

hu-ahmed (approved):
LGTM!

PR #2540, opened 2026-09-03
Render Helm customIndexes at the config path Ditto actually reads

thjaeckle:
`thingsSearch.config.customIndexes` never reached Ditto. The template emitted the block under

```
  ditto.search.updater.persistence.custom-indexes
```

while SearchConfigValue declares

```
  CUSTOM_INDEXES("index-initialization.custom-indexes", Collections.emptyMap())
```

so DittoSearchConfig#loadCustomIndexes looks under `ditto.search.index-initialization.custom-indexes` and finds nothing. search.conf documents the same location - `custom-indexes {}` sits inside the `index-initialization { }` block, next to `activated-index-names`.

The consequence is worse than the feature silently doing nothing. Because IndexInitializer.dropUndefinedIndices() removes every index whose name is neither built-in nor a *configured* custom index, declaring an index in the chart made Ditto drop it on startup rather than create and keep it. Operators who had pre-created the index by hand - the recommended approach on a large collection, where a rolling build avoids a cold create - lost it on the next things-search restart.

### DIFF
--- a/deployment/helm/ditto/CHANGELOG.md
+++ b/deployment/helm/ditto/CHANGELOG.md
@@ -15,6 +15,18 @@ sections: `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`.
 
 ## [Unreleased]
 
+## [4.7.1]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


kalinkostashki (approved):
nice catch 👍

PR #2542, opened 2026-09-03
Document PUT create for connections #1746

### LINKED ISSUE #1746: Can't create Hono connection via Connections HTTP API
A Hono connection requires the connection ID to be set to the ID of the Hono tenant that this connection is to be used for.

But, when trying to create a Hono connection via the Connections HTTP API, along with specifying the connection ID in the request JSON, there is an error response:
```
{
  "status": 400,
  "error": "connectivity:id.notsettable",
  "message": "It is not allowed to provide an explicit ID for a connection.",
  "description": "Remove the ID from the connection as it will be generated."
}
```
Therefore it seems a Hono connection can't be created in the recommended way using the HTTP API. Instead, the piggyback commands API needs to be used.

AzazelSensei:
POST `/api/2/connections` still generates the ID and rejects one in the body. I documented `PUT /api/2/connections/{connectionId}` with `If-None-Match: *` as the create-only path, and pointed the Hono page at that so the connection ID can be the tenant ID.

Fixes #1746

### DIFF
--- a/documentation/src/main/resources/pages/ditto/connectivity-manage-connections.md
+++ b/documentation/src/main/resources/pages/ditto/connectivity-manage-connections.md
@@ -38,7 +38,8 @@ for configuration details.
 
 ### Create a connection
 
-Send an HTTP `POST` request:
+Send an HTTP `POST` request. Ditto generates the connection ID. Do not put an `id` in the body
+(that returns `connectivity:id.notsettable`).
 
 ```
 POST /api/2/connections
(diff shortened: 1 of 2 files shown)


thjaeckle:
Thanks for the documentation enhancement @AzazelSensei 👍   
In order for us to be able to accept and merge this, you would need to sign an ECA (Eclipse Contributor Agreement) and link your email-address you used for the Git commit in your Eclipse profile as described in the [https://github.com/eclipse-ditto/ditto/blob/master/CONTRIBUTING.md](CONTRIBUTING.md)

AzazelSensei:
I'll sign the ECA and add this GitHub account to the Eclipse profile.

Issue #2146, opened 2025-03-18
WoT Model Validation Blocks Removal of Undefined Features and Causes Internal Server Error on Deletion


hu-ahmed:
When strict WoT model validation is enabled, attempting to remove features that are not defined in the model leads to validation failures. Specifically:

**PATCH Request Issue:**

- Action: Attempted to remove extra features by setting them to null.
- Outcome: The request failed with a 400 Bad Request error due to WoT validation rejecting the presence of undefined features.

**DELETE Request Issue:**

- Action: Attempted to delete the extra features directly.
- Outcome: The request failed with a 500 Internal Server Error, and logs indicate a NoSuchElementException occurring in validateFeatureScopedDeletion().

PR #2545, opened 2026-09-05
Fix 500 on WoT delete of undefined features

sankalpsthakur:
### Summary
With strict WoT validation enabled, deleting a feature that is not in the Thing model threw `NoSuchElementException` in `validateFeatureScopedDeletion()` and returned HTTP 500.

The method now returns success when the feature is not defined in the model, so extra-feature deletes succeed instead of failing as an ISE.

PATCH of extra features to null still 400 under `forbidNonModeledFeatures` (strict WoT). That half of #2146 is not in this PR.

### Validation
- Reproduced: `validateFeatureDeletionSucceedsForUndefinedFeature` completed exceptionally with `NoSuchElementException` before the guard.
- `mvn -pl wot/validation -am test -Dtest=WotThingModelValidationFeatureLevelTest` — 76 tests, 0 failures (Java 25)

Related to #2146

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.

### DIFF
--- a/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/DefaultWotThingModelValidation.java
+++ b/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/DefaultWotThingModelValidation.java
@@ -364,9 +364,13 @@ final class DefaultWotThingModelValidation implements WotThingModelValidation {
             final JsonPointer resourcePath,
             final ValidationContext context
     ) {
+        if (!featureThingModels.containsKey(featureId)) {
+            // Feature is not defined in the Thing model; deleting it (or its properties) is always allowed.
+            return success();
+        }
+
         if (validationConfig.getFeatureValidationConfig().isEnforcePresenceOfModeledFeatures() &&
-                resourcePath.equals(Thing.JsonFields.FEATURES.getPointer().addLeaf(JsonKey.of(featureId))) &&
-                featureThingModels.containsKey(featureId)
+                resourcePath.equals(Thing.JsonFields.FEATURES.getPointer().addLeaf(JsonKey.of(featureId)))
         ) {
             final WotThingModelPayloadValidationException.Builder exceptionBuilder =
                     WotThingModelPayloadValidationException
--- a/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/DefaultWotThingModelValidation.java
+++ b/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/DefaultWotThingModelValidation.java
@@ -364,13 +364,9 @@ final class DefaultWotThingModelValidation implements WotThingModelValidation {
             final JsonPointer resourcePath,
             final ValidationContext context
     ) {
-        if (!featureThingModels.containsKey(featureId)) {
-            // Feature is not defined in the Thing model; deleting it (or its properties) is always allowed.
-            return success();
-        }
-
         if (validationConfig.getFeatureValidationConfig().isEnforcePresenceOfModeledFeatures() &&
-                resourcePath.equals(Thing.JsonFields.FEATURES.getPointer().addLeaf(JsonKey.of(featureId)))
+                resourcePath.equals(Thing.JsonFields.FEATURES.getPointer().addLeaf(JsonKey.of(featureId))) &&
+                featureThingModels.containsKey(featureId)
         ) {
             final WotThingModelPayloadValidationException.Builder exceptionBuilder =
                     WotThingModelPayloadValidationException
--- a/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/InternalFeatureValidation.java
+++ b/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/InternalFeatureValidation.java
@@ -368,6 +368,11 @@ final class InternalFeatureValidation {
         final JsonPointer propertiesPath =
                 resourcePath.getSubPointer(2).orElse(resourcePath); // cut /features/<featureId>
 
+        if (propertiesPath.isEmpty()) {
+            // Whole-feature deletion is governed by enforcePresenceOfModeledFeatures, not property requirements.
+            return success();
+        }
+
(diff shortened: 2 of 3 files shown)


thjaeckle:
@hu-ahmed could you have a look as you reported #2146

hu-ahmed:
Thanks for digging into #2146 — the stack trace you identified is real, and the added test does reproduce it (it fails on `master`, passes on this branch; I verified both).

That said, I think the guard is placed on the wrong condition, and as written it trades the 500 for a silent gap in validation. Details below, with the checks I ran.

## Root cause

The `NoSuchElementException` originates at `InternalValidation.java:173`:

```java
final JsonPointer resourcePointer = resourcePath.getSubPointer(1).orElseThrow();
```

and the empty pointer that reaches it is produced one frame up, in `InternalFeatureValidation.enforcePresenceOfRequiredPropertiesUponFeatureLevelDeletion` (line 368):

```java
final JsonPointer propertiesPath = resourcePath.getSubPointer(2).orElse(resourcePath); // cut /features/<featureId>
```

For a whole-feature delete, `resourcePath` is `/features/<id>` (levelCount 2). `ImmutableJsonPointer.getSubPointer(2)` calls `subList(2, 2)`, which is legal and returns an *empty list* — so it yields `Optional.of(<empty pointer>)`, not `Optional.empty()`, and the `orElse(resourcePath)` fallback never fires. `propertiesPath` then has `getLevelCount() == 0`, misses the `== 1` branch at `InternalValidation.java:154`, falls into the `else`, and `getSubPointer(1).orElseThrow()` throws.

So the trigger is **an empty properties path on a whole-feature delete** — not "the feature is absent from the submodel map."

## 1. The guard disables required-property enforcement for a legitimate class of features

`featureThingModel` (singular) is resolved from the **feature instance's own `definition`**, independently of the Thing TM's submodels — see `DefaultWotThingModelValidator.java:667-672` and `DeleteFeaturePropertyStrategy.java:76-80`. A feature can therefore carry its own valid model while not being a submodel of the Thing's TM.

`!featureThingModels.containsKey(featureId) → success()` short-circuits `enforcePresenceOfRequiredPropertiesUponFeatureLevelDeletion` for every such feature, so deleting a **required** property of one now succeeds silently:

```java
@Test
public void deleteRequiredPropertyOfNonSubmodelFeatureWithOwnModel() {
    internalCheckFail(true, sut.validateFeatureScopedDeletion(
            Map.of(),                   // not a Thing-TM submodel
            KNOWN_FEATURE_LEVEL_TM,     // but has its own model, with required "config" properties
            KNOWN_FEATURE_ID,
            JsonPointer.of("features/" + KNOWN_FEATURE_ID + "/properties/" + CATEGORY_CONFIG),
            provideValidationContext()
    ));
}
```

On `master` this passes (the deletion is correctly rejected). On this branch it fails — the future completes normally.

Worth noting that moving `containsKey` out of the first condition is behaviour-neutral there, since the check was already inside it. Disabling the property enforcement is the only behavioural change this PR makes.

## 2. The 500 is still reachable

The same exception still occurs when the feature **is** a modeled submodel and `enforcePresenceOfModeledFeatures` is `false` — the first branch is skipped for a different reason, and the empty path reaches the same `orElseThrow`:

```
java.util.NoSuchElementException: No value present
  at InternalValidation.enforcePresenceOfRequiredPropertiesUponDeletion(InternalValidation.java:173)
  at InternalFeatureValidation...UponFeatureLevelDeletion$0(InternalFeatureValidation.java:383)
```

Identical before and after the patch.

## Suggested fix

Guarding the empty path at its source handles both cases, and `validateFeatureScopedDeletion` then needs no change at all — `featureThingModels` stays used only where it belongs:

```diff
--- a/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/InternalFeatureValidation.java
+++ b/wot/validation/src/main/java/org/eclipse/ditto/wot/validation/InternalFeatureValidation.java
@@
         final JsonPointer propertiesPath =
                 resourcePath.getSubPointer(2).orElse(resourcePath); // cut /features/<featureId>
 
+        if (propertiesPath.isEmpty()) {
+            // deletion of the complete feature: this is governed by "enforcePresenceOfModeledFeatures",
+            //  no property scoped "required" check applies here
+            return success();
+        }
+
         final CompletableFuture<Void> firstStage;
```

Applied to `master` on its own (without the change in this PR), the full `wot/validation` module is green — `Tests run: 127, Failures: 0, Errors: 0` — including this PR's scenario, the `enforcePresenceOfModeledFeatures=false` case, and the required-property case above.

If the preference is instead to treat a whole-feature delete as "delete all properties" and reject it with a 400 when required properties exist, the alternative is `getLevelCount() <= 1` at `InternalValidation.java:154`. That's a broader semantic change though, and `enforcePresenceOfModeledFeatures` already expresses that intent.

## Minor: the added test's fixture isn't a reachable state

`validateFeatureDeletionSucceedsForUndefinedFeature` passes `KNOWN_FEATURE_LEVEL_TM` as the resolved model for `"unknown-feature"`, i.e. a feature that simultaneously has and hasn't a model. In production a feature with no `definition` never reaches this code — `fetchResolveAndValidateWith(null, …)` returns `success()` at `DefaultWotThingModelValidator.java:856-860`. The case that *does* reach it is a feature with its own definition that isn't a submodel, which is the same population affected by issue 1 above. Might be worth naming the test for that scenario instead.

sankalpsthakur:
Fixed in 090b5abe89. The guard now handles the empty property path, and non-submodel features still enforce their own required properties. Added the disabled-presence case and renamed the original fixture test. All 127 WoT validation tests pass on JDK 25.

hu-ahmed (approved):
LGTM!
Thanks for the improvements

PR #2562, opened 2026-09-18
Fix lost MQTT publishes received right after connect

alstanchev:
Summary

MQTT publishes received right after a (re)connect could get lost — on any MQTT version. MqttClientActor stopped buffering incoming publishes immediately after creating the consumer actors, while a consumer actor only subscribed to the publish flow asynchronously (Source.fromPublisher subscribes within the stream interpreter's preStart). A publish arriving in that window was routed to BufferingFlowableWrapper's unbuffered subject without any real subscriber, i.e. it was dropped and never acknowledged (QoS 1/2: no PUBREC/PUBACK).

This is what made MqttClientActorIT.testSingleTopicAfterReconnect fail "flakily": the window is sub-millisecond and opens mostly on a cold JVM, which is why the first parametrizations (MQTT 3.1.1) failed and it looked protocol-specific.

Evidence

Mosquitto (log_type all): in failing rounds the broker logs Sending PUBLISH to ditto-… ~2 ms after SUBACK, and no PUBREC from Ditto ever follows.
Instrumented wrapper: stopBuffering() → item arrives with only the unsolicited-ack subscriber attached → consumer actor's stream attaches 0.3 ms later.
Changes

MqttConsumerActor takes a Publisher<GenericMqttPublish> instead of a Source, subscribes to it itself within preStart (Source.asSubscriber) and reports its readiness to its parent (Status.Success, as KafkaPublisherActor does). As an actor processes no message before preStart completed, that report proves the subscription.
MqttClientActor handles the subscribe results as a message: consumer actors are started and tracked on the actor's thread (as BaseClientActor#startConsumerActors requires), and buffering is stopped only once all consumer actors reported their readiness (same pattern as the pending status reports of KafkaClientActor).
SubscribeResult carries the Publisher of MQTT publishes instead of a Source.
BufferingFlowableWrapper no longer completes the buffered subject when buffering stops and uses volatile flags, which removes the remaining unsynchronized access from two threads.
No protocol-visible change: same CONNECT/SUBSCRIBE/PUBLISH/ack exchange with the broker; backpressure handling is unchanged (toFlowable(BUFFER) remains the buffering point).

Verification

New tests: MqttConsumerActorTest#consumesMqttPublishesBeforeReportingReadinessToParent and MqttClientActorTest#bufferingOfPublishesIsStoppedOnlyAfterConsumerActorsAndUnsolicitedPublishesAckAreSubscribed (the latter fails under the old ordering).
MqttClientActorIT#testSingleTopicAfterReconnect*: 3 consecutive runs 24/24 green (before: 1–2 failures per run); broker log shows every PUBLISH acknowledged.
Full MqttClientActorIT: 56/56. MQTT package unit tests: 628/628.
MqttConsumerActorTest#mqttPublishSourceGetsThrottledIfThrottlingIsEnabled was removed: it mocked Source#throttle on the injected Source, an implementation detail that no longer exists; mqttPublishSourceThrottlingWorksAsExpected covers the behaviour.


alstanchev:
https://github.com/eclipse-ditto/ditto/actions/runs/35379312827

thjaeckle (approved):
LGTM 👍

